PR #1, opened 2017-09-02
Fixing the implementation of MOSA

apanichella:
Improving the implementation of MOSA, which is now closer to the original implementation.

Instead of using the traditional dominance comparator and the non-dominated sorting algorithm, this implementation contains a number of improvements to speed-up the search, including (i) a fast non-dominance comparator described in [J1], and (ii) using a partial non-dominated sorting which ranks only the top N individual, with N=population size (see the original implementation of MOSA)

[J1] A. Panichella, "Automated Test Case Generation as a Many-Objective Optimisation Problem with Dynamic Selection of the Targets", TSE

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
@@ -48,9 +48,8 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
 
                 var ind = selection()
 
-                var eval = getMutatator().mutateAndSave(ind, archive)
-
-                nextPop.add(Data(eval as EvaluatedIndividual))
+                getMutatator().mutateAndSave(ind, archive)
+                        ?.let{nextPop.add(Data(it))}
 
                 if (!time.shouldContinueSearch()) {
                     break
@@ -85,8 +84,9 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
             // Assign crowding distance to individuals
             subvectorDominance(notCovered, front)
             // Add the individuals of this front
-            for (d in front)
+            for (d in front) {
                 population.add(d)
+            }
 
             // Decrement remain
             remain = remain - front.size
--- a/core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
@@ -5,6 +5,10 @@ import org.evomaster.core.search.EvaluatedIndividual
 import org.evomaster.core.search.Individual
 import org.evomaster.core.search.Solution
 import org.evomaster.core.search.service.SearchAlgorithm
+import org.evomaster.core.LoggingUtil
+import java.util.ArrayList
+
+
 
 /**
  * Implementation of MOSA from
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
```diff
@@ -32,51 +42,73 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
...
-            val nextPop: MutableList<EvaluatedIndividual<T>> = mutableListOf()
+            val nextPop: MutableList<Data> = mutableListOf()
+
+            while (nextPop.size < n-1) {
 
-            while (nextPop.size < n) {
+                var ind = selection()
 
-                getMutatator().mutateAndSave(selection(), archive)
-                        ?.let { nextPop.add(it) }
```
arcuri82: evaluating an individual might return null (long story...). so, should add to population only when != null. To do this, a nice way is Kotlin is to use "?.let" , which means: execute the lambda expression only if the caller is not null. because it is a "let", inside the lambda the "it" refers to the caller

core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
```diff
@@ -32,51 +42,73 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
...
 
-        subvectorDominance(notCovered, list)
+        // Obtain the next front
+        var front = fronts[index]
+
+        while (front!=null && remain > 0 && remain >= front.size && front.isNotEmpty()) {
+            // Assign crowding distance to individuals
+            subvectorDominance(notCovered, front)
+            // Add the individuals of this front
+            for (d in front)
```
arcuri82: please, all blocks should be within "{}", eg "for(d in front){...}"
I should add something like a "Checkstyle" to enforce those rules...

core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
```diff
@@ -108,78 +140,147 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
...
+            if (population[sel].rank < population[min].rank) {
                 min = sel
+            } else if (population[sel].rank == population[min].rank){
+                if (population[sel].crowdingDistance < population[min].crowdingDistance)
+                    min = sel
             }
         }
 
-        return population[min].copy()
+        return (population[min].ind as EvaluatedIndividual<T>).copy()
```
arcuri82: "as" is just a cast. but are you not adding "Data" to the population. Have you try to run EvoMaster with MOSA? u d need to use "--algorithm MOSA"

apanichella: The individual is added to the population later in the method "Search()" as follows:
    _var ind = selection()
    var eval = getMutatator().mutateAndSave(ind, archive)
    nextPop.add(Data(eval as EvaluatedIndividual))_

I have run indeed EvoMaster with the parameter "--algorithm MOSA" and it works on my laptop to work after these changes

core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
apanichella: Right!

core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
apanichella: Given the new version of the code, I guess I should change these lines:
_var eval = getMutatator().mutateAndSave(ind, archive)
nextPop.add(Data(eval as EvaluatedIndividual))_
as follows:
getMutatator().mutateAndSave(ind, archive)?.let(nextPop.add(Data(it))_

Let me make this change and test it briefly on my laptop.

apanichella:
Hi Andrea,
I have implemented the few fixes you suggested. I also tested the code by running EvoMaster with the parameter "--algorithm MOSA". It works pretty good with EMB, much better than the previous version of the code.

arcuri82:
thx for the PR. now it is merged

PR #6, opened 2018-08-28
format body para. of generated test cases

ZhangMan1126:
The original code of TestCaseWriter allows formatting json string if "\n" exists.  So I only create a formatter for json string (add "\n"s) using gson as shown in OutputFormatter.kt. This class can be also extended for another formats.

In addition, we may also format outputs (i.e., getFormattedString() ) based on the type, such as BodyParam or ObjectGene. But I just take string currently to reduce changes on original code.


arcuri82:
hi, thx. but as you can see on the links above, the build on Travis fails due to failing tests. can u look into them?

PR #15, opened 2018-09-27
check for repeated table/columns in Java controller

jgaleotti:


jgaleotti:
Hi, 
I tried with the latest revision of the master branch and the problem in scout-api was still occurring.

PR #25, opened 2019-03-01
Fix for import generation (to reflect the updated package structure).

bogdanmarculescu:
Tiny change. The test should compile now.


arcuri82:
thanks!

PR #26, opened 2019-03-05
Object generation

bogdanmarculescu:
For each gene, objects are generated based on the Swagger definitions. 

The objects are stored in a UsedObjs objects, attached to the RestIndividual.


core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilder.kt
```diff
@@ -274,13 +276,14 @@ class RestActionBuilder {
                 always be empty. Using a MapGene of strings would allow us to at least
                 try to add some fields to it
              */
-            return createMapGene(
+            val mapGene = createMapGene(
```
arcuri82: we need to discuss this one... as I am not sure it is correct

core/src/main/kotlin/org/evomaster/core/problem/rest/RestIndividual.kt
```diff
@@ -77,5 +77,33 @@ class RestIndividual(val actions: MutableList<RestAction>,
         }
     }
 
-
+    fun enforceCoherence(): Boolean {
```
arcuri82: add a comment for what is this function needed for

core/src/main/kotlin/org/evomaster/core/problem/rest/RestIndividual.kt
```diff
@@ -77,5 +77,33 @@ class RestIndividual(val actions: MutableList<RestAction>,
...
+        actions.forEach { action ->
+            action.seeGenes().forEach { gene ->
```
arcuri82: instead of two nested forEach, i think this could be re-written with a flatMap

core/src/main/kotlin/org/evomaster/core/problem/rest/UsedObjs.kt
```diff
@@ -0,0 +1,91 @@
+package org.evomaster.core.problem.rest
+
+import org.evomaster.core.search.gene.*
+
+class UsedObjs {
```
arcuri82: abbreviated words are fine for local variables. but, for classes, I would prefer something more explicit like UsedObjects

core/src/main/kotlin/org/evomaster/core/problem/rest/UsedObjs.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun assign(key:Pair<RestCallAction, Gene>, value:Gene, selectedField:Pair<String, String>){
+        mapping[Pair(key.first.id, key.second.getVariableName())] = value
+        selection[Pair(key.first.id, key.second.getVariableName())] = selectedField
+    }
+
+    fun selectbody(action:RestCallAction, obj:Gene){
+        select_body[action.id] = obj
+    }
+
+    fun clearLists(){
```
arcuri82: why "Lists" in the name? you are clearing all state, including maps, isn't it? what about just "clear()"?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -190,18 +201,176 @@ class RestSampler : Sampler<RestIndividual>() {
...
+                    restrictedModels[k] = ObjectGene(model.name, fields)
+                }
+            }
+        }
 
-        return RestIndividual(actions, SampleType.RANDOM)
+        //Having filtered the objects, build the probability map.
+        val likely = likelyhoodsExtended(g.getVariableName(), restrictedModels)
+                .toList()
+                .sortedBy { (_, value) -> -value}
```
arcuri82: i don't understand. what's the point to sorting it if then u collect the results in an unsorted map?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -190,18 +201,176 @@ class RestSampler : Sampler<RestIndividual>() {
...
+                val innerGene = when (g::class){
+                    OptionalGene::class -> (g as OptionalGene).gene
+                    DisruptiveGene::class -> (g as DisruptiveGene<*>).gene
+                    else -> g
+                }
+
+                val (proposed, field) = proposeObject(innerGene)
+
+                when(field.second) {
+                    "Single_gene" -> {
```
arcuri82: we should try to avoid "magic" strings... where is that one coming from? likely could put it in a shared final variable

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -190,18 +201,176 @@ class RestSampler : Sampler<RestIndividual>() {
...
+                        if (g.isMutable()) g.randomize(randomness, probabilistic)
+                    }
+                    "Complete_object" -> {
```
arcuri82: see comment above

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -594,4 +772,80 @@ class RestSampler : Sampler<RestIndividual>() {
                 }
     }
 
+
+
+    fun lcs(a: String, b: String): String {
```
arcuri82: what does lcs stand for? best to use full names. Remember that, in IntelliJ, there is auto-completion... so long descriptive names are not a problem :)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -594,4 +772,80 @@ class RestSampler : Sampler<RestIndividual>() {
...
+                    }
+                }
+            }
+        }
+        return res
+    }
+
+    fun pickWithProbability(map: MutableMap<Pair<String, String>, Float>): Pair<String, String>{
+        val randFl = randomness.nextFloat()
+        var temp = 0.toFloat()
```
arcuri82: wouldn't "0f" just work here? (or any equivalent in Kotlin)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestSampler.kt
```diff
@@ -594,4 +772,80 @@ class RestSampler : Sampler<RestIndividual>() {
...
+        return found
+    }
+
+
+
+
+    fun <T> likelyhoodsExtended(parameter: String, candidates: MutableMap<String, T>): MutableMap<Pair<String, String>, Float>{
+        //TODO BMR: account for empty candidate sets.
+        val result = mutableMapOf<Pair<String, String>, Float>()
+        var sum : Float = 0.toFloat()
```
arcuri82: see previous comments

PR #33, opened 2019-04-04
modify tracking

ZhangMan1126:
enable "add an element into undo track for evaluated individual"


ZhangMan1126:
the problem is fixed

PR #34, opened 2019-04-04
Naming sorting

bogdanmarculescu:
The current proof-of-concept for custom naming and sorting. 
The sorting is currently done by applying the comparators. At some point I will introduce a more explicit sorting index.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -529,4 +529,8 @@ class EMConfig {
             "Note that we enforced that set enableTrackIndividual false when enableTrackEvaluatedIndividual is true since information of individual is part of evaluated individual")
     var enableTrackEvaluatedIndividual = false
 
+
+    @Experimental
+    @Cfg("Enable custom naming and sorting criteria")
+    var customNaming = false
 }
--- a/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
@@ -1,5 +1,11 @@
 package org.evomaster.core.output
 
+
+import org.evomaster.core.problem.rest.HttpVerb
+import org.evomaster.core.problem.rest.RestCallAction
+import org.evomaster.core.problem.rest.RestCallResult
+import org.evomaster.core.problem.rest.RestIndividual
+import org.evomaster.core.search.EvaluatedIndividual
 import org.evomaster.core.search.Solution
 
 /**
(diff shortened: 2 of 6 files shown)


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -241,10 +241,6 @@ class EMConfig {
                     "collecting heuristics with 'heuristicsForSQL'")
         }
```
arcuri82: did u rebase agaisnt the most recent version? looks like here you modified a file, Man did modifications on it, and you either didn't pull that version, or didn''t merge her changes

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -529,4 +525,8 @@ class EMConfig {
             "Note that we enforced that set enableTrackIndividual false when enableTrackEvaluatedIndividual is true since information of individual is part of evaluated individual")
     var enableTrackEvaluatedIndividual = false
 
+
+    @Experimental
+    @Cfg("Enable non-standard naming and sorting criteria")
```
arcuri82: "non-standard"? as the name of the variable, "custom" sounds better :)

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+            return sortingHelper.sort(solution, namingHelper, customNaming)
+        }
+    }
+}
+
+class NamingHelper {
+    private fun criterion1_500 (individual: EvaluatedIndividual<*>): String{
+        var found500 = false
+        individual.results.forEach {
+            if((it as RestCallResult).getStatusCode() == 500) found500 = true
```
arcuri82: this (ie the "as" cast) will break in the future as soon as we address something more than REST. Should rather use a filterIsInstance on the stream

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+        }
+        val suggestedName = when (found500) {
```
arcuri82: if found500 is a boolean, do we need a "when" instead of an "if"?

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+            true ->  "_with500"
+            false ->  ""
+            else -> throw IllegalStateException("This should not happen")
+        }
+        return suggestedName
+    }
+
+    private fun criterion2_hasPost (individual: EvaluatedIndividual<*>): String{
+        var foundPost = false
+        individual.individual.seeActions().forEach {
```
arcuri82: looks inefficient. better to use ".any{predicate}" instead of "forEach" here (apply this also on other similar code in this file)

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+        val suggestedName = when (foundPost) {
+            true ->  "_hasPost"
+            false ->  ""
+            else -> throw IllegalStateException("This should not happen")
+        }
+        return suggestedName
+    }
+
+    private fun criterion3_sampling(individual: EvaluatedIndividual<*>): String{
+        return "_" + (individual.individual as RestIndividual).sampleType
```
arcuri82: this will break in the future when we use something besides REST

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+    }
+
+    private fun criterion4_dbInit(individual: EvaluatedIndividual<*>): String{
+        if ((individual.individual as RestIndividual).dbInitialization.isNotEmpty()){
```
arcuri82: the "as" here, in the rest of the class is incorrect, as it will break once we start to support something more than REST.
Notice: when you have a block in which you check type with "is", then in such block Kotlin does automatic casting

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+        }
+        else return ""
+    }
+
+    val namingCriteria =  mutableListOf(::criterion1_500, ::criterion2_hasPost, ::criterion3_sampling, ::criterion4_dbInit)
+
+
+    fun suggestName(individual: EvaluatedIndividual<*>): String{
+        var suggestedName = ""
+        namingCriteria.forEach {
```
arcuri82: can replace whole function code with a simple 
return namingCriteria.map{it(individual)}.joinToString("")

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,131 @@ import org.evomaster.core.search.Solution
...
+    }
+
+}
+
+
+class SortingHelper {
+    /** maxStatusCodeComparatorInd sorts Evaluated individuals based on the highest status code (e.g., 500s are first).
+     * **/
+    val maxStatusCodeComparatorInd = compareBy<EvaluatedIndividual<*>>{ind ->
+        val max = ind.results.maxBy { (it as RestCallResult).getStatusCode()!! }
```
arcuri82: again, check all the "as"

arcuri82: also, if a sorting only makes sense on RestIndividuals, could have EvaluatedIndividual<RestIndividual> instead of EvaluatedIndividual<*>

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -13,22 +20,140 @@ import org.evomaster.core.search.Solution
...
+}
+
+
+class SortingHelper {
+    /** maxStatusCodeComparatorInd sorts Evaluated individuals based on the highest status code (e.g., 500s are first).
+     *
+     * **/
+    val maxStatusCodeComparatorInd = compareBy<EvaluatedIndividual<*>>{ind ->
+        val max = ind.results.filterIsInstance<RestCallResult>().maxBy { it.getStatusCode()!! }
+            (max as RestCallResult).getStatusCode() ?: 0
```
arcuri82: is the "as" check here necessary? or redundant?

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
bogdanmarculescu: Using filterIsInstance instead. One exception, for maxStatusCodeComparator, I am using maxBy, which is nullable. As a result, a casting is needed, though it comes after a filterIsInstance, so the worst it can be is null.

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
bogdanmarculescu: Introduced checks to see if it is a RestIndividual.

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
bogdanmarculescu: I kept it as EvaluatedIndividual<*>. In some cases it can be applied on more than just RestIndividuals, so I thought it would be a good idea to keep it more general. 

Having a mix of EvaluatedIndividual<*> and EvaluatedIndividual < RestIndividual > was not possible (compiler was complaining about it).

PR #36, opened 2019-04-09
U1 f984

arcuri82:


arcuri82:
the current "You need to decode it if you want to know what is says..." in the code is the hint... specifying the used encoding would make it too easy... :)

arcuri82:
btw, if you "approved" the PR, you can merge it :)

PR #42, opened 2019-04-24
WiP. Generating basic assertions and initial work on "expectations" (or whatever we will end up calling them).

bogdanmarculescu:
author Bogdan Marculescu <marculescu.bogdan@gmail.com> 1551873900 +0100
committer Bogdan Marculescu <marculescu.bogdan@gmail.com> 1554389113 +0200

WiP. Check added for content type when response code is 200.

WiP. Check added for content type when response code is 200.

WiP. Checks for contents of those calls returning 200 added.

Includes checking the type and contents of the returned JSON objects,
but not, for the moment, collections of objects.

WiP. Checks for contents of those calls returning 200 added.

Includes checking the type and contents of the returned JSON objects,
but not, for the moment, collections of objects.

UPDATE: Fixed bug. Added basic "sort by code". Both WiP.

WiP. Sorting and naming criteria examples added.

WiP. Small changes to handling criteria.

WiP. Small changes to handling criteria. UPDATE: Some pointer issues.

WiP. Separating naming and sorting from the test suite organizer,
for the purpose of maintainability and extendability.

WiP. Minor changes. Just before a rebase.

WiP. Checks for contents of those calls returning 200 added.

Includes checking the type and contents of the returned JSON objects,
but not, for the moment, collections of objects.

WiP. Checks for contents of those calls returning 200 added.

Includes checking the type and contents of the returned JSON objects,
but not, for the moment, collections of objects.

UPDATE: Fixed bug. Added basic "sort by code". Both WiP.

WiP. Sorting and naming criteria examples added.

WiP.
Tiny changes. Preparing for Expectation generation.

WiP. Check added for content type when response code is 200.

WiP. Checks for contents of those calls returning 200 added.

Includes checking the type and contents of the returned JSON objects,
but not, for the moment, collections of objects.

WiP. Small changes to handling criteria.

Fixed minor problem (in experiments?).

Added import for hamcrest matchers.

Build no longer breaks.

Basic assertions added, but disabled in config.
Expectations WiP.


client-java/controller/src/main/java/org/evomaster/client/java/controller/expect/ExpectationHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private ExpectationHandler(){}
+
+    @Override
+    public IndividualExpectation expect(){
+        return this;
+    }
+
+    @Override
+    public IndividualExpectation that(Boolean active, Boolean condition){
```
arcuri82: any reason for using `Boolean` instead of `boolean`? note that something like `!active` would then throw a NPE if `null`. This comment does apply to all cases in which you chose to use  `Boolean`

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -428,6 +531,26 @@ class TestCaseWriter {
...
+    }
+
+    private fun handleExpectations(result: RestCallResult, lines: Lines, active: Boolean){
+
+
+
+        lines.add("expectationHandler()")
+        lines.indented {
+            lines.add(".expect()")
+            lines.add(".that(activeExpectations, true)")
```
arcuri82: this is bit weird... you asserting both `true` and `false` on `activeExpectations`? if this is indeed intended, please add a comment to clarify it

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -539,4 +539,13 @@ class EMConfig {
      */
     @Cfg("QWN0aXZhdGUgdGhlIFVuaWNvcm4gTW9kZQ==")
     var e_u1f984 = false
+
```
arcuri82: add a bit more description in the `@Cfg` annotations

bogdanmarculescu: I think that's the unicorn. I must've screwed something up during the merge.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: The idea is to have a short example of conditions that fail, but still not cause the test case to fail (unless the expectations are active). 
I've added a comment to reflect this, but this will change quite a lot.

client-java/controller/src/main/java/org/evomaster/client/java/controller/expect/ExpectationHandler.java
bogdanmarculescu: Just reflex. Fixed.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: issue with GitHub: it is showing only the snippet above the comment, whereas I meant the 2 new `@Cfg` after that below

bogdanmarculescu: I've expanded the descriptions of those a bit. I'll expand them more, if needed. I'll push the changes in a minute.

PR #46, opened 2019-05-04
refactor of DB constraints

jgaleotti:
check expressions are processed in the core instead of the client


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/DbTableCheckExpression.java
```diff
@@ -0,0 +1,23 @@
...
+    private final /*non-null*/ String sqlCheckExpression;
+
+
+    public DbTableCheckExpression(String tableName, String sqlCheckExpression) {
+        super(tableName);
+        if (sqlCheckExpression == null) {
+            throw new IllegalArgumentException("check expression cannot be null");
+        }
+        this.sqlCheckExpression = sqlCheckExpression;
+    }
```
arcuri82: if you prefer, you could also write it like:
`this.sqlCheckExpression = Objects.requireNonNull(sqlCheckExpression);`
it would simplify the code quite a bit. there are many constructors in this PR in which it could used

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/DbTableUniqueConstraint.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+
+public class DbTableUniqueConstraint extends DbTableConstraint {
+
+    private final List<String> uniqueColumnNames;
+
+    public DbTableUniqueConstraint(String tableName, List<String> uniqueColumnNames) {
+        super(tableName);
+        this.uniqueColumnNames = uniqueColumnNames;
+    }
```
arcuri82: in general, unless performance is a major issue, should copy the input collection, as otherwise external entities could manipulate directly the internal state of this class, eg:
`this.uniqueColumnNames = new ArrayList(uniqueColumnNames)`
furthermore, do we need the ordering? I mean, could use a `Set` which would imply uniqueness

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
```diff
@@ -0,0 +1,132 @@
...
+        allConstraints.addAll(columnConstraints);
+        allConstraints.addAll(tableCheckExpressions);
+        return allConstraints;
+    }
+
+    /**
+     * For each table in the schema DTO, this method appends
+     * the constraints that are originated in the ALTER TABLE commands
+     * for those particular tables.
+     * <p>
```
arcuri82: aren't `CHECK` constraints possible directly on table declaration (ie `CREATE`)  besides `ALTER`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
```diff
@@ -0,0 +1,132 @@
...
+                    continue;
+                } else if (constraintType.equals("CHECK")) {
+                    assert (columnList == null);
+
+                    DbTableCheckExpression constraint = new DbTableCheckExpression(tableName, sqlCheckExpression);
+                    tableCheckExpressions.add(constraint);
+
+                } else {
+                    throw new RuntimeException("Unknown constraint type : " + constraintType);
+                }
```
arcuri82: not sure we should throw an exception here, which might stop EM from working. We could still be able to generate test cases even if cannot handle all possible constraints in different SUTs.
See for example how `HeuristicsCalculator.cannotHandle` works

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
```diff
@@ -0,0 +1,132 @@
...
+     * Appends all Column constraints (i.e. CHECK contraints) to the DTO
+     *
+     * @param connectionToH2 a connection to a H2 database
+     * @param schemaDto
+     * @throws SQLException if the connection to the database fails
+     */
+    private List<DbTableConstraint> extractColumnConstraints(Connection connectionToH2, DbSchemaDto schemaDto) throws SQLException {
+        String tableSchema = schemaDto.name;
+        List<DbTableConstraint> columnConstraints = new LinkedList<>();
+        for (TableDto tableDto : schemaDto.tables) {
```
arcuri82: any specific reason for using `LinkedList` instead of `ArrayList`? I always thought that `LinkedList` should be avoided as more expensive / slower, but maybe I am wrong?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/PostgresTableConstraintExtractor.java
```diff
@@ -0,0 +1,75 @@
...
+                        }
+                        break;
+                    }
+                    case CONSTRAINT_TYPE_FOREIGN_KEY:
+                    case CONSTRAINT_TYPE_PRIMARY_KEY:
+                    case CONSTRAINT_TYPE_UNIQUE:
+                    case CONSTRAINT_TYPE_TRIGGER:
+                    case CONSTRAINT_TYPE_EXCLUSION: {
+                        // ignore
+                    }
```
arcuri82: could add a comment clarifying why they should be ignored

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/TableConstraintExtractor.java
```diff
@@ -0,0 +1,14 @@
+package org.evomaster.client.java.controller.internal.db.constraint;
+
+import org.evomaster.client.java.controller.api.dto.database.schema.DbSchemaDto;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.List;
+
+public abstract class TableConstraintExtractor {
```
arcuri82: Add a comment specifying that constraints are handled differently in different databases (there is no standard for it) and JDBC does not support it, and so we need customize extractors for each single database we want to support

core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
```diff
@@ -121,6 +129,78 @@ class SqlInsertBuilder(
...
+            return rangeConstraint.minValue.toInt()
+        }
+
+        val lowerBounds = tableConstraints
+                .asSequence()
+                .filter { c -> c is LowerBoundConstraint }
+                .map { c -> c as LowerBoundConstraint }
+                .filter { c -> c.columnName.equals(columnName, true) }
+                .map { c -> c.lowerBound.toInt() }
+                .toList()
```
arcuri82: Kotlin (in contrast to Java) has a stream function called `filterIsInstance` which could simplify the code here, as it does a `map` as well in a single call. Could apply also in the other places in this file where there is similar code

core/src/test/resources/sql_schema/check_in.sql
```diff
@@ -0,0 +1,18 @@
...
+  LAST_UPDATED_AT    TIMESTAMP,
+  LAST_UPDATED_BY    TEXT,
+  W_ID               TEXT NOT NULL,
+  EXPR_DATE          DATE NOT NULL
+);
+
+ALTER TABLE X ADD CONSTRAINT check_status
+  CHECK (STATUS in ('A', 'B'));
+
+
```
arcuri82: could also make sense to have a case in which the `CHECK` is directly in the `CREATE` and not just the `ALTER`. I would assume that the databases would treat them exactly the same, but can never be sure :)

db-constraints/src/main/java/org/evomaster/constraint/EnumConstraint.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private final List<String> valuesAsStrings;
+
+    private final String columnName;
+
+    public EnumConstraint(String tableName, String columnName, List<String> valuesAsStrings) {
+        super(tableName);
+        this.columnName = columnName;
+        this.valuesAsStrings = valuesAsStrings;
+    }
```
arcuri82: in this class (and the others), if an input X must not be null, could write it as 
`this.x = Objects.requireNonNull(x);`

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/db/base/DbBaseEMTest.java
```diff
@@ -1,14 +1,14 @@
...
 import org.evomaster.core.problem.rest.HttpVerb;
 import org.evomaster.core.problem.rest.RestIndividual;
 import org.evomaster.core.search.Solution;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class DbBaseEMTest extends DbBaseTestBase {
+public class
+DbBaseEMTest extends DbBaseTestBase {
```
arcuri82: why on 2 lines?

db-constraints/src/main/java/org/evomaster/constraint/AndConstraint.java
```diff
@@ -0,0 +1,14 @@
+package org.evomaster.constraint;
+
```
arcuri82: when we have a module, the package name should be matching somehow the name of the of module.
Here, the package is `constraint` whereas the module is `db-constraints`. I think it would be better if we rename the package with something like `org.evomaster.dbconstraint` (should not have `-` in the name of a package)

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/DbTableCheckExpression.java
jgaleotti: Ok. I will add those checks and remove the if(...) statements.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/DbTableUniqueConstraint.java
jgaleotti: Thanks. I looked for all the references to LinkedList and replace them with ArrayLists.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
jgaleotti: Thanks. I replaced all references to LinkedList with ArrayList

core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
jgaleotti: Thanks. I changed three parts where the filter{...}.map{...} was being used with a corresponding filterIsInstance().

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/db/base/DbBaseEMTest.java
jgaleotti: It appears to be a phantom ENTER keypress... Fixed.

db-constraints/src/main/java/org/evomaster/constraint/AndConstraint.java
jgaleotti: Thanks. I renamed the module to "dbconstraint" and the package to org.evomaster.dbconstraint to satisfy the naming convention.

db-constraints/src/main/java/org/evomaster/constraint/EnumConstraint.java
jgaleotti: Yes. I missed that. I added the checks on the missing classes in packages org.evomaster.dbconstraint, org.evomaster.client.java.controller.internal.db.constraint and org.evomaster.dbconstraint.ast

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
jgaleotti: Indeed. The documentation is not clear. I changed the documentation and I also added two more test cases for extracting constraints to clarify the behaviour.

core/src/test/resources/sql_schema/check_in.sql
jgaleotti: Makes sense. I added a new test to confirm that the enum constraint is also extracted when declared within a CREATE TABLE statement.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/H2ConstraintExtractor.java
jgaleotti: Understood. I will add a "cannotHandle" warning when extracting unsupported constraints in both H2 and Postgres

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/PostgresTableConstraintExtractor.java
jgaleotti: Thanks. I added the missing documentation and by doing that I realised that we were not properly handling Postgres unique constraints.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/constraint/TableConstraintExtractor.java
jgaleotti: Thanks. I added the required comments.

PR #48, opened 2019-05-06
fix nullPointerException when sql selecting a column with null value

ZhangMan1126:
fix java.lang.nullPointerException when a sql command is to select a column with null value.


arcuri82:
thx!

PR #50, opened 2019-05-15
retrieving auto-ids when doing SQL insertions

arcuri82:
now possible to execute SQL insertions, and retrieve mapping for the auto-generated primary keys. this is a starting point to support multi-insertions in test cases.


client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
```diff
@@ -145,16 +145,30 @@ public static void runCommands(Connection connection, List<String> commands) {
     }
 
 
-    public static void execInsert(Connection conn, List<InsertionDto> insertions) throws SQLException {
+    //From DTO Insertion Id to generated Id in database
```
arcuri82: instead of //, better to use /** */ , so that it becomes part of the JavaDocs

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
```diff
@@ -145,16 +145,30 @@ public static void runCommands(Connection connection, List<String> commands) {
...
+    public static Map<Long, Long> map = new HashMap<>();
```
arcuri82: first, mutable fields should not be `public`
second, I am usually against static mutable fields (apart from very special cases). Read what I wrote in the `for_developers.md` file. I think this should be refactored. If not possible, we need to discuss it

PR #52, opened 2019-06-04
dynamically set fk based on sql id maps

ZhangMan1126:


core/src/main/kotlin/org/evomaster/core/database/DbActionTransformer.kt
```diff
@@ -10,7 +10,7 @@ import org.evomaster.core.search.gene.SqlPrimaryKeyGene
 
 object DbActionTransformer {
 
-    fun transform(insertions: List<DbAction>) : DatabaseCommandDto {
+    fun transform(insertions: List<DbAction>, sqlIdMap : Map<Long, Long> = mapOf()) : DatabaseCommandDto {
```
arcuri82: add comment on what <Long, Long> represent

ZhangMan1126: thanks. comments are added.

PR #53, opened 2019-06-05
Basic assertions v2

bogdanmarculescu:
Some issues persist. 

1. Messages containing timestamps and/or object references cause failures.
2. Extension of Json for ProxyPrint causes some trouble. The string describing it is (in both EvoMaster and in the Api itself "application/vnd.tsdes.news+json;charset=UTF-8;version=2", but for some reason the RestAssured header reads "application/vnd.tsdes.news+json; version=2;charset=UTF-8". The reversal of charset and version appears to cause the problems.

### DIFF
--- a/client-java/controller/pom.xml
+++ b/client-java/controller/pom.xml
@@ -144,6 +144,11 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>compile</scope>
+        </dependency>
 
 
     </dependencies>
--- a/client-java/controller/pom.xml
+++ b/client-java/controller/pom.xml
@@ -128,6 +128,7 @@
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-all</artifactId>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
@@ -144,12 +145,6 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
-            <scope>compile</scope>
-        </dependency>
-
 
     </dependencies>
 
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
@@ -5,7 +5,7 @@ import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 
 public class NumberMatcher extends TypeSafeMatcher<Number> {
-    private double value;
+    private final double value;
 
     public NumberMatcher(double value) {
         this.value = value;
@@ -14,16 +14,19 @@ public class NumberMatcher extends TypeSafeMatcher<Number> {
     @Override
     public void describeTo(Description description) {
         //The point of the matcher is to allow comparisons between int and double that have the same valueE.g. that (int) 0 == (double) 0.0
+        description.appendValue(value);
     }
 
     @Override
     protected boolean matchesSafely (Number item) {
-        return item.doubleValue() == value;
+        if (item == null) return false;
+        else return item.doubleValue() == value;
     }
     public static Matcher<Number> numberMatches(Number item) {
         return new NumberMatcher(item.doubleValue());
     }
     public static boolean numbersMatch(Number item1, Number item2){
+        if (item1 == null || item2 == null) return false;
         NumberMatcher n1 = new NumberMatcher(item1.doubleValue());
         return n1.matchesSafely(item2);
     }
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
@@ -1,0 +1,30 @@
+package org.evomaster.client.java.controller.contentMatchers;
+
+import org.hamcrest.TypeSafeMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+
+public class NumberMatcher extends TypeSafeMatcher<Number> {
+    private double value;
+
+    public NumberMatcher(double value) {
+        this.value = value;
+    }
+
+    @Override
+    public void describeTo(Description description) {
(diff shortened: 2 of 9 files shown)


client-java/controller/pom.xml
```diff
@@ -144,6 +144,11 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
arcuri82: this should be removed. hamcrest-all is already imported in that pom, with scope <test> (declared in the parent-pom). See discussion on adding libraries to controller in comments at the top of this file

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
```diff
@@ -0,0 +1,30 @@
+package org.evomaster.client.java.controller.contentMatchers;
+
+import org.hamcrest.TypeSafeMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+
+public class NumberMatcher extends TypeSafeMatcher<Number> {
+    private double value;
```
arcuri82: i guess this could be "final"?

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public NumberMatcher(double value) {
+        this.value = value;
+    }
+
+    @Override
+    public void describeTo(Description description) {
+        //The point of the matcher is to allow comparisons between int and double that have the same valueE.g. that (int) 0 == (double) 0.0
+    }
```
arcuri82: empty body???

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    protected boolean matchesSafely (Number item) {
+        return item.doubleValue() == value;
+    }
```
arcuri82: i guess should check if item is null

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
```diff
@@ -0,0 +1,30 @@
...
+    public static Matcher<Number> numberMatches(Number item) {
+        return new NumberMatcher(item.doubleValue());
+    }
+    public static boolean numbersMatch(Number item1, Number item2){
+        NumberMatcher n1 = new NumberMatcher(item1.doubleValue());
+        return n1.matchesSafely(item2);
+    }
+}
```
arcuri82: what to do when the items are null?

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -381,50 +387,99 @@ class TestCaseWriter {
...
+                        .replace("\"", "\\\"")
+                        .replace("\n", "\\n")
+                        .replace("\r", "\\r")}\")"
+                //Note: checking a string can cause (has caused) problems due to unescaped quotation marks
+                // The above solution should be refined.
                 else -> return "NotCoveredYet"
             }
         }
+        /* BMR: the code above is due to a somewhat unfortunate problem:
+        - Gson does parses all numbers as Double
```
arcuri82: a comment: JSON has no concept of integers. all numeric values are floating point precision. Swagger could specify that a number is indeed an integer, but Gson would not know about it

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -527,18 +591,53 @@ class TestCaseWriter {
...
+                                if(printableTh != "null"
+                                        && printableTh != "NotCoveredYet"
+                                ){
+                                    lines.add(".that(activeExpectations, (\"${it}\" == \"${resContents[it]}\"))")
+                                }
+                            }
+                        }
+                        else -> {
+                            // this shouldn't be run if the JSON is okay. Panic! Update: could also be null. Pause, then panic!
+                        }
```
arcuri82: add a log statement or an assertion for this empty else branch

core/src/main/kotlin/org/evomaster/core/problem/rest/RestCallAction.kt
```diff
@@ -37,7 +37,8 @@ class RestCallAction(
          * by a POST, where the POST itself might use a location for
          * path coming from a previous POST
          */
-        var locationId: String? = null
+        var locationId: String? = null,
+        var produces: MutableList<String> = mutableListOf()
```
arcuri82: why is this mutable?

core/src/main/kotlin/org/evomaster/core/problem/rest/RestCallAction.kt
```diff
@@ -46,7 +47,7 @@ class RestCallAction(
 
     override fun copy(): Action {
         val p = parameters.asSequence().map(Param::copy).toMutableList()
-        return RestCallAction(id, verb, path, p, auth, saveLocation, locationId)
+        return RestCallAction(id, verb, path, p, auth, saveLocation, locationId, produces)
```
arcuri82: if `produces` is mutable, then here we have a bug: a change in the copy would change the state of this object

core/src/main/kotlin/org/evomaster/core/problem/rest/RestCallAction.kt
bogdanmarculescu: Not sure. I changed it to List. It will be visible once I push the changes.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: I clarified the comment a bit. The swagger could specify that info, but I cannot be certain that it will. In addition, the NumberMatcher can also be used as an example of additional matchers (should they be needed in future). Clarification will become visible once I push the changes.

client-java/controller/pom.xml
bogdanmarculescu: For the record. Updated the scope of hamcrest for the controller to "provided".

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/NumberMatcher.java
bogdanmarculescu: Currently, the match is false if any of the values is null. Added some checks to reflect this behaviour.

arcuri82:
waiting for CI before merging

PR #57, opened 2019-06-13
Graphql

mmehdishahabi:
The graphql package is added to the project and GraphqlCallAction and GraphqlIndivisual and related classes are implemented.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
@@ -1,0 +1,2 @@
+package org.evomaster.core.problem.graphql
+
--- a/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
@@ -1,2 +1,36 @@
 package org.evomaster.core.problem.graphql
 
+import org.evomaster.core.problem.graphql.param.GraphqlParam
+import org.evomaster.core.problem.rest.GraphqlAction
+import org.evomaster.core.search.Action
+import org.evomaster.core.search.gene.Gene
+
+class GraphqlCallAction (
+        /**
+         * Identifier unique within the individual
+         * **/
+        val id: String,
+        val functionName: String,
+        val parameters:  MutableList<GraphqlParam>,
+        val operationType: String
+) : GraphqlAction {
+    override fun shouldCountForFitnessEvaluations(): Boolean = true
+
+    override fun copy(): Action {
+        val p = parameters.asSequence().map(GraphqlParam::copy).toMutableList()
+        return GraphqlCallAction(id, functionName, p, operationType)
+    }
+
+    override fun getName(): String {
+        return "$functionName"
+    }
+
+    override fun seeGenes(): List<out Gene> {
+
+        return parameters.flatMap { it.seeGenes() }
+    }
+
+    override fun toString(): String {
+        return getName()
+    }
+}
--- a/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlIndividual.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlIndividual.kt
@@ -1,0 +1,109 @@
+package org.evomaster.core.problem.graphql
+
+import org.evomaster.core.Lazy
+import org.evomaster.core.database.DbAction
+import org.evomaster.core.database.DbActionUtils
+import org.evomaster.core.problem.rest.*
+import org.evomaster.core.search.Action
+import org.evomaster.core.search.Individual
+import org.evomaster.core.search.gene.*
+import org.evomaster.core.search.service.Randomness
+import org.evomaster.core.search.tracer.TraceableElement
+import org.evomaster.core.search.tracer.TrackOperator
+
+class GraphqlIndividual (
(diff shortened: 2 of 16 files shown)


core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.evomaster.core.problem.rest.GraphqlAction
+import org.evomaster.core.search.Action
+import org.evomaster.core.search.gene.Gene
+
+class GraphqlCallAction (
+        /**
+         * Identifier unique within the individual
+         * **/
+        val id: String,
+        val functionName: String,
```
arcuri82: provide /** */ comment stating what functionName is

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
```diff
@@ -0,0 +1,36 @@
...
+        val parameters:  MutableList<GraphqlParam>,
+        val operationType: String
```
arcuri82: provide /** */ comment stating what operationType is.
furthermore, if the types are limited, could rather be an enum

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlIndividual.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    override fun seeInitializingActions(): List<Action> {
+        return dbInitialization
+    }
+
+    override fun verifyInitializationActions(): Boolean {
+        return DbActionUtils.verifyActions(dbInitialization.filterIsInstance<DbAction>())
+    }
+
+    override fun repairInitializationActions(randomness: Randomness) {
```
arcuri82: this is fine for now.
at a certain point, I will move the DB code in its own module, so it can be re-used.
but, if copy&paste code like this, do write a //TODO comment specifying it will need to be removed/refactored

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlSampleType.kt
```diff
@@ -0,0 +1,5 @@
+package org.evomaster.core.problem.graphql
+
+enum class GraphqlSampleType {
```
arcuri82: write /** */ comment specifying what this is for

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlUsedObjects.kt
```diff
@@ -0,0 +1,99 @@
+package org.evomaster.core.problem.graphql
+
+import org.evomaster.core.problem.rest.GraphqlAction
+import org.evomaster.core.search.gene.*
+
+class GraphqlUsedObjects {
```
arcuri82: as for any copy&paste, need a //TODO comment.
furthermore, "Objects" handling is an experimental feature in REST, which might or might not be needed in GraphQL.
I would suggest to ignore it for now (unless u think it is useful), and maybe add it later if needed

e2e-tests/graphql-spring-examples/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+    <packaging>jar</packaging>
+
+
+
+    <dependencies>
+
+        <!--
+            Issue with Jersey (coming from client-java) using old version
+        -->
+        <dependency>
```
arcuri82: dependency versions should not be here, but rather in the root pom file.
if you do not specify \<version\>, then it is inherited from \<parent\>

e2e-tests/graphql-spring-examples/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.5.1</version>
```
arcuri82: as inherited from \<parent\>, no need for \<version\> and \<configuration\> here

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.evomaster.core.search.Action
+import org.evomaster.core.search.gene.Gene
+
+class GraphqlCallAction (
+        /**
+         * Identifier unique within the individual
+         * **/
+        val id: String,
+        val functionName: String,
+        val parameters:  MutableList<GraphqlParam>,
```
arcuri82: brainstorming here: a GraphQL query can be represented as JSON object. so we could use a single ObjectGene to represent it. or are there cases in which we might need several different type of parameters? I can only think of HTTP Header parameters

arcuri82:
ah, didn't realize this was not your "fork", but the main EM repository :)

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphqlCallAction.kt
mmehdishahabi: I believe we just need one type of parameter for HTTP Header and one for the JSON object.

PR #59, opened 2019-06-14
Resource mio ready

ZhangMan1126:
brief description: 
Rest individual is modified in order to support resource-based structure.

For handling resource-based methods, I created new services:
- ResourceManageService manages all resources.
- ResourceDepManageService manages dependencies of resources.
- RestResourceFitness is for manipulating the execution of resource-based individual, e.g., dbactions (we may merge it with RestFitness).
- RestResourceSampler is resource-based sampler, which enables sampling resource-based individual.
- RestResourceMutator/RestResourceStructureMutator is resource-based mutator, which enables mutating resource-based individual.
- SmartSamplingController is for dynamically determining resource-based sample method.

To assemble services for handling resource-based solution, ResourceRestModule is created.


arcuri82:
thx! a comment before I ll start reviewing it: it looks like you added to git the `report/target` folder... leading to 915 files in this PR... :)
built `target` folders should never be added to Git

ZhangMan1126:
thanks! I run "mvn clean", and submit a new commit.

core/pom.xml
```diff
@@ -135,6 +138,18 @@
             <artifactId>evomaster-dbconstraint</artifactId>
         </dependency>
 
+        <!--for token parser-->
+        <dependency>
+            <groupId>edu.stanford.nlp</groupId>
+            <artifactId>stanford-corenlp</artifactId>
+            <version>${nlp.version}</version>
```
arcuri82: versions should be in the root pom, not in the submodules. see what I wrote recently in `for_developers.md`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+        /**
+         * probability for applicable strategy is adaptive with performance, i.e., Archive
+         */
+        ConArchive (true)
+    }
+
+    @Experimental
+    @Cfg("Specify whether to enable resource-based strategy to sample an individual during search. " +
+            "Note that resource-based sampling is only applicable for REST problem with MIO algorithm.")
+    var resourceSampleStrategy = ResourceSamplingStrategy.NONE
```
arcuri82: in the code (`updateProperties()`), we do have checks if configs are not inconsistent. should add check that, if != NONE, then must be REST and MIO

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Specify whether to enable resource dependency heuristics, i.e, probOfEnablingResourceDependencyHeuristics > 0.0. " +
+            "Note that the option is available to be enabled only if resource-based smart sampling is enable. " +
+            "This option has an effect on sampling multiple resources and mutating a structure of an individual.")
+    var probOfEnablingResourceDependencyHeuristics = 0.5
```
arcuri82: as a probability, should add `@Min(0.0) @Max(1.0)`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Specify whether to export derived dependencies among resources. " +
+            "if enabled, the dependencies are stored at the same location of generated test cases.")
+    var exportDependencies = false
```
arcuri82: this only applies to REST resources, isn't it? in that case, clarify it in the description

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Whether to involve db when applying resource-based methods, e.g., resource-based sampling.")
+    var doesInvolveDB = false
```
arcuri82: "DB" is fine as local variable name, but for descriptions and public variables, best to fully spell it out: `Database`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Specify a minimal number of rows in a table")
+    var minRowOfTable = 10
```
arcuri82: I do not understand what this is for. Please add bit more description.
Furthermore, would a `Min` constraint make sense here?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Specify a probability that enables selection of data from db instead of insertion")
+    @Min(0.0) @Max(1.0)
+    var probOfSelectFromDB = 0.1
```
arcuri82: "from db instead of insertion" is a bit confusing: please clarify it

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Whether to apply token parser to derive relationships between name entities, e.g., a resource identifier with a name of table")
+    var doesApplyNameMatching = false
```
arcuri82: what does "token parser" mean here? did you mean something like "apply text/name analysis"?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -550,4 +550,103 @@ class EMConfig {
...
+    @Min(0.0)@Max(1.0)
+    var S1dR : Double = 0.25
+    @Experimental
+    @Cfg("Specify a probability to apply S2dR when sample control is Customized")
+    @Min(0.0)@Max(1.0)
+    var S2dR : Double = 0.25
+    @Experimental
+    @Cfg("Specify a probability to apply SMdR when sample control is Customized")
+    @Min(0.0)@Max(1.0)
+    var SMdR : Double = 0.25
```
arcuri82: could be good to add an empty line between all these variable declarations, to improve readability

core/src/main/kotlin/org/evomaster/core/database/SQLGenerator.kt
```diff
@@ -0,0 +1,96 @@
+package org.evomaster.core.database
+
+import org.evomaster.core.database.schema.Column
+import org.evomaster.core.database.schema.ColumnDataType
+import org.evomaster.core.database.schema.Table
+
+class SQLGenerator{
```
arcuri82: add comment stating what this class is needed for

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilder.kt
```diff
@@ -25,7 +25,8 @@ class RestActionBuilder {
 
         fun addActionsFromSwagger(swagger: Swagger,
                                   actionCluster: MutableMap<String, Action>,
-                                  endpointsToSkip: List<String> = listOf()) {
+                                  endpointsToSkip: List<String> = listOf(),
+                                  doParserDescription : Boolean = false) {
```
arcuri82: add comment specifying what `doParserDescription`  does

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/CallsTemplate.kt
```diff
@@ -0,0 +1,9 @@
+package org.evomaster.core.problem.rest.resource
+
+class CallsTemplate (
+        val template: String,
+        val independent : Boolean,
+        var size : Int = 1,
+        var times : Int = 0,
+        var sizeAssured : Boolean= !template.contains("POST")
+)
```
arcuri82: add some description for this class and its fields

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
```diff
@@ -0,0 +1,78 @@
+package org.evomaster.core.problem.rest.resource
+
+open class RToken(
```
arcuri82: add comment for this class is used for

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
```diff
@@ -0,0 +1,78 @@
...
+    val originalText : String,
+    protected val lemma : String,
+    var assuredVerb : Boolean = false
```
arcuri82: what is an `assuredVerb `?

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceNode.kt
```diff
@@ -0,0 +1,818 @@
...
+        val actions: MutableList<RestAction> = mutableListOf(),
+        val initMode : InitMode
+) {
+
+    companion object {
+        private const val PROB_EXTRA_PATCH = 0.8
+        val log: Logger = LoggerFactory.getLogger(RestResourceNode::class.java)
+    }
+
+    private val tokens : MutableMap<String, PathRToken> = mutableMapOf()
```
arcuri82: what is the String key here representing?

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceNode.kt
```diff
@@ -0,0 +1,818 @@
...
+
+    private val segments : MutableMap<Boolean, List<String>> = mutableMapOf()
```
arcuri82: a map where the key is a boolean??? (which can only have 2 values)
seems bit weird

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SmartSamplingController.kt
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    fun initialize(){
+        initApplicableStrategies()
+        initProbability()
+        validateProbability()
+    }
+
+    private fun initApplicableStrategies(){
+        ResourceSamplingMethod.S1iR.applicable = true //mutableMap.values.filter { r -> r.hasIndependentAction }.isNotEmpty()
```
arcuri82: I am bit skeptical about this `ResourceSamplingMethod` enum. Unless I misunderstanding, it would look like a global class with 4 instances with mutable state. If that is indeed the case, it would be equivalent to a static state, which we should avoid in EvoMaster

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SmartSamplingController.kt
```diff
@@ -0,0 +1,345 @@
...
+ * ResourceSamplingMethod presents detailed information about how to sample,
+ * including 1) [applicable] a set of applicable strategies; 2) [probability] a probability to select an applicable strategy; 3) [times] times to select an applicable strategy; and 4) [improved] times to help to improve Archive
+ *
+ * @property applicable presents whether a strategy is applicable for a system under test, e.g., if all resource of the SUT are independent, then only [S1iR] is applicable
+ * @property probability presents a probability to apply the strategy during sampling phase
+ * @property times presents how many times the strategy is applied
+ * @property improved presents how many times the strategy helps to improve Archive. Note that improved <= times
+ *
+ * */
+enum class ResourceSamplingMethod (var applicable : Boolean = false, var probability : Double = 0.0, var times : Int = 0, var improved : Int = 0){
```
arcuri82: if indeed this is a mutable public static state, it should be refactored

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -68,13 +68,21 @@ abstract class Individual (trackOperator: TrackOperator? = null, traces : Mutabl
...
      * test after this method is invoked.
      */
-    abstract fun repairInitializationActions(randomness: Randomness);
+    abstract fun repairInitializationActions(randomness: Randomness)
+
+
+    /**
+     * generate id for genes of the individual
+     */
+    open fun seeGenesIdMap() : Map<Gene, String>{
```
arcuri82: this does not seem correct. A `Gene` is a mutable object. You can't have a mutable object as a key of a map, as it will leads to bug as soon as you change the state of any of the key (as those will impact the `hashCode` and `equals` methods)

restresource-aritifical/resource-dependency-model/target/maven-archiver/pom.properties
```diff
@@ -0,0 +1,4 @@
+#Created by Apache Maven 3.5.4
```
arcuri82: this file should not be in git

restresource-aritifical/spring-examples/target/maven-archiver/pom.properties
```diff
@@ -0,0 +1,4 @@
+#Created by Apache Maven 3.5.4
```
arcuri82: this file should not be in git

core/pom.xml
ZhangMan1126: thanks, fixed

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: thanks, all resource-related configurations are fixed.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: thanks, fixed

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
ZhangMan1126: I have added comments for all properties. 

`assuredVerb`  indicates whether the token is a verb. In our case, the tokens in a path is not a complete sentence, so a word may be a noun or a verb. so I use `assuredVerb` to present that the word must be a verb.

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceNode.kt
ZhangMan1126: thanks, all missing comments for map type are fixed.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SmartSamplingController.kt
ZhangMan1126: thanks, fixed.

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
ZhangMan1126: this method is used when applying archived-based mutation, return `identifier`s for all genes in this individual. 

In archived-based mutation, i need to track which gene is mutated and how it is mutated. In this case, each `Gene` requires to have an `identifier` for tracking its evolution among
`individual`. `value` of the `return` of this method is the `identifier` which is composed of gene info and rest action info.  In addition, this method is a kind of utility for getting the `identifier` for genes in this individual, and we do not use the map later. so it is fine even the state of gene changes. Moreover, there is a reason for the usage of the map type as the return:  rest action info is required to get an `identifier` for a gene, so it may take a similar search cost between getting an `identifier` and all `identifier`s. In addition, more than one genes of an individual may be mutated in evomaster. therefore, this method is developed for getting all `identifier`s. because I want to get the `identifier` by gene, I used gene as a key of the map.

arcuri82:
thx! I got bit busy :(  I ll try to check your new commits today, or hopefully by tomorrow

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
arcuri82: i see... but still, mutable keys are a red flag.
what about rather have the option to "bind" a Gene to an Action, and then have a `Gene.identifier` property directly  in the Gene class? (which might throw an exception if not bound to an action)
Might have some side effects on `copy()` method though.

arcuri82: it other words: now an Action knows the Genes it contains, but a Gene does not know the Action it belongs to. Would it be helpful to add such info? (but it would complicate how we create and copy genes)

ZhangMan1126:
Hi, Andrea,

I am also adding more tests, you may start to review once I finish all together. I will re-request a review to inform you that the code is ready.

thank you

ZhangMan1126:
Hi, Andrea,

resource dependency mio is ready to merge into master.
To assess the effects on the resource experiment based on SQL updates, I created another branch to make a further check.

arcuri82:
thanks! i ll review it within the next few days (unfortunately bit busy at the moment :( )

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/SutRunDto.java
```diff
@@ -18,12 +18,27 @@
...
+        this.calculateSqlHeuristics = calculateSqlHeuristics;
+        this.extractSqlExecutionInfo = extractSqlExecutionInfo;
+    }
+
     public SutRunDto(Boolean run, Boolean resetState, Boolean calculateSqlHeuristics) {
         this.run = run;
         this.resetState = resetState;
         this.calculateSqlHeuristics = calculateSqlHeuristics;
+        this.extractSqlExecutionInfo = calculateSqlHeuristics!=null && calculateSqlHeuristics;
     }
```
arcuri82: must avoid copy&paste of same code
this constructor could be rewritten with something like:
```
this(run, resetState, calculateSqlHeuristics, calculateSqlHeuristics )
```

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -555,7 +589,7 @@ class EMConfig {
     var customNaming = false
 
     /*
-        You need to decode it if you want to know what it says...
+        You need to decode it if you want to know what is says...
```
arcuri82: i think i did recently fix this typo, but that was likely after last time you rebased/merged from master?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: what is a `la` parser? is it just a typo?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -575,4 +609,112 @@ class EMConfig {
...
+    @Min(0.0)@Max(1.0)
+    var SMdR : Double = 0.25
+
+    @Experimental
+    @Cfg("Specify a probability to enable archive-based mutation")
+    @Min(0.0) @Max(1.0)
+    var probOfArchiveMutation = 0.0
+
+    @Experimental
+    @Cfg("Specify a percentage to prepare candidates of genes to mutate")
```
arcuri82: I do not understand this one. Can you please add a bit more description?

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -311,6 +315,18 @@ class Main {
             val process = injector.getInstance(SearchProcessMonitor::class.java)
             process.saveOverall()
         }
+
+        private fun writeDependencies(injector: Injector) {
```
arcuri82: can you add a couple of sentences in a comment here explaining what those "dependencies" are?

core/src/main/kotlin/org/evomaster/core/database/DbActionUtils.kt
```diff
@@ -345,5 +348,25 @@ object DbActionUtils {
         }
     }
 
+    fun repairFK(dbAction: DbAction, previous : MutableList<DbAction>, createdDbActions : MutableList<DbAction>,sqlInsertBuilder: SqlInsertBuilder?) : MutableList<SqlPrimaryKeyGene>{
```
arcuri82: add some comments explaining what it means to repair a FK, and when it could happen that a FK is "broken"

core/src/main/kotlin/org/evomaster/core/problem/rest/RestIndividual.kt
```diff
@@ -153,4 +179,101 @@ class RestIndividual (
...
+        resourceCalls.set(position1, resourceCalls[position2])
+        resourceCalls.set(position2, first)
+    }
+
+
+    fun repairDBActions(sqlInsertBuilder: SqlInsertBuilder?){
+        val previousDbActions = mutableListOf<DbAction>()
+
+        getResourceCalls().forEach {
+            if (it.dbActions.isNotEmpty()){
```
arcuri82: this could be a `filter` before the `forEach`, isn't it?

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
```diff
@@ -280,4 +280,16 @@ class RestPath(path: String) {
     private fun encode(s: String): String {
         return URLEncoder.encode(s, "UTF-8")
     }
+
+    fun copy() : RestPath{
+        return RestPath(this.toString())
+    }
+
+    fun getStaticTokens(): List<String> {
```
arcuri82: add a comment clarifying what is a "static" token

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
arcuri82: can u just add this comment? ie, the `assuredVerb indicates whether the token is a verb. In our case, the tokens in a path is not a complete sentence, so a word may be a noun or a verb. so I use assuredVerb to present that the word must be a verb.`
with the `I use` replaced by `we use`

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
```diff
@@ -0,0 +1,93 @@
...
+){
+    var isCollection: Boolean =  !assuredVerb && lemma != originalText
+        private set
+
+    fun equals(other : RToken):Boolean{
+        return isCollection == other.isCollection && lemma == other.lemma
+    }
+
+    fun equals(text: String) : Boolean{
+        return text.toLowerCase() == lemma.toLowerCase() || text.toLowerCase() == originalText.toLowerCase()
```
arcuri82: i think this could be optimized / simplified.
if i remember correctly, the .equals in Kotlin as a extra boolean parameter to specify if case should be ignored

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestFitness.kt
```diff
@@ -0,0 +1,465 @@
...
+                        fv.updateTarget(okId, 1.0, it)
+                        fv.updateTarget(faultId, 0.5, it)
+                    } else if (status in 400..499) {
+                        fv.updateTarget(okId, 0.1, it)
+                        fv.updateTarget(faultId, 0.1, it)
+                    } else if (status in 500..599) {
+                        fv.updateTarget(okId, 0.5, it)
+                        fv.updateTarget(faultId, 1.0, it)
+                    }
+                }
```
arcuri82: I recently made some changes in the RestFitness class. Make sure those will end up here as well

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampleMethodController.kt
```diff
@@ -0,0 +1,337 @@
...
+    private fun initApplicableStrategies(){
+        methods.getValue(S1iR).applicable = true //mutableMap.values.filter { r -> r.hasIndependentAction }.isNotEmpty()
+        rm.getResourceCluster().values.filter { r -> !r.isIndependent() }.let {
+            methods.getValue(S1dR).applicable = it.isNotEmpty()
+            methods.getValue(S2dR).applicable = it.size > 1 && config.maxTestSize > 1
+            methods.getValue(SMdR).applicable = it.size > 2 && config.maxTestSize > 2
+        }
+
+        //FIXME Man Zhang
+        if(methods.values.filter { it.applicable }.size == 1 ) config.maxTestSize = 1
```
arcuri82: can u add a comment here saying what would need to be fixed?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampleMethodController.kt
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    private fun printApplicableStr(){
+        println("Applicable SmartSampleStrategy>>>")
+        println( methods.filter{ it.value.applicable }.mapNotNull { "${it.key} : ${it.value.probability}"}.joinToString (" - "))
+    }
+
+    private fun printSummaryOfResources(mutableMap: Map<String, RestResourceNode>){
+        println("Summary of abstract resources and actions>>>")
```
arcuri82: is this `println` (and the others in this file) actual outputs on standard output? if so, should never be done, and rather use something like `log.info` or `log.debug`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampler.kt
```diff
@@ -0,0 +1,278 @@
...
+
+    private val adHocInitialIndividuals: MutableList<RestIndividual> = mutableListOf()
+
+    private var sqlInsertBuilder: SqlInsertBuilder? = null
+    var existingSqlData : List<DbAction> = listOf()
+        private set
+
+    private val samplingResourceCounter : MutableMap<String , Int> = mutableMapOf()
+    private val samplingComResourceCounter : MutableMap<String , Int> = mutableMapOf()
+    private val separatorResources = "::"
```
arcuri82: in these maps, add comment specifying what is the key and what is the value

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampler.kt
```diff
@@ -0,0 +1,278 @@
...
+    }
+    private fun sampleRandomComResource(resourceCalls: MutableList<RestResourceCalls>){
+        val keys = selectAComResource(randomness)
+        var size = config.maxTestSize
+        var num = 0
+        for (key in keys.split(separatorResources)){
+            rm.sampleCall(key, true, resourceCalls, config.maxTestSize)
+            size -= resourceCalls.last().actions.size
+            num++
+        }
```
arcuri82: are `num` and `size` actually used here? it looks like you are only writing their values, isn't it?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestFitness.kt
```diff
@@ -1,48 +1,20 @@
...
-import javax.ws.rs.ProcessingException
-import javax.ws.rs.client.Client
-import javax.ws.rs.client.ClientBuilder
-import javax.ws.rs.client.Entity
-import javax.ws.rs.core.MediaType
-import javax.ws.rs.core.Response
 
 
-class RestFitness : FitnessFunction<RestIndividual>() {
+class RestFitness : AbstractRestFitness<RestIndividual>() {
```
arcuri82: it is good that you refactored this class, if needed.
however, this kind of refactoring is bit dangerous to do when we have a big PR like this one. it should have rather be done in an independent small, short-lived PR.
for example, there have been changes on master, and those might be missing / get lost here, as it is harder to compare changes when methods are moved around. 
please, double check the last version on Master, and make sure all recent changes will be included in this refactoring

core/src/main/kotlin/org/evomaster/core/problem/util/ParamUtil.kt
```diff
@@ -0,0 +1,671 @@
...
+        fun bindParam(dbAction: DbAction, param : Param, previousToken: String, existingData : Boolean) : Boolean{
+            var gene  : Gene? = null
+            var similarity = 0.0
+            dbAction.seeGenes().forEach findGene@{
+                val score = compareDBGene(dbAction, it, param.name, previousToken)
+                if(score>similarity){
+                    similarity = score
+                    gene = it
+                }
+                if(score == 1.0) return@findGene
```
arcuri82: does the value `1.0` have a special meaning? if so, it could be put in a constant val, with a meaningful name, as it is used in quite a few places in this class

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/SutRunDto.java
ZhangMan1126: fixed, thanks

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: it is a mistake I made. when I solved the conflicts, I accepted my version. the problem is fixed now.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: fixed, thanks

core/src/main/kotlin/org/evomaster/core/Main.kt
ZhangMan1126: fixed

core/src/main/kotlin/org/evomaster/core/database/DbActionUtils.kt
ZhangMan1126: fixed

core/src/main/kotlin/org/evomaster/core/problem/rest/RestIndividual.kt
ZhangMan1126: fixed. thanks.

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
ZhangMan1126: changed another name and added some comments.

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceToken.kt
ZhangMan1126: fixed. the function is redundant and removed now.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampleMethodController.kt
ZhangMan1126: fixed, thanks

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampleMethodController.kt
ZhangMan1126: no. It is just enabled it when I am debugging. I have modified the code.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampler.kt
ZhangMan1126: Thanks. They are redundant code and have been removed. I am checking whether there are redundant code left.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceSampler.kt
ZhangMan1126: fixed. It is a bug, and `size` should be used when sampling the next set of actions.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestFitness.kt
ZhangMan1126: checked.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestFitness.kt
ZhangMan1126: checked and merged

core/src/main/kotlin/org/evomaster/core/problem/util/ParamUtil.kt
ZhangMan1126: this function is not used anymore. ResourceDepManageService handles the identification of params to bind the same values regarding inserted/selected row. Regarding this kind of improper implementation in other places, I tried to fix them.  Besides, unused imports and functions are also fixed.

ZhangMan1126:
hi, Andrea, the problems you mentioned are fixed. the code is ready to revise. 

Thank you!

PR #71, opened 2019-07-04
Sql null 2

arcuri82:
2 things:
1) I had issues with a git merge, so would need to double-check if I didn't screw up any existing code
2) `Ind0EMTest` fails any now and then due to constraint handling. would need to look into it


jgaleotti:
Now I remembered you told me not to actually "merge" when you made a pull request. Is this still the case, right?

arcuri82:
yep, if you accept the PR, then just merge it ;) (unless I explicitly ask for not merging it yet)

PR #76, opened 2019-08-16
Basic assertions v3

bogdanmarculescu:
Seems to work. Changes are related to:
- handling contents of arrays (especially arrays of objects).
- which strings are escaped (some strings related to the path are not escaped, as that appeared to lead to additional test failures).

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
@@ -33,7 +33,6 @@ class TestCaseWriter {
 
 
     fun convertToCompilableTestCode(
-            //format: OutputFormat,
             config: EMConfig,
             test: TestCase,
             baseUrlOfSut: String)
@@ -375,7 +374,8 @@ class TestCaseWriter {
         val verb = call.verb.name.toLowerCase()
         lines.add(".$verb(")
         if (call.locationId != null) {
-            lines.append("resolveLocation(${locationVar(call.locationId!!)}, $baseUrlOfSut + \"${applyEscapes(call.resolvedPath())}\")")
+            //lines.append("resolveLocation(${locationVar(call.locationId!!)}, $baseUrlOfSut + \"${applyEscapes(call.resolvedPath())}\")")
+            lines.append("resolveLocation(${locationVar(call.locationId!!)}, $baseUrlOfSut + \"${call.resolvedPath()}\")")
 
         } else {
 
@@ -383,13 +383,13 @@ class TestCaseWriter {
 
             if (call.path.numberOfUsableQueryParams(call.parameters) <= 1) {
                 val uri = call.path.resolve(call.parameters)
-                lines.append("\"${applyEscapes(uri)}\"")
+                lines.append("\"$uri\"")
             } else {
                 //several query parameters. lets have them one per line
                 val path = call.path.resolveOnlyPath(call.parameters)
                 val elements = call.path.resolveOnlyQuery(call.parameters)
 
-                lines.append("\"${applyEscapes(path)}?\" + ")
+                lines.append("\"$path?\" + ")
 
                 lines.indented {
                     (0 until elements.lastIndex).forEach { i -> lines.add("\"${applyEscapes(elements[i])}&\" + ") }
@@ -418,15 +418,8 @@ class TestCaseWriter {
             return "nullValue()"
         } else {
             when (resContentsItem::class) {
-                //Double::class -> return "anyOf(equalTo(${(Math.floor(resContentsItem as Double).toInt())}), closeTo(${(resContentsItem as Double)}, 0.1))"
                 Double::class -> return "numberMatches(${resContentsItem as Double})"
                 String::class -> return "containsString(\"${applyEscapes(resContentsItem as String)}\")"
-                /*String::class -> return "containsString(\"${(resContentsItem as String)
-                        .replace("\"", "\\\"")
-                        .replace("\n", "\\n")
-                        .replace("\r", "\\r")}\")"*/
-                //Note: checking a string can cause (has caused) problems due to unescaped quotation marks
-                // The above solution should be refined.
                 LinkedTreeMap::class -> return "NotCoveredYet"
                 ArrayList::class -> return "NotCoveredYet"
                 else -> return "NotCoveredYet"
@@ -453,7 +446,8 @@ class TestCaseWriter {
                 //lines.add(".body(\"sort{it.toString()}[$index].$it\" , $printableTh)") //sort and index
                 //reinstated above due to the problem of non-deterministic ordering of retrieved collections
                 // (eg. ScoutAPI - users)
-                lines.add(".body(\"get($index).$it\" , $printableTh)")
+                //lines.add(".body(\"get($index).$it\" , $printableTh)")
+                lines.add(".body(\"$it\", hasItem($printableTh))")
             }
         }
     }
--- a/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
@@ -399,8 +399,8 @@ class TestCaseWriter {
                 lines.append("\"$path?\" + ")
 
                 lines.indented {
-                    (0 until elements.lastIndex).forEach { i -> lines.add("\"${GeneUtils.applyEscapes(elements[i])}&\" + ") }
-                    lines.add("\"${GeneUtils.applyEscapes(elements.last())}\"")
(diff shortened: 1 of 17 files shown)


core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -418,15 +418,10 @@ class TestCaseWriter {
...
-                //Double::class -> return "anyOf(equalTo(${(Math.floor(resContentsItem as Double).toInt())}), closeTo(${(resContentsItem as Double)}, 0.1))"
                 Double::class -> return "numberMatches(${resContentsItem as Double})"
                 String::class -> return "containsString(\"${applyEscapes(resContentsItem as String)}\")"
-                /*String::class -> return "containsString(\"${(resContentsItem as String)
-                        .replace("\"", "\\\"")
-                        .replace("\n", "\\n")
-                        .replace("\r", "\\r")}\")"*/
-                //Note: checking a string can cause (has caused) problems due to unescaped quotation marks
-                // The above solution should be refined.
+                LinkedTreeMap::class -> return "NotCoveredYet"
```
arcuri82: do not use "magic" strings, but rather put them in a field (ie, don't copy&paste the same string several times). especially if you have other function that the use the value "NotCoveredYet" in its if statements

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -455,40 +469,35 @@ class TestCaseWriter {
...
-                                val printableTh = handleFieldValues(value)
-                                if (printableTh != "null"
-                                        && printableTh != "NotCoveredYet"
-                                        && !printableTh.contains("logged")
-                                ) {
-                                    lines.add(".body(\"get($test_i)\", $printableTh)")
+                        //resContents.sortBy { it.toString() }
+                        //assertions on contents
+                        if(resContents.size > 0){
+                            if(resContents.first()::class == LinkedTreeMap::class) resContents.sortBy { it.toString() }
```
arcuri82: this would break as soon as Gson will use something different than a LinkedTreeMap, isn't it? i think it would be better to rather check if the class is implementing "Map"

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -455,40 +469,35 @@ class TestCaseWriter {
...
+                            // Sorting needed as sometimes retrieving collections results in non-deterministic order
+                            // (eg ScoutAPI - users).
```
arcuri82: even if we sort it here when we generate the JUnit, how can guarantee to have same order when running the generated tests?

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -509,6 +518,23 @@ class TestCaseWriter {
         //handleExpectations(res, lines, true)
     }
 
+    private fun addObjectAssertions(resContents: LinkedTreeMap<*,*>, lines: Lines){
+        resContents.keys
+                .filter{ !(it as String).contains("timestamp")}
```
arcuri82: are u assuming the key is always a String?
could also add a comment specifying we want to avoid assertion on time-based fields, as those would lead to flaky tests

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -668,12 +694,26 @@ class TestCaseWriter {
...
-        val ret = string.replace("\"", "\\\"")
+
+        val timeRegEx = "[0-2]?[0-9]:[0-5][0-9]".toRegex()
+
+        val ret = string.split("@")[0] //first split off any reference that might differ between runs
+                .split(timeRegEx)[0] //split off anything after specific timestamps that might differ
+                .replace("""\\""", """\\\\""")
+                .replace("\"", "\\\"")
                 .replace("\n", "\\n")
                 .replace("\r", "\\r")
```
arcuri82: would really need some more test cases to check if generated tests can be correctly compiled

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -63,11 +62,11 @@ class TestCaseWriter {
...
                 //usedObjects = test.test.individual.usedObjects
 
-                if (config.enableCompleteObjects) {
-                    usedObjects = test.test.individual.usedObjects
+                if (config.enableCompleteObjects && test.test.individual is RestIndividual) {
+                    usedObjects = (test.test.individual as RestIndividual).usedObjects
                 }
 
-                handleDbInitialization(format, test.test.individual.dbInitialization, lines)
+                handleDbInitialization(format, (test.test.individual as RestIndividual).dbInitialization, lines)
```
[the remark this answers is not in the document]
bogdanmarculescu: The line should be in an if statement block (line 61)- "if (test.test.individual is RestIndividual)" that ensures that. This section also takes care of making used objects available (now that I think of it, I will remove the check from line 64 for the same reason).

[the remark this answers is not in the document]
arcuri82: ok

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: Change to Map.

I was mostly concerned about using methods specific to LinkedTreeMap, but since I'm not using any, Map should work the same.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: Done.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: At the moment, I am doing a basic lookup (as we discussed today). 

The assertion is that in the collection retrieved, there is one element that has a particular value. (e.g. for the field "id", the collection x.id contains one matching element). 

I will write up a more detailed comment on this, as the current approach could be improved. It should, however, work for basic assertions.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: For the moment, I am making that assumption, yes. The solution here is a bit of a hack to get it up and running. A more permanent (and hopefully elegant) solution would be to handle this via the flaky test handling process (when that is ready).

I have added a comment to this effect as well.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: Agreed, I just haven't gotten around to that yet. Depending on how we will handle the escapes, it may require some more refining as well.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -24,6 +24,9 @@ class TestCaseWriter {
 
     private var counter = 0
     private var usedObjects = UsedObjects()
+    private var previous_chained = false
+    private var previous_id = ""
+    private var chained = false
```
arcuri82: better to call them `previousChained` and `previousId`. `_` should rather be used for constants in uppercase, eg `PREVIOUS_ID`

bogdanmarculescu: Done.

bogdanmarculescu:
Updated and turned basic assertions on by default. 
The verify appears to work locally. 
I don't think the failures are related to escapes, but I will fix any additional problems as I see them (or are pointed out to me).

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
arcuri82: after the discussion we had, shouldn't we remove this sorting here?

bogdanmarculescu:
Yep. The sorting was a leftover from a previous version. Removed it.

arcuri82:
all the CI pipelines are failing.
looks like the new tests are failing on CI, eg from logs 
```
[ERROR]   EscapeManualTest.testContainsQuote2:80 1 expectation failed.
Expected status code <200> but was <400>.
```

bogdanmarculescu:
Yes. I tried to add a test (and I think I messed things up). I'm working on fixing it now. The test in question works in isolation, but fails in maven. Trying to address this now.

PR #82, opened 2019-09-10
modify the solution to save mutated genes which will be used by archi…

man-zhang:
modify the solution to save mutated genes which will be used by archive-mutation
modify the resource tests in order to tests whether the mutated info is correct to be saved.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceMutator.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceMutator.kt
@@ -50,7 +50,10 @@ class ResourceRestMutator : StandardMutator<RestIndividual>() {
     }
 
     override fun update(previous: EvaluatedIndividual<RestIndividual>, mutated: EvaluatedIndividual<RestIndividual>, mutatedGenes : MutatedGeneSpecification?) {
-        //update resource dependency after mutating structure of the resource-based individual
+        /*
+            update resource dependency after mutating structure of the resource-based individual
+            NOTE THAT [this] can be only applied with MIO. In MIO, [mutatedGenes] must not be null.
+         */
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceMutator.kt
```diff
@@ -48,9 +49,9 @@ class ResourceRestMutator : StandardMutator<RestIndividual>() {
         return individual.getResourceCalls().flatMap { it.seeGenes() }.filter(Gene::isMutable)
     }
 
-    override fun update(previous: EvaluatedIndividual<RestIndividual>, mutated: EvaluatedIndividual<RestIndividual>, mutatedGenes : MutableList<Gene>) {
+    override fun update(previous: EvaluatedIndividual<RestIndividual>, mutated: EvaluatedIndividual<RestIndividual>, mutatedGenes : MutatedGeneSpecification?) {
         //update resource dependency after mutating structure of the resource-based individual
-        if(mutatedGenes.isEmpty() && (previous.individual.getResourceCalls().size > 1 || mutated.individual.getResourceCalls().size > 1) && config.probOfEnablingResourceDependencyHeuristics > 0){
+        if(mutatedGenes!!.mutatedGenes.isEmpty() && (previous.individual.getResourceCalls().size > 1 || mutated.individual.getResourceCalls().size > 1) && config.probOfEnablingResourceDependencyHeuristics > 0){
```
arcuri82: don't understand this: in the function input, you marked it nullable with "?", but here you say it is never null with "!!"

ZhangMan1126: RestResourceMutator can be only applied with MIO. In MIO, the input is not null by default.

arcuri82: in that case, can you add such comment there in the code?

PR #84, opened 2019-09-11
tracking and impact handling

ZhangMan1126:
update tracking
update impact handling


core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -69,12 +72,29 @@ abstract class Gene(var name: String) {
...
+                             targets: Set<Int>){
+        standardMutation(randomness, apc, allGenes)
+//        if (!archiveMutator.enableArchiveMutation()){
+//            standardMutation(randomness,apc, allGenes)
+//        }else{
+//            if (archiveMutator.doesSupport(this))
+//                archiveMutator.mutate(this)
+//            else
+//                TODO("ArchiveGeneMutation for ${this::class.java.simpleName} was NOT IMPLEMENT")
+//        }
```
arcuri82: i guess this commented code could be removed?
plus could add the comment: "if this method is not overridden, just default to standard mutation"

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -129,4 +149,12 @@ abstract class Gene(var name: String) {
      */
     abstract fun containsSameValueAs(other: Gene): Boolean
 
+    /**
+     * indicates if the gene reaches its optimal value.
+     */
+    open fun reachOptimal() = false
```
arcuri82: I need more explanation here...
How can a gene be "optimal"?
the optimality depends on testing target, and a gene could involved in the reaching of hundreds of targets (if not even more)

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -129,4 +149,12 @@ abstract class Gene(var name: String) {
...
+
+    open fun archiveMutationUpdate(original: Gene, mutated: Gene, doesCurrentBetter: Boolean, archiveMutator: ArchiveMutator){
```
arcuri82: provide docs for this method

core/src/main/kotlin/org/evomaster/core/search/gene/MapGene.kt
```diff
@@ -100,4 +112,72 @@ class MapGene<T>(
...
+            randomness.nextBoolean(MODIFY_SIZE)
+        }
+        if (fmodifySize){
+            val p = randomness.nextBoolean()
+            add = add || p
+            delete = delete || !p
+        }
+
+        if (add && add == delete)
+            log.warn("add and delete an element cannot happen in a mutation, and size of elements: {} and maxSize: {}", elements.size, maxSize)
```
arcuri82: please put parentheses. When I read code like that, I am never sure if it is `(add && add) == delete`, or `add && (add == delete)`

core/src/main/kotlin/org/evomaster/core/search/gene/OptionalGene.kt
```diff
@@ -56,14 +69,63 @@ class OptionalGene(name: String,
...
+            standardMutation(randomness, apc, allGenes)
+            return
+        }
+
+        val preferActive = if (!archiveMutator.applyArchiveSelection() || impact == null || impact !is OptionalGeneImpact) true
+        else {
+            !impact.activeImpact.run {
+                this.timesToManipulate > 5
+                        &&
+                        (this._false.timesOfImpact.filter { targets.contains(it.key) }.map { it.value }.max()?:0) > ((this._true.timesOfImpact.filter { targets.contains(it.key) }.map { it.value }.max()?:0) * 1.5)
```
arcuri82: consider previous comments on `_` as prefix in variable names

core/src/main/kotlin/org/evomaster/core/search/gene/StringGene.kt
```diff
@@ -408,4 +451,178 @@ class StringGene(
...
+
+    override fun archiveMutationUpdate(original: Gene, mutated: Gene, doesCurrentBetter: Boolean, archiveMutator: ArchiveMutator) {
+        if (!archiveMutator.enableArchiveGeneMutation()) return
+
+        original as? StringGene ?: throw IllegalStateException("$original should be StringGene")
+        mutated as? StringGene ?: throw IllegalStateException("$mutated should be StringGene")
+
+        if (this != mutated){
+            dependencyInfo.mutatedtimes +=1
+            if (this.mutatedIndex == -2){
```
arcuri82: shouldn't rather use `NEVER_ARCHIVE_MUTATION` here?

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlNullable.kt
```diff
@@ -49,6 +63,76 @@ class SqlNullable(name: String,
...
+    override fun archiveMutation(randomness: Randomness, allGenes: List<Gene>, apc: AdaptiveParameterControl, selection: GeneMutationSelectionMethod, impact: GeneImpact?, geneReference: String, archiveMutator: ArchiveMutator, evi: EvaluatedIndividual<*>, targets: Set<Int>) {
+        if(!archiveMutator.enableArchiveMutation()){
+            standardMutation(randomness, apc, allGenes)
+            return
+        }
+
+        val preferPresent = if (!archiveMutator.applyArchiveSelection() || impact == null || impact !is SqlNullableImpact) true
+                    else {
+            !impact.presentImpact.run {
+                this.timesToManipulate > 5
```
arcuri82: this seems repeated at least twice. should put 5 in a meaningfully named variable, or at least provide some comments

core/src/main/kotlin/org/evomaster/core/search/impact/ImpactUtils.kt
```diff
@@ -0,0 +1,179 @@
...
+            individual.seeInitializingActions().forEachIndexed { index, action ->
+                action.seeGenes().filter { mutatedGenes.contains(it) }.forEach { g->
+                    val id = generateGeneId(action, g)
+                    val contexts = mutatedGenesWithContext.getOrPut(id){ mutableListOf()}
+                    val previous = findGeneById(previousIndividual, id, action.getName(), index, true)
+                    contexts.add(MutatedGeneWithContext(g, action.getName(), index, previous))
+                }
+            }
+
+            assert(mutatedGenesWithContext.values.sumBy { it.size } == mutatedGenes.size)
```
arcuri82: as u are iterating over a collection to sum values, it would be better to put this in a `Lazy.assert`.

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -35,7 +35,7 @@ class Statistics : SearchListener {
     @Inject(optional = true)
     private var sampler: Sampler<*>? = null
 
-    @Inject(optional = true)
+    //@Inject(optional = true)
     private var remoteController: RemoteController? = null
```
arcuri82: we had a discussion about this at a certain point, but I don't remember :(  can u please remind me what it was? (maybe I was supposed to do some fixes, and then I forgot?)

core/src/main/kotlin/org/evomaster/core/search/service/mutator/geneMutation/MutationBoundaryUpdate.kt
```diff
@@ -0,0 +1,46 @@
+package org.evomaster.core.search.service.mutator.geneMutation
+
+/**
+ * created by manzh on 2019-09-19
+ */
+abstract class MutationBoundaryUpdate<T> (var preferMin : T, var preferMax : T, var counter : Int = 0, var reached : Boolean = false) where T : Number{
+
+    abstract fun copy() : MutationBoundaryUpdate<T>
+
+    fun setCounter(improved : Boolean){
```
arcuri82: methods that starts with `set` and `get` have very special meaning in Kotlin (eg, they will create properties). as here you are not setting the counter, should use a different name, eg something like `updateCounter`

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
ZhangMan1126: hi Andrea. the problem is that when exporting statistics info with `Statistics`, `remoteController` is always Instantiated. But `remoteController` may not exist, in this case, a null pointer exception will be thrown by `val unitsInfo = remoteController?.getSutInfo()?.unitsInfoDto`

PR #90, opened 2019-09-18
copyright clarifications

arcuri82:


README.md
```diff
@@ -267,7 +267,9 @@ e.g., details on the Many Independent Algorithm (MIO).
 
 ### License
 _EvoMaster_'s source code is released under the LGPL (v3) license.
-
+For a list of the used third-party libraries, you can directly see the root [pom.xml](./pom.xml) file.
+For a list of code directly imported (and then possibly modified/updated) from 
+other open-source projects, see [here](./docs/reused_code.md).
 
 
 ### ![](https://www.yourkit.com/images/yklogo.png)
```
jgaleotti: One of the libraries used by the regex distance code has BSD license, would that be compatible?

docs/for_developers.md
```diff
@@ -236,6 +236,25 @@ in submodules can lead to duplicated `<version>` declarations with different ver
 All version numbers should be easily audited, and so should be in a single file (i.e., the
 *root* `pom.xml`).
 
+
+### THIRD-PARTY LIBRARIES
```
jgaleotti: One of the libraries used by the regex distance code has BSD license, would that be compatible?

docs/reused_code.md
```diff
@@ -0,0 +1,19 @@
...
+* _RegexEcma262.g4_: from [Antlr examples](https://github.com/antlr/grammars-v4/blob/master/ecmascript/ECMAScript.g4).
+  Released under MIT license.
+  
+* _CustomSummaryGeneratingListener.java_ and _CustomMutableTestExecutionSummary.java_:
+  from JUnit 5.
+  Released under Eclipse Public License.  
+  These classes will be removed when this [PR](https://github.com/junit-team/junit5/issues/1947)
+  will be part of the next JUnit release.  
+  
+
```
jgaleotti: Understood, I will add any class added to the code directly here once this PR is merged into master.

docs/for_developers.md
arcuri82: As far as I know, only GPL is a problem. other OS licenses are usually fine

docs/reused_code.md
arcuri82: if you approve these clarifications, and require no further info, you can just go ahead and merge ;)

PR #91, opened 2019-09-19
Sorting update

bogdanmarculescu:
Custom sorting and naming activated by default.
Various other minor tweaks included.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
@@ -150,7 +150,7 @@ class SortingHelper {
      */
     private val dbInitSize: Comparator<EvaluatedIndividual<*>> = compareBy<EvaluatedIndividual<*>>{ ind ->
         if(ind.individual is RestIndividual) {
-            (ind.individual as RestIndividual).dbInitialization.size
+            ind.individual.dbInitialization.size
         }
         else 0
     }.reversed()
@@ -169,7 +169,7 @@ class SortingHelper {
      */
 
     //var comparatorList = mutableListOf(statusCode, minActions)
-    var comparatorList = mutableListOf(coveredTargets)
+    var comparatorList = mutableListOf(statusCode, coveredTargets)
 
     private val availableSortCriteria = mutableListOf(statusCode, minActions, maxStatusCode, maxActions, dbInitSize, coveredTargets)
 
--- a/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
@@ -6,6 +6,8 @@ import org.evomaster.core.problem.rest.RestCallResult
 import org.evomaster.core.problem.rest.RestIndividual
 import org.evomaster.core.search.EvaluatedIndividual
 import org.evomaster.core.search.Solution
+import kotlin.reflect.KFunction
+import kotlin.reflect.KFunction1
 
 /**
  * This class is responsible to decide the order in which
@@ -19,13 +21,14 @@ import org.evomaster.core.search.Solution
  */
 class TestSuiteOrganizer {
     companion object {
-        val sortingHelper = SortingHelper()
-        val namingHelper = NamingHelper()
+        private val sortingHelper = SortingHelper()
+        private val namingHelper = NamingHelper()
 
+        private val default_sorting = mutableListOf(0, 1)
 
         fun sortTests(solution: Solution<*>, customNaming: Boolean = false): List<TestCase> {
-
-            //TODO here in the future we will have something bit smarter
+            sortingHelper.selectCriteriaByIndex(default_sorting)
+            //TODO here in the future we will have something a bit smarter
             return sortingHelper.sort(solution, namingHelper, customNaming)
         }
     }
@@ -70,21 +73,48 @@ class NamingHelper {
         else return ""
     }
 
-    val namingCriteria =  mutableListOf(::criterion1_500, ::criterion3_sampling)
+    private var namingCriteria =  mutableListOf(::criterion1_500, ::criterion3_sampling)
+    private val availableCriteria = mutableListOf(::criterion1_500, ::criterion2_hasPost, ::criterion3_sampling, ::criterion4_dbInit)
 
 
     fun suggestName(individual: EvaluatedIndividual<*>): String{
         return namingCriteria.map { it(individual) }.joinToString("")
     }
 
+    fun getAvailableCriteria(): MutableList<KFunction1<EvaluatedIndividual<*>, String>> {
+        return availableCriteria
+    }
+
+    fun selectCriteria(selected: MutableList<KFunction1<EvaluatedIndividual<*>, String>>){
+        if (availableCriteria.containsAll(selected)){
+            namingCriteria = availableCriteria
+        }
+        else {
+            throw UnsupportedOperationException("The naming criteria chosen appear to not be supported at the moment.")
+        }
+    }
+
+    fun selectCriteriaByIndex(selected: MutableList<Int>){
+        if (availableCriteria.indices.toMutableList().containsAll(selected)){
+            namingCriteria = availableCriteria.filterIndexed{ index, _ ->
+                selected.contains(index)
+            } as MutableList<KFunction1<EvaluatedIndividual<*>, String>>
+        }
+        else {
+            throw UnsupportedOperationException("The naming criteria chosen appear to not be supported at the moment.")
+        }
+    }
+
+
 }
 
 
 class SortingHelper {
-    /** [maxStatusCodeComparatorInd] sorts Evaluated individuals based on the highest status code (e.g., 500s are first).
+
+    /** [maxStatusCode] sorts Evaluated individuals based on the highest status code (e.g., 500s are first).
      *
      * **/
-    val maxStatusCodeComparatorInd = compareBy<EvaluatedIndividual<*>>{ind ->
+    private val maxStatusCode: Comparator<EvaluatedIndividual<*>> = compareBy<EvaluatedIndividual<*>>{ ind ->
         val max = ind.results.filterIsInstance<RestCallResult>().maxBy { it.getStatusCode()!! }
             (max as RestCallResult).getStatusCode() ?: 0
     }.reversed()
@@ -96,20 +126,20 @@ class SortingHelper {
      *          - third:    4xx
      */
 
-    val statusCode = compareBy<EvaluatedIndividual<*>>{ind ->
+    private val statusCode: Comparator<EvaluatedIndividual<*>> = compareBy { ind ->
         val max = ind.results.filterIsInstance<RestCallResult>().maxBy { it.getStatusCode()!! }
         ((max as RestCallResult).getStatusCode()?.rem(500)) ?: 0
     }
 
-    /** [maxNumberOfActionsComparatorInd] sorts Evaluated individuals based on the number of actions (most actions first).
+    /** [maxActions] sorts Evaluated individuals based on the number of actions (most actions first).
      */
-    val maxNumberOfActionsComparatorInd = compareBy<EvaluatedIndividual<*>>{ ind ->
+    private val maxActions: Comparator<EvaluatedIndividual<*>> = compareBy<EvaluatedIndividual<*>>{ ind ->
         ind.individual.seeActions().size
     }.reversed()
 
     /** [minActions] sorts Evaluated individuals based on the number of actions (most actions first).
      */
-    val minActions = compareBy<EvaluatedIndividual<*>>{ ind ->
+    private val minActions: Comparator<EvaluatedIndividual<*>> = compareBy { ind ->
         ind.individual.seeActions().size
     }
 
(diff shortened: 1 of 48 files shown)


core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -138,7 +138,7 @@ object GeneUtils {
 
      */
 
-    fun applyEscapes(string: String, mode: String = "none", format: OutputFormat = OutputFormat.KOTLIN_JUNIT_5): String{
+    fun applyEscapes(string: String, mode: String = "text", format: OutputFormat = OutputFormat.KOTLIN_JUNIT_5): String{
```
arcuri82: i think the `mode` should be refactored, and rather use an enum, if possible

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -153,48 +153,48 @@ object GeneUtils {
...
                 .replace("\t", "\\t")
 
+        when{
+            format.isKotlin() -> return ret.replace("\$", "\\\$")
+            else -> return ret
+        }
+
     }
 
     fun applyUriEscapes(string: String, format: OutputFormat = OutputFormat.JAVA_JUNIT_4):String{
```
arcuri82: I m bit unsure of this method. there is much more that need encoding, and some of it is already handled when evaluating the fitness function.
look at  `RestPath.encode`, `RestPath.resolveOnlyPath` and `AbstractRestFitness.handleRestCall`. In particular, in this latterthe following will likely need to be removed / updated:

```
 .let {
                    /*
                        TODO this will be need to be done properly, and check if
                        it is or not a valid char.
                        Furthermore, likely needed to be done in resolveLocation,
                        or at least check how RestAssured would behave
                     */
                    it.replace("\"", "")
                }
```

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -70,39 +73,84 @@ class NamingHelper {
...
+    fun selectCriteria(selected: MutableList<KFunction1<EvaluatedIndividual<*>, String>>){
+        if (availableCriteria.containsAll(selected)){
+            namingCriteria = availableCriteria
+        }
+        else {
+            throw UnsupportedOperationException("The naming criteria chosen appear to not be supported at the moment.")
+        }
+    }
+
+    fun selectCriteriaByIndex(selected: MutableList<Int>){
```
arcuri82: you are not modifying such input list, so why passing it as `MutableList` instead of a `List`?

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -111,46 +159,81 @@ class SortingHelper {
...
-                              comparators: MutableList<Comparator<EvaluatedIndividual<*>>> = mutableListOf(maxStatusCodeComparatorInd)
+                              comparators: MutableList<Comparator<EvaluatedIndividual<*>>> = mutableListOf(statusCode)
 
     ): List<TestCase> {
         var counter = 0
 
-        comparators.forEach { solution.individuals.sortWith(it) }
+        val inds = solution.individuals
+
+        comparators.asReversed().forEach {
```
arcuri82: bit confused... wouldn't the last sorting override/change all the previous sortings? if not, add a comment to clarify it

arcuri82:
Hi,
 as a rule of thumb, we should prefer immutable/read-only collections (eg, `List` in Kotlin) over mutable ones (e.g., `MutableList`)

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -96,6 +96,7 @@ class NamingHelper {
 
     fun selectCriteriaByIndex(selected: MutableList<Int>){
         if (availableCriteria.indices.toMutableList().containsAll(selected)){
+            for (i in selected)
```
arcuri82: why `indices` was transformed into a mutable list?

core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
arcuri82: ok. just make sure to have a test case checking this behavior

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -13,6 +13,15 @@ object GeneUtils {
      */
     private val intpow2 = (0..30).map { 2.0.pow(it).toInt() }
 
+    enum class EscapeMode {
```
arcuri82: add a comment clarifying what this enum is for

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -212,9 +222,21 @@ object GeneUtils {
...
 
-        if (format.isKotlin()) return ret.replace("\$", "\${\'\$\'}")
+        if (format.isKotlin())  return ret.replace("\$", "\${\'\$\'}")
         //ret.replace("\$", "\\\$")
         else return ret
     }
 
-}
\ No newline at end of file
+    fun getMode(mode: String?): EscapeMode{
```
arcuri82: can't this be implemented with just `EscapeMode.valueOf(mode)` ?

bogdanmarculescu: It could. But as far as I understand it, a value that does not match anything will cause an exception to be thrown. I don't think I covered all the possible modes (and I am not sure I want to enforce this strictly). 

Currently, if the mode is something not supported (i.e. EscapeMode.NONE), then the raw string is returned, rather than crashing the system. 

Let me know if you want a different kind of behaviour for this.

arcuri82: as the escape mode is not some inputs from the user, but rather an internal detail of EM, then we need to support all modes that we enumerate.

arcuri82:
Hi, 
 please merge my changes of today, and check if the 3 CIs would still fail

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -13,6 +13,26 @@ object GeneUtils {
...
+         *
+         * Different purposes require different modes of escape (e.g. URI may require percent encoding). This is to
+         * keep track of what modes are supported and how they map to the respective implementations.
+         *
+         * Any mode that is not supported will go under NONE, and will result in no escapes being applied at all. The
+         * purpose is to ensure that, even if the mode being used is unsupported, the system will not throw an exception.
+         * It may not behave as desired, but it should not crash.
+         *
+         */
+        URI,
```
arcuri82: the `/** */` creates JavaDocs for the item following. I think such comment should be applied on the the enum `EscapeMode`, and not to its value `URI`

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -138,50 +158,63 @@ object GeneUtils {
...
-        else string
-        return ret.replace("\\", "%5C")
+    fun applyUriEscapes(string: String, format: OutputFormat):String{
+        val ret = string.replace("\\", "%5C")
+                .replace("\"", "%22")
+                .replace("\n", "%0A")
+
+        if(format.isKotlin()) return ret.replace("\$", "%24")
+        else return ret
     }
```
arcuri82: as a I said before, why not using the existing functions to escape URIs?

core/src/test/kotlin/org/evomaster/core/search/gene/GeneUtilsEscapeTest.kt
```diff
@@ -10,28 +11,36 @@ internal class GeneUtilsEscapeTest {
...
+        println("TestJson Java: ${testJson.getValueAsPrintableString(mode = "json", targetFormat = formatKotlin)}")
 
-        assertTrue(GeneUtils.applyEscapes(testJson, "json", formatKotlin).contains("\\"))
-        assertTrue(GeneUtils.applyEscapes(testJson, "json", formatJava5).contains("\\"))
+        //assertTrue(testJson.getValueAsPrintableString(mode = "json", targetFormat = formatKotlin).contains("\\"))
+        //assertTrue(testJson.getValueAsPrintableString(mode = "json", targetFormat = formatKotlin).contains("\\"))
 
-        assertTrue(GeneUtils.applyEscapes(testJson, "json", formatKotlin).contains("\""))
-        assertTrue(GeneUtils.applyEscapes(testJson, "json", formatKotlin).contains("\$"))
+        assertTrue(testJson.getValueAsPrintableString(mode = "json", targetFormat = formatKotlin).contains("\""))
```
arcuri82: isn't `mode` an enum now instead of a string?

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
```diff
@@ -0,0 +1,67 @@
...
+                            assertTrue(previous.getName().contains("400") || previous.getName().contains("500"));
+                        }
+
+                        //Check that neighbouring TC with the same code are sorted by number of actions
+                        if((current.getName().contains("500") && previous.getName().contains("500")) ||
+                                (current.getName().contains("400") && previous.getName().contains("400")) ||
+                                (current.getName().contains("200") && previous.getName().contains("200"))
+                        ) {
+                            assertTrue(current.getTest().evaluatedActions().size() >= previous.getTest().evaluatedActions().size());
+                        }
```
arcuri82: besides the ordering, shouldn't we assert that there is at least one 500, at least one 200 and at least one 400?

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
```diff
@@ -0,0 +1,67 @@
...
+                    Iterator<TestCase> iterator = tclist.iterator();
+                    TestCase current, previous = iterator.next();
+                    while(iterator.hasNext()){
+                        current = iterator.next();
+                        // Check that a TC with 500 in the name does not follow a TC without a 500 in the name (500 should be first).
+                        if(current.getName().contains("500")){
+                            assertTrue(previous.getName().contains("500"));
+                        }
+                        // Check that a TC with 400 in the name is preceded by another 400 or by a 500.
+                        if(current.getName().contains("400")){
```
arcuri82: by default, 4xx is less important than 2xx codes. 2xx should come before the 4xx in the sorted tests

core/src/test/kotlin/org/evomaster/core/search/gene/GeneUtilsEscapeTest.kt
bogdanmarculescu: I didn't refactor the Gene (and its methods) to take an EscapeMode instead of a string, so as not to damage other bits of the code. I can do that refactoring now.

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
bogdanmarculescu: Yes. The code was right, but the test code was wrong.

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
bogdanmarculescu: I can add that, if needed. I'm using NRTestBase, so the SUT is being tested separately (as part of NREMTest). Any assertions that pass in that test should also pass in this one.

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
bogdanmarculescu: I tried that and got some trouble, but it may be due to another bug overlapping it. I'll have another look at it.

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
arcuri82: yes, but NREMTest has checks only on 2xx and 4xx, not 5xx.

e2e-tests/spring-examples/src/test/java/org/evomaster/e2etests/spring/examples/sort/SortEMTest.java
arcuri82: but if the test was wrong, why didn't it fail?

bogdanmarculescu: I was daft. The test code only looked at the name of the test cases (and, naturally, most info is not there). 

I am now fixing both the test and the sorting (since the updated test helped me find a bug in the sorting part).

PR #94, opened 2019-09-23
distance to regular expression

jgaleotti:
Implementation of TT for Java Regular Expressions

### DIFF
--- a/client-java/instrumentation/pom.xml
+++ b/client-java/instrumentation/pom.xml
@@ -47,6 +47,16 @@
             <artifactId>mockito-core</artifactId>
         </dependency>
 
+        <!-- Dependencies for Java Regex distances -->
+        <dependency>
+            <groupId>net.sf.jgrapht</groupId>
+            <artifactId>jgrapht</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>dk.brics.automaton</groupId>
+            <artifactId>automaton</artifactId>
+        </dependency>
+
     </dependencies>
 
     <build>
--- a/client-java/instrumentation/pom.xml
+++ b/client-java/instrumentation/pom.xml
@@ -99,6 +99,14 @@
                                     <pattern>org.objectweb</pattern>
                                     <shadedPattern>shaded.org.objectweb</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>org.jgrapht</pattern>
+                                    <shadedPattern>shaded.org.jgrapht</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>dk.brics.automaton</pattern>
+                                    <shadedPattern>shaded.dk.brics.automaton</shadedPattern>
+                                </relocation>
                             </relocations>
                         </configuration>
                     </execution>
--- a/core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
@@ -4,7 +4,6 @@ import org.evomaster.core.output.OutputFormat
 import org.evomaster.core.search.gene.Gene
 import org.evomaster.core.search.service.AdaptiveParameterControl
 import org.evomaster.core.search.service.Randomness
-import java.lang.IllegalStateException
 
 /*
 \w	Find a word character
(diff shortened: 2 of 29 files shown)


client-java/instrumentation/pom.xml
```diff
@@ -47,6 +47,16 @@
...
+        <!-- Dependencies for Java Regex distances -->
+        <dependency>
+            <groupId>net.sf.jgrapht</groupId>
+            <artifactId>jgrapht</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>dk.brics.automaton</groupId>
+            <artifactId>automaton</artifactId>
+        </dependency>
+
```
arcuri82: these libraries should be added as well to the shade plugin

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
```diff
@@ -44,9 +43,14 @@ class CharacterClassEscapeRxGene(
     }
 
     override fun standardMutation(randomness: Randomness, apc: AdaptiveParameterControl, allGenes: List<Gene>) {
+        if (value=="") {
+            // if standardMutation was invoked before calling to randomize
+            // then do a randomize first
+            randomize(randomness, forceNewValue = true, allGenes = allGenes)
+        }
```
arcuri82: i think this should be removed. if it does happen that standard mutation is called before a randomized, then it is a bug. we should always randomize before an individual can be mutated. so we should rather investigate why it was not the case (ie bug)

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
```diff
@@ -44,9 +43,14 @@ class CharacterClassEscapeRxGene(
...
 
         value = when(type){
-            "d" -> ((value.toInt() + randomness.choose(listOf(1,-1) + 10)) % 10).toString()
+            "d" -> ((Math.abs(value.toInt() + randomness.choose(listOf(1,-1) + 10))) % 10).toString()
```
arcuri82: why was `abs` necessary? can it ever happen that it is negative?

client-java/instrumentation/pom.xml
jgaleotti: ok

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
jgaleotti: yes. I found cases where the value is set to 0 and the randomness chose -1.
The result of -1 % 10 is -1.

arcuri82: yes, but that is the reason for why there is a `+ 10`, so a `0` on which `-1` is applied become a `9` :)

jgaleotti: I wrote the code to:
            "d" -> ((value.toInt() + randomness.choose(listOf(1,-1)) + 10) % 10).toString()

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
jgaleotti: Indeed. DIsjunctionListRxGene.randomize() can leave un-initialised the inactive disjunction. Due to this, if standardMutation() changes the activeDisjunction, the gene can be left without randomising.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterClassEscapeRxGene.kt
```diff
@@ -44,9 +43,14 @@ class CharacterClassEscapeRxGene(
...
     override fun standardMutation(randomness: Randomness, apc: AdaptiveParameterControl, allGenes: List<Gene>) {
+        if (value=="") {
+            // if standardMutation was invoked before calling to randomize
+            // then we signal an exception
+            throw IllegalStateException("Cannot apply mutation on an uninitalized gene")
+        }
 
         value = when(type){
-            "d" -> ((value.toInt() + randomness.choose(listOf(1,-1) + 10)) % 10).toString()
+            "d" -> ((value.toInt() + randomness.choose(listOf(1,-1)) + 10) % 10).toString()
```
arcuri82: indeed, it was a bug! thx for fixing it ;)

PR #96, opened 2019-09-26
SqlTimestampGene is replaced with DateTimeGene

jgaleotti:


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/DateTimeParsingUtils.java
```diff
@@ -212,4 +214,86 @@ public static double getDistanceToISOLocalDate(CharSequence input) {
...
+            return H_REACHED_BUT_NULL;
+        }
+
+        try {
+            new SimpleDateFormat("YYYY-MM-DD HH:MM").parse(input.toString());
+            /*
+                due to the simplification later on, still must make
+                sure to get a 1 if no exception is thrown
+             */
+            return 1d;
```
arcuri82: if this is a "distance", I guess that we should rather return a 0? 
we used 1 when dealing with satisfied heuristic (ie covered target), not for distances

core/src/main/kotlin/org/evomaster/core/search/gene/DateTimeGene.kt
```diff
@@ -44,9 +53,22 @@ open class DateTimeGene(
...
+        val formattedDate = GeneUtils.let {
+            "${it.padded(date.year.value, 4)}-${it.padded(date.month.value, 2)}-${it.padded(date.day.value, 2)}"
+        }
+        val formattedTime = GeneUtils.let {
+            "${it.padded(time.hour.value, 2)}:${it.padded(time.minute.value, 2)}:${it.padded(time.second.value, 2)}"
+        }
+        return when (dateTimeGeneFormat) {
+            DateTimeGeneFormat.ISO_LOCAL_DATE_TIME_FORMAT -> {
+                "${formattedDate}" + "T" + "${formattedTime}"
+            }
```
arcuri82: why not simplifying it with `"${formattedDate}T${formattedTime}"`?

core/src/main/kotlin/org/evomaster/core/search/gene/DateTimeGene.kt
```diff
@@ -44,9 +53,22 @@ open class DateTimeGene(
...
+
+            DateTimeGeneFormat.DEFAULT_DATE_TIME -> {
+                "${formattedDate}" + " " + "${formattedTime}"
+            }
```
arcuri82: see comment above

core/src/test/kotlin/org/evomaster/core/database/SqlInsertBuilderTest.kt
```diff
@@ -184,7 +184,7 @@ class SqlInsertBuilderTest {
 
         val genes = actions[0].seeGenes()
         assertEquals(1, genes.size)
-        assertTrue(genes[0] is SqlTimestampGene)
+        assertTrue(genes[0] is DateTimeGene)
```
arcuri82: shouldn't also add a check on the format of the date gene being the one expected for SQL?

core/src/main/kotlin/org/evomaster/core/search/gene/DateTimeGene.kt
jgaleotti: agree

core/src/main/kotlin/org/evomaster/core/search/gene/DateTimeGene.kt
jgaleotti: agree

core/src/test/kotlin/org/evomaster/core/database/SqlInsertBuilderTest.kt
jgaleotti: agree. Added more assertions to check all the expected behaviour of such dateTime genes

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/DateTimeParsingUtils.java
jgaleotti: you're right. I did not have a test for that case.

jgaleotti: There was a comment in the code that states the following:        
//recall h in [0,1] where the highest the distance the closer to 0
So, this menas that (according to this comment)
- h should be 1 when the value is successfully parsed
- h should be 0 when the distance to successfully parsing the input

jgaleotti: I think I understand your point now. Distance should be 0 closer to your target, >0 further away from your target. But the "h" does not mean "distance", but rather "heuristic". So, I will rename all method names in this class to explicitly say that we are returning a "heuristic" value in [0,1], not a "distance".

jgaleotti:
Hi. I applied your requests and pushed the changes.

PR #106, opened 2019-10-15
@Probability constraints instead of @Min(0) @Max(1)

arcuri82:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -698,32 +716,32 @@ class EMConfig {
...
     @Experimental
     @Cfg("Specify a probability to enable archive-based mutation")
-    @Min(0.0) @Max(1.0)
+    @Probability
     var probOfArchiveMutation = 0.0
 
     @Experimental
     @Cfg("Specify a percentage which is used by archived-based gene selection method (e.g., APPROACH_GOOD) for selecting top percent of genes as potential candidates to mutate")
-    @Min(0.0) @Max(1.0)
+    @Probability
```
ZhangMan1126: thanks, Andrea. I am not sure whether we change this to `Probability` since this indicates a percentage. If we decide to change it, we also need to change `focusedSearchActivationTime`.

arcuri82: good point! thx. i ll make a fix

arcuri82: just pushed a fix

PR #111, opened 2019-10-17
Basic assertions v4

bogdanmarculescu:
Supporting hal+json, empty responses, and responses with empty objects.

Let me know if there is anything I've missed.


core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -482,7 +485,7 @@ class TestCaseWriter {
 
         if (res.getBodyType() != null) {
             val type = res.getBodyType()!!
-            if (type.isCompatible(MediaType.APPLICATION_JSON_TYPE)) {
+            if (type.isCompatible(MediaType.APPLICATION_JSON_TYPE) || type.toString().contains("+json")) {
```
arcuri82: i think it should ignore the case, eg something like `type.toString().toLowerCase().contains("+json")` as I think it should handle `+JSON` as well

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -506,6 +509,12 @@ class TestCaseWriter {
                                 }
                             }
                         }
+                        else{
+                            // the object is empty
+                            if(format.isKotlin())  lines.add(".body(\"isEmpty()\", `is`(true))")
+                            else lines.add(".body(\"isEmpty()\", is(true))")
+                            //lines.add(".body(\"isEmpty()\", is(true))")
```
arcuri82: is this commented // out line still necessary?

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -728,29 +785,25 @@ class TestCaseWriter {
...
-    private fun applyEscapes(string: String): String {
-        val timeRegEx = "[0-2]?[0-9]:[0-5][0-9]".toRegex()
-        val ret = string.split("@")[0] //first split off any reference that might differ between runs
-                .split(timeRegEx)[0] //split off anything after specific timestamps that might differ
-                .replace("""\\""", """\\\\""")
-                .replace("\"", "\\\"")
-                .replace("\n", "\\n")
-                .replace("\r", "\\r")
-
+    fun flattenForAssert(k: MutableList<*>, v: Any): Map<MutableList<*>, Any>{
```
arcuri82: does this method need to be public?
if so, provide some docs for it, eg explaining its inputs and outputs

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: Done.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: Nope. Cleaned up the code a bit of other leftover comments.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
bogdanmarculescu: I don't think it will be used elsewhere, so I made it private for now. 
I also added a little bit of an explanation as to what it does. It's more of an auxiliary method (I looked for ready-made stuff that did similar things, but ultimately it was less time-consuming just to write the thing).

PR #113, opened 2019-10-18
able to configure whether include nlp libraries

ZhangMan1126:


arcuri82:
Hi,
 there is one problem, which leads all the 3 CI checks to fail.
`provided` dependencies are not transitive. So, the E2E modules dealing with classes in those 2 libraries will fail with a `NoClassDefFound` exception (eg, see `ResourceDependencyEMTest`). An easy solution (I guess) would be to explicitly add those 2 dependencies in the E2E modules in `test` scope.

PR #120, opened 2019-10-28
Npe fix

bogdanmarculescu:
Should handle null status code now. 
I also wrapped the sorting and test case writing in try-catch. In theory, if the writing of a test case fails for some reason, it should not impact the other test cases.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
@@ -93,11 +93,10 @@ class TestSuiteWriter {
 
                 }
                 catch (ex: Exception){
-                    Lines().apply {
-                        log.warn("A failure has occurred in writing test ${test.name}. \n "
-                                + "Exception: ${ex.localizedMessage} \n"
-                                + "At ${ex.stackTrace.joinToString(separator = " \n -> ")}. ")
-                    }
+                    log.warn("A failure has occurred in writing test ${test.name}. \n "
+                            + "Exception: ${ex.localizedMessage} \n"
+                            + "At ${ex.stackTrace.joinToString(separator = " \n -> ")}. ")
+                    Lines()
                 }
                 lines.add(testLines)
             }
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -68,13 +68,32 @@ class TestSuiteWriter {
 
             beforeAfterMethods(controllerName, lines)
 
-            val tests = testSuiteOrganizer.sortTests(solution, config.customNaming)
+            //catch any sorting problems (see NPE is SortingHelper on Trello)
+            val tests = try{
+                testSuiteOrganizer.sortTests(solution, config.customNaming)
+            }
+            catch (ex: Exception){
+                var counter = 0
```
arcuri82: add a WARN log

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -68,13 +68,32 @@ class TestSuiteWriter {
...
+                    TestCaseWriter()
+                            .convertToCompilableTestCode(config, test, baseUrlOfSut)
+
+                }
+                catch (ex: Exception){
+                    Lines().apply {
+                        this.add("// There was an exception ${ex.message}")
+                        this.add("// At ${ex.stackTrace}")
+                        this.add("// That prevented the test from being written")
+
```
arcuri82: this should be in a WARN log, not in the generated test file

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -68,13 +71,34 @@ class TestSuiteWriter {
...
-                        .convertToCompilableTestCode(config, test, baseUrlOfSut)
+                // catch writing problems on an individual test case basis
+                val testLines = try {
+                    TestCaseWriter()
+                            .convertToCompilableTestCode(config, test, baseUrlOfSut)
+
+                }
+                catch (ex: Exception){
+                    Lines().apply {
+                        log.warn("A failure has occurred in writing test ${test.name}. \n "
```
arcuri82: what is the reason of using `Lines().apply` here?

bogdanmarculescu: Mostly leftover from previous tries (where I was writing that stuff to the test file).
I am modifying it now to return an empty Lines() object (since that is what val testLines expects for each individual test case).

PR #123, opened 2019-11-01
Fix for some JSON assertion errors. Because of how the names of the

bogdanmarculescu:
The cache was not the problem. The problem was that, to allow access to some of the field (especially ones containing dashes or dots), the key has to be wrapped in single quotes. Should be fixed now.


bogdanmarculescu:
Added a test to the EscapeRest manual tests. I also amended the EscapeRest to return Json with dash and dot, to ensure that the generation works okay. Let me know if this is okay, or if they should be somehow separate.

PR #127, opened 2019-11-05
Expectations

bogdanmarculescu:
Part of the wrapping up.

Two types of expectations are generated:
1. expectations about the content (activated by expectationsMasterSwitch) which should duplicate basic assertions.
2. expectations about the structure (activated by responseStructureOracle) which make statements about the expected structure of returned objects (where the swagger has such objects to begin with).

Not sure why the CI assessments are not run at the moment.


arcuri82:
hi,
 all CI checks are failing. please look into them

core/src/main/kotlin/org/evomaster/core/output/service/PartialOracles.kt
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.gson.Gson
+import org.evomaster.core.output.Lines
+import org.evomaster.core.problem.rest.RestCallAction
+import org.evomaster.core.problem.rest.RestCallResult
+import org.evomaster.core.search.gene.ObjectGene
+import org.evomaster.core.search.gene.OptionalGene
+
+
+class PartialOracles {
```
arcuri82: provide some comments explaining what are the the partial oracles

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/StringMatcher.java
```diff
@@ -0,0 +1,37 @@
+package org.evomaster.client.java.controller.contentMatchers;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+public class StringMatcher extends TypeSafeMatcher<String> {
```
arcuri82: can  u add some test cases for this new class?

client-java/controller/src/main/java/org/evomaster/client/java/controller/contentMatchers/SubStringMatcher.java
```diff
@@ -0,0 +1,37 @@
+package org.evomaster.client.java.controller.contentMatchers;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+public class SubStringMatcher  extends TypeSafeMatcher<String> {
```
arcuri82: can  u add some test cases for this new class?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -703,7 +703,7 @@ class EMConfig {
 
     @Experimental
     @Cfg("Enable Expectation Generation. If enabled, expectations will be generated. " +
-            "A variable called activeExpectations is added to each test case, with a default value of false. If set to true, an expectation that fails will cause the test case containing it to fail.")
+            "A variable called expectationsMasterSwitch is added to the test suite, with a default value of false. If set to true, an expectation that fails will cause the test case containing it to fail.")
     var expectationsActive = false
```
arcuri82: are the e2e tests passing when this is set to `true`?

core/src/main/kotlin/org/evomaster/core/output/service/PartialOracles.kt
bogdanmarculescu: Will do. I'm expecting this to change quite a bit, but I will do my best to have the explanation there and to keep it as up to date as possible.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
bogdanmarculescu: Generated tests pass, yes.
EM's e2e tests fail (at least the TestCaseWriterTest does). I'll adjust that and see if anything else causes a problem.

bogdanmarculescu:
Hej. I think I've covered the required changes. The CI tests seem to pass. (Note that for the tests regarding TestSuiteWriter and TestCaseWriter, I forced the expectationsEnabled option to false).

Do you intend to have the expectations enabled by default?

arcuri82:
hi. thx. we can wait a bit before before putting them on by default (want to do some manual testing first, later this week)

PR #130, opened 2019-11-12
Covered targets issue

ZhangMan1126:


core/src/main/kotlin/org/evomaster/core/search/service/Archive.kt
```diff
@@ -385,7 +385,13 @@ class Archive<T> where T : Individual {
             }
 
             val limit = apc.getArchiveTargetLimit()
-            if (current.size < limit) {
+
+            /**
+             individual can be added only if the target k is not covered.
+             If a target is covered and a 'better'(e.g., shorter) individual appears,
+             it would be handled as replacement, see line 404
+             */
```
arcuri82: do not refer to a line by number, as that ll likely change at the next modification of this file :)

core/src/main/kotlin/org/evomaster/core/search/service/Archive.kt
```diff
@@ -385,7 +385,13 @@ class Archive<T> where T : Individual {
             }
 
             val limit = apc.getArchiveTargetLimit()
-            if (current.size < limit) {
+
+            /**
+             individual can be added only if the target k is not covered.
```
arcuri82: inside a function, no need to use JavaDoc notation `/**`, just use `/*`

core/src/main/kotlin/org/evomaster/core/search/service/Archive.kt
ZhangMan1126: modified :)

core/src/main/kotlin/org/evomaster/core/search/service/Archive.kt
ZhangMan1126: modified

PR #133, opened 2019-11-28
Documentation

arcuri82:


docs/publications.md
```diff
@@ -1,5 +1,12 @@
 # Publications
 
+The development of _EvoMaster_ is rooted in academia.
+Here, you can find the PDFs of all the academic publications based on _EvoMaster_. 
+Furthermore, slides of presentations can be  found [here](publications.md).
```
ZhangMan1126: presentations.md ?

arcuri82: you are right. thx! fixed now

jgaleotti:
Apologies for the delayed comment.

Regarding writer_driver.md I would suggest the following to improve the organisation.
- separate documents for embedded and external controllers (recommending embedded as the easier driver to write)
- an additional document describing how to deal with authentication (so, if no authentication is needed, the reader does not need to know that) 
- an additional document describing changes needed to add database (i.e. resetStateOfSUT, P6Spy, etc). 

Then, the organisation could be: 
- create driver for embedded springboot with no DB, no authentication
- changes for authentication (if needed)
- changes fot DB (if needed)

IMHO, these changes are a nice-to-have, but they are not mandatory.

arcuri82:
thx! I added those feedback on Trello, as not much time now to do major changes / restructuring now :(

PR #135, opened 2019-12-03
Split files

bogdanmarculescu:
Seems to work. 
Tried it on available case-studies, and have spotted no new problems introduced.


core/src/main/kotlin/org/evomaster/core/output/ExpectationsWriter.kt
```diff
@@ -103,6 +103,7 @@ class ExpectationsWriter {
                                 if (printableElement != "null"
                                         && printableElement != TestCaseWriter.NOT_COVERED_YET
                                         && !printableTh.contains("logged")
+                                        && !printableTh.contains("""\w+:\d{4,5}""".toRegex())
                                 ) {
```
arcuri82: can u add a comment to describe what this for?

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -650,6 +651,7 @@ class TestCaseWriter {
                         if (printableTh != "null"
                                 && printableTh != NOT_COVERED_YET
                                 && !printableTh.contains("logged")
+                                && !printableTh.contains("""\w+:\d{4,5}""".toRegex())
                         ) {
```
arcuri82: you are repeating the same regular expression several times. put it to a final constant variable

core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
```diff
@@ -19,6 +21,38 @@ object TestSuiteSplitter {
 
         return when(type){
             EMConfig.TestSuiteSplitType.NONE -> listOf(solution)
+            EMConfig.TestSuiteSplitType.CODE -> splitByCode(solution)
         }
     }
+
+    fun <T:Individual> splitByCode(solution: Solution<T>): List<Solution<T>>{
```
arcuri82: clarify it is HTTP codes

core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
```diff
@@ -19,6 +21,38 @@ object TestSuiteSplitter {
...
+
+
+        val s500 = solution.individuals.filter {
+            it.evaluatedActions().any { ac ->
+                (ac.result as RestCallResult).getStatusCode() == 500
+
```
arcuri82: clarify in a comment that here we just look at 500, and not at all the 5xx (eg, 501), as those are not so interesting (eg, not really related to find bugs)

core/src/main/kotlin/org/evomaster/core/output/ExpectationsWriter.kt
```diff
@@ -14,6 +14,7 @@ class ExpectationsWriter {
     private val responseStructureOracle = "responseStructureOracle"
     private lateinit var swagger: Swagger
     private lateinit var partialOracles: PartialOracles
+    private val portRegex = """\w+:\d{4,5}""".toRegex()
```
arcuri82: this would not catch `localhost:80` for example. i guess `{1,5}` would be better

and what about `127.0.0.1:8080`? that would not be considered.

i guess could have something like: `(\w+|(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})):\d{1,5}`

bogdanmarculescu: Agreed. Although I'm still hoping to think of a better and more long-term solution than this.

arcuri82:
Hi,
 can you add some E2E tests for the CODE option? (I thought I wrote it in a comment somewhere last time, but I do not see it anymore)

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -339,9 +339,16 @@ class Main {
...
             writer.setSwagger(swagger)
-            solutions.forEach {
+            /*solutions.forEach {
                 if(it.individuals.isNotEmpty()) writer.writeTests(it, controllerInfoDto?.fullName)
             }
+
+             */
+
+            solutions.filter { !it.individuals.isNullOrEmpty() }.map {
+                writer.writeTests(it, controllerInfoDto?.fullName)
```
arcuri82: that's weird. why a `map` instead of a `forEach`?
furthermore, for readability, the `.forEach{` would be best on next line instead of being 
on same line as `.filter{`

bogdanmarculescu: Just pushed a fix for it. It was mostly a leftover from a previous version.

arcuri82:
Hi,
 build fails. As `EMConfig` is modified, you need to do a rebase/merge, and recreate the `.md` documentation file

bogdanmarculescu:
Damn. Did the merge for the other branch, forgot about this one. Fixed. Sorry!

PR #145, opened 2020-01-15
Clustering

bogdanmarculescu:
Not quite sure what the build for Java 11 fails.


core/pom.xml
```diff
@@ -156,6 +156,17 @@
...
+        <dependency>
+            <groupId>org.nield</groupId>
+            <artifactId>kotlin-statistics</artifactId>
+            <version>1.2.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.edwardraff</groupId>
+            <artifactId>JSAT</artifactId>
+            <version>0.0.9</version>
+        </dependency>
```
arcuri82: dependency versions must not be in submodules, but rather in root pom file, under dependencyManagment.
Recall what explained in `for_developers.md`.
You can pass by my office and I can explain in person the details of Maven if that is not clear :)

core/src/main/kotlin/org/evomaster/core/output/Clusterer.kt
```diff
@@ -0,0 +1,60 @@
...
+                if (code!=null) (code == 500)
+                else false
+            }
+        }
+
+        /*
+        In order to "participate" in the clustering process, an action must be a RestCallResult
+        and it must be a failed result (i.e. a 500 call).
+         */
+        val cluterableActions = sol1.flatMap {
```
arcuri82: add a comment that in the future there can be other ways to identify if an action reveal a fault: eg if response is not matching the OpenApi schema.
Maybe could refactor in a function that, given as input an action, it tells us if it is revelaing a fault or not

core/src/main/kotlin/org/evomaster/core/output/Clusterer.kt
```diff
@@ -0,0 +1,60 @@
...
+            it.evaluatedActions()
+                    .filter{ ac -> ac.result is RestCallResult }
+                    .map { ac -> (ac.result as RestCallResult) }
```
arcuri82: this can be simplified with `.filterIsInstance<RestCallResult>()` (don't remember the exact syntax)

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
+package org.evomaster.core.output.clustering
+
+import org.evomaster.core.output.clustering.metrics.DistanceMetric
+
+/**
+ * Implementation of density-based clustering algorithm DBSCAN.
+ *
+ * Original Publication:
+ * Ester, Martin; Kriegel, Hans-Peter; Sander, J�rg; Xu, Xiaowei (1996).
+ * Simoudis, Evangelos; Han, Jiawei; Fayyad, Usama M., eds.
```
arcuri82: ah, the beauty of copy&paste a text from encoding ISO-8859-1 into a UTF-8 Kotlin file :)

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    constructor(values: Collection<V> , epsilon: Double = 0.5, minimumMembers: Int = 2, metric: DistanceMetric<V>){
+        setEpsilon(epsilon)
+        setMinimumNrOfMembers(members = minimumMembers)
+        setInputValues(values)
+        setMetric(metric)
+    }
+
+    fun setMinimumNrOfMembers(members: Int){
+        if(members > 1) this.minimumNumberOfClusterMembers = members
```
arcuri82: what happens if else branch? silently ignoring a user error is not a good idea.
Should throw an `IllegalArgumentExcpetion` on wrong inputs

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    fun setEpsilon(epsilon: Double){
+        if(epsilon > 0.0) this.epsilon = epsilon
+    }
```
arcuri82: see comment on `IllegalArgumentExcpetion`

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    fun setMetric(metric: DistanceMetric<V>){
+        this.metric = metric
+    }
+
+    fun setInputValues(values: Collection<V>){
+        if(values.isEmpty()){
+            throw Exception("DBSCAN: The list of values appears to be empty")
```
arcuri82: rather throw a `IllegalArgumentExcpetion`

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
...
+        for ( i in 0..neighbours2.size-1){
+            val temPt = neighbours2.get(i)
+            if(!neighbours1.contains(temPt)) neighbours1.add(temPt)
+        }
+        return neighbours1
+    }
+
+
+    fun performCLustering(): MutableList<MutableList<V>>{
+        if(!::inputValues.isInitialized) throw Exception("DBSCAN: List of inputs has not been initialized")
```
arcuri82: rather throw an `IllegalStateException`. Please check what I wrote in `for_developers.md`, in particular the section on `PRE AND POST CONDITIONS`.

core/src/main/kotlin/org/evomaster/core/output/clustering/DBSCANClusterer.kt
```diff
@@ -0,0 +1,137 @@
...
+        if(inputValues.isEmpty()) throw Exception("DBSCAN: List of inputs is empty")
+        if(inputValues.size < 2) throw Exception("DBSCAN: clustering less than 2 values is problematic")
+        if(epsilon < 0.0) throw Exception("DBSCAN: Maximum distance cannot be less than 0 (though I don't know how you got here, because it shouldn't be possible)")
+
```
arcuri82: for all checks above, see comment on `IllegalStateException`

core/src/main/kotlin/org/evomaster/core/output/clustering/metrics/DistanceMetricAction.kt
```diff
@@ -0,0 +1,80 @@
...
+        else {
+            val2.getBody()
+        }
+        return LevenshteinDistance.distance(message1.toString(), message2.toString())
+    }
+
+
+}
+
+object LevenshteinDistance {
```
arcuri82: `object` are classes in Kotlin. Kotlin allows to have different public classes in the same file, but I do not like it. I prefer Java that allows only 1 per file. Please move this in its own `LevenshteinDistance.kt` file, unless it is supposed to be a private object only used here

arcuri82:
Regarding Java 11, the build is done twice: once for 8 and once for 11. As we currently we have flaky tests, it can happen that one of the two fails while the other passes.

core/pom.xml
```diff
@@ -156,7 +156,12 @@
             <classifier>models</classifier>
             <scope>provided</scope>
         </dependency>
-
+        <!-- Clustering attempt-->
+        <dependency>
+            <groupId>org.nield</groupId>
+            <artifactId>kotlin-statistics</artifactId>
+            <version>1.2.1</version>
+        </dependency>
```
arcuri82: as said, version number should not be here. See `for_developers.md` for more details

core/src/main/kotlin/org/evomaster/core/output/Clusterer.kt
arcuri82: if the refactoring works, and the tests executing this new code pass, why is the old code left as a comment?

core/pom.xml
bogdanmarculescu: Minor oversight. Turns out I didn't need any of them, since I reimplemented the DBSCAN clusterer in kotlin. Removed the other attempt, but missed this one. Sorry.

PR #162, opened 2020-03-18
Clustering

bogdanmarculescu:
Changes regarding the clustering. In preparation for clustering experiment.


arcuri82:
hi. due to non-determinism, sometime 1 of the 3 CI runs fails. but 3 out of 3 failing means that most likely the changes have introduced a regression bug

PR #166, opened 2020-04-07
Use template when location fails parsing

agusaldasoro:
I found that:

## Get Resource Id
- When a resource ID was a string, then when calling the `getResourceId` method to generate the Location Header, then it was incorrectly parsed as it escaped the char `"`. For example if the payload is:
```
{ "id": "aStringId" }
```
Then, the resource id of the response was: `"\"aStringId\""`.

That resulted into generating a wrong location header, for example: `/v0/"aStringId"`. So it failed when trying to parse it as a URI in `resolveLocation`

## Resolve Location
- When a Location Header can not be parsed into a URI, then instead of returning the expected template the invalid location was returned. This caused making requests to an invalid URL and made Evomaster to stop running.

### DIFF
--- a/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
+++ b/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
@@ -37,7 +37,7 @@ public class EMTestUtils {
         try{
             locationURI = URI.create(locationHeader);
         } catch (Exception e){
-            return expectedTemplate;
+            return locationHeader;
         }
         String locationPath = locationURI.getPath();
         String[] locationTokens = locationPath.split("/");
--- a/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
+++ b/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
@@ -37,7 +37,7 @@ public class EMTestUtils {
         try{
             locationURI = URI.create(locationHeader);
         } catch (Exception e){
-            return locationHeader;
+            return expectedTemplate;
         }
         String locationPath = locationURI.getPath();
         String[] locationTokens = locationPath.split("/");
--- a/client-java/controller-api/src/test/java/org/evomaster/client/java/controller/api/EMTestUtilsTest.java
+++ b/client-java/controller-api/src/test/java/org/evomaster/client/java/controller/api/EMTestUtilsTest.java
@@ -54,7 +54,7 @@ public class EMTestUtilsTest {
 
         String resolvedLocation = EMTestUtils.resolveLocation(locationHeader, expectedTemplate);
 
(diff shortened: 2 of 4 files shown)


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
```diff
@@ -37,7 +37,7 @@ public static String resolveLocation(String locationHeader, String expectedTempl
         try{
             locationURI = URI.create(locationHeader);
         } catch (Exception e){
-            return locationHeader;
+            return expectedTemplate;
```
arcuri82, line 40: Hi! thanks for this PR! ;)
But I m not sure I do agree on this specific change.
If the Location header is wrong (ie it is not a valid URI), what would be the point to return the `expectedTemplate` here?
The idea was that, if a location header is wrong, to still have it in the generated tests, so a user can see that the SUT is doing something wrong. 
But, still, the EvoMaster process should not crash

client-java/controller-api/src/test/java/org/evomaster/client/java/controller/api/EMTestUtilsTest.java
```diff
@@ -45,6 +47,16 @@ public void testResolveLocation_null() {
...
     }
 
+    @Test
+    public void givenAnInvalidLocationHeaderWhenResolveLocationThenTheExpectedTemplateIsReturned() {
+        String expectedTemplate = "http://localhost:12345/a/x";
+        String locationHeader = "/a/\"52\"";
+
+        String resolvedLocation = EMTestUtils.resolveLocation(locationHeader, expectedTemplate);
+
+        assertEquals(expectedTemplate, resolvedLocation);
```
arcuri82, line 57: as template will have invalid {} characters, i think here we should expect to get `locationHeader`, and not `expectedTemplate`.

core/src/test/kotlin/org/evomaster/core/problem/rest/RestCallResultTest.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Test
+    fun givenANumericIdWhenGetResourceIdThenItIsReturnedAsString() {
+        val rc = RestCallResult(false)
+        rc.setBody("{\"id\":735}")
+        rc.setBodyType(MediaType.APPLICATION_JSON_TYPE)
+
+        val res = rc.getResourceId()
+
+        assertEquals("735", res)
+    }
```
arcuri82, line 28: nice tests!
indeed, whether the id is a number or a string, we should not quote it with " in the URI paths

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/EMTestUtils.java
agusaldasoro: Oh, you are right, I misunderstood this method. I'll undo this change, as with only changing the method `getResourceId()`, EvoMaster doesn't crash.

PR #168, opened 2020-04-24
Add interval for writing snapshots

agusaldasoro:
Added the functionality to enable writing snapshots of the generated tests in a custom interval.

### Summary
The snapshot tests are written in `src/em/EvoMasterTest_snaphost.java`. In the file a warning comment message is in the class comment:

```
/**
 * This file was automatically generated by EvoMaster on 2020-04-24T00:08:48.734-03:00[America/Argentina/Buenos_Aires]
 * <br>
 ************************************ WARNING ************************************ 
 * This is an snapshot of the generated tests after 88 seconds elapsed. *
 * The execution of Evomaster have not finished. *
 ********************************************************************************* 
```

While running evomaster, the following message is shown in the console:
```
* Going to save 60 tests to src/em
```

### Changes in Code
To centralize the writing of the test cases, a refactor has been made to separate the logic of writing the snapshots from one of creating the tests. This is located in `SearchAlgorithm.kt`


arcuri82:
hi,
 thx! as discussed by email:
- check how it works with `--testSuiteSplitType` on (eg `CODE`)
- look at why the build fails (there are some issues on CircleCI, but Travis must pass)

agusaldasoro:
Hi @arcuri82 , I've modified the method used to print the snapshots in order to behave in the same way no matter which the splitting type is: it always creates an unique file with the snapshots of the tests which gets re-written when a new snapshot is generated.
I've also fixed the failing tests. Thanks!

arcuri82:
thx! unfortunately I m very busy right now :(  might take me a few days before I ll have time to look at this PR :(

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -749,6 +749,13 @@ class EMConfig {
     var extraHeuristicsFile = "extra_heuristics.csv"
 
 
+    @Cfg("Enable to print snapshots of the generated tests during the search in an interval defined in snapshotsInterval.")
```
arcuri82, line 752: add `@Experimental`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -749,6 +749,13 @@ class EMConfig {
...
+    var enableWriteSnapshotTests = false
+
+    @Cfg("The size (in seconds) of the interval that the snapshots will be printed, if enabled.")
```
arcuri82, line 755: add `@Experimental`

core/src/main/kotlin/org/evomaster/core/search/algorithms/MosaAlgorithm.kt
```diff
@@ -275,11 +272,11 @@ class MosaAlgorithm<T> : SearchAlgorithm<T>() where T : Individual {
         // the population is not fully sorted
         var min = randomness.nextInt(population.size)
 
-        (0 until config.tournamentSize-1).forEach {
+        (0 until config.tournamentSize - 1).forEach { _ ->
```
arcuri82, line 275: why the `_ ->` ?

core/src/main/kotlin/org/evomaster/core/search/service/SearchAlgorithm.kt
```diff
@@ -30,16 +30,45 @@ abstract class SearchAlgorithm<T> where T : Individual {
...
 
+    private var lastSnapshot = 0
 
-    protected fun getMutatator() : Mutator<T> {
+    protected fun getMutator(): Mutator<T> {
         return mutator
     }
 
-    abstract fun search() : Solution<T>
+    abstract fun searchOnce()
```
arcuri82, line 42: can u add JavaDoc to explain what is this one? I guess it is doing a single step in the search process

core/src/main/kotlin/org/evomaster/core/search/service/SearchAlgorithm.kt
```diff
@@ -30,16 +30,45 @@ abstract class SearchAlgorithm<T> where T : Individual {
...
+
+    abstract fun setupBeforeSearch()
+
+    fun search(writeTests: (s: Solution<T>, t: String) -> Unit): Solution<T> {
```
arcuri82, line 46: can u add JavaDoc explaining what `writeTests` is?
in particular, it is unclear what `t` in the lambda is supposed to do

arcuri82: furthermore, in a lot of place you call `.search { _: Solution<*>, _: String ->  }`, which I assume is simply a lambda that tells that the snapshot printing does nothing

in that case, it would make sense to make `writeTests` nullable, and already initialized with default parameter `= null`. 
This would allow to simply call `.search()` when no snapshot is required

experiments/src/main/kotlin/org/evomaster/experiments/linear/Main.kt
```diff
@@ -157,7 +157,7 @@ class Main {
             val manager = a.first.getInstance(LifecycleManager::class.java)
 
             manager.start()
-            val sol = a.second.search()
+            val sol = a.second.searchOnce()
```
arcuri82, line 160: why the change from `search` to `searchOnce`? wouldn't that break this code, as only a single step in the search would be made?

experiments/src/main/kotlin/org/evomaster/experiments/objects/Main.kt
```diff
@@ -112,7 +112,7 @@ class Main {
             println("===============================================>")
             println("The search:")
 
-            val solution = imp.search()
+            val solution = imp.searchOnce()
```
arcuri82, line 115: same as above

experiments/src/main/kotlin/org/evomaster/experiments/pair/Main.kt
```diff
@@ -89,7 +89,7 @@ class Main {
             val manager = a.first.getInstance(LifecycleManager::class.java)
 
             manager.start()
-            val sol = a.second.search()
+            val sol = a.second.searchOnce()
```
arcuri82, line 92: same as above

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -749,6 +749,13 @@ class EMConfig {
...
+    var writeSnapshotTestsIntervalInSeconds = 3600
```
arcuri82, line 756: add comment on this line like `// ie, 1 hour`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: once we have used/tests this feature for a bit, then we can remove the `@Experimental` and put it `true` by default

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: also, would be good to have a E2E in which this value is set to something very short, eg 1 sec, to force the execution of this new code. not sure if there is an easy assertions to write here (eg check if files are generated / updated), but at least we should check that EM does not crash when this snapshot are generated

arcuri82:
Hi,
 thx for the PR! I wrote some comments.
Also, the build seems failing on Travis. please check it. 

again, thx!

arcuri82:
another question: snapshot files are generated each hour, but are they deleted once the search is completed? ie, when the final tests are generated, there is no need of the snapshot ones. or are the names exactly the same, and the final test suites simply override the snapshot ones?

arcuri82:
thx! Unfortunately I am bit busy these days... :( i ll have a look at it sometime next week

PR #176, opened 2020-05-06
Fix resource tests

ZhangMan1126:
move tests to `core-it` module


arcuri82:
Hi, 
 thx. 2  comments:

- do not copy&paste SQL and JSON files. You can import those from core (eg, see `test-jar` and `maven-jar-plugin`)
- add the new module to `report` (see pom file there)

PR #179, opened 2020-05-20
Clustering

bogdanmarculescu:
Fixes for the comments from the latest review.


core/src/main/kotlin/org/evomaster/core/output/PartialOracles.kt
```diff
@@ -32,15 +35,17 @@ class PartialOracles {
     private var oracles = mutableListOf(SupportedCodeOracle(), SchemaOracle())
     private val expectationsMasterSwitch = "ems"
 
-    fun variableDeclaration(lines: Lines, format: OutputFormat){
+    fun variableDeclaration(lines: Lines, format: OutputFormat, active: MutableMap<String, Boolean>){
```
arcuri82, line 38: provide JavaDoc to explain the the Key/Value in `active`

core/src/main/kotlin/org/evomaster/core/output/PartialOracles.kt
```diff
@@ -71,4 +76,45 @@ class PartialOracles {
         else return false;
     }
 
+    fun generatesExpectation(individual: EvaluatedIndividual<RestIndividual>): Boolean{
```
arcuri82, line 79: add some JavaDoc here.

core/src/main/kotlin/org/evomaster/core/output/PartialOracles.kt
```diff
@@ -71,4 +76,45 @@ class PartialOracles {
...
+            individual.evaluatedActions().any {
+                oracle.generatesExpectation(
+                        (it.action as RestCallAction),
+                        (it.result as RestCallResult)
+                )
+            }
+        }
+    }
+
+    fun failByOracle(individuals: MutableList<EvaluatedIndividual<RestIndividual>>): MutableMap<String, MutableList<EvaluatedIndividual<RestIndividual>>>{
```
arcuri82, line 90: unclear what this method is for. add some JavaDoc.
furthermore, do we need `Mutable` data structures here for `individuals`? are you modiyfing it by side-effects?

arcuri82: also explain what is the Key/Value in the returned map

core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
```diff
@@ -107,16 +107,27 @@ object TestSuiteSplitter {
 
     private fun execSummary(clusters : MutableMap<String, MutableList<MutableList<RestCallResult>>>,
                             solution: Solution<RestIndividual>,
+                            oracles: PartialOracles,
                             splitResult: SplitResult
             ) : Solution<RestIndividual> {
-        val execSol = mutableListOf<EvaluatedIndividual<RestIndividual>>()
+        val execSol = mutableSetOf<EvaluatedIndividual<RestIndividual>>()
```
arcuri82, line 113: a set requires immutability of its element. in general u cannot have a set of objects that can be mutated. u can use a temporary set to enforce uniqueness (which is fine) before converting to a list (as done here), but in that case you should add some comment

core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
```diff
@@ -13,6 +14,11 @@ abstract class ImplementedOracle {
     abstract fun variableDeclaration(lines: Lines, format: OutputFormat)
```
arcuri82, line 14: add JavaDocs for these methods

core/src/main/kotlin/org/evomaster/core/output/PartialOracles.kt
```diff
@@ -32,15 +35,29 @@ class PartialOracles {
...
+     * generate a failing expectation, and thus require the additional variables) or inactive (i.e. it will
+     * not generate a failing expectation and thus can be skipped).
+     *
+     * The goal of this method is to ensure that only relevant variables are generated (i.e. to avoid
+     * generating stub variables that are never used).
+     */
+    fun variableDeclaration(lines: Lines, format: OutputFormat, active: MutableMap<String, Boolean>){
         for (oracle in oracles){
-            oracle.variableDeclaration(lines, format)
+            if(active.get(oracle.getName()) == true) {
```
arcuri82, line 52: if on `active` you only call `get`, why did you choose to have it mutable? 
recall that `MutableMap` extends `Map`, so that should be the type as input (as more restrictive)

core/src/main/kotlin/org/evomaster/core/output/oracles/SupportedCodeOracle.kt
```diff
@@ -35,14 +39,29 @@ class SupportedCodeOracle : ImplementedOracle() {
...
+        //if(!supportedCode(call, res)){
+        if(generatesExpectation(call, res)){
             // The code is not among supported codes, so an expectation will be generated
             //val actualCode = res.getStatusCode() ?: 0
             //lines.add(".that($oracleName, Arrays.asList(${getSupportedCode(call)}).contains($actualCode))")
-            val supportedCode = getSupportedCode(call).joinToString(", ")
+            val supportedCodes = getSupportedCode(call)
+            //BMR: this will be a problem if supportedCode contains both codes and default...
+            if(supportedCodes.contains("0")){
+                lines.add("// WARNING: the code list seems to contain an unsupported code. The issue has been logged.")
```
arcuri82, line 50: the value (0 in this case) should be in the logs

PR #185, opened 2020-06-02
weight-based mutation rate

ZhangMan1126:


core/src/main/kotlin/org/evomaster/core/search/gene/ArrayGene.kt
```diff
@@ -103,87 +103,47 @@ class ArrayGene<T>(
...
+        if ( elements.isEmpty() || elements.size > maxSize){
+            return listOf()
+        }
+        val p = when(selectionStrategy){
+            SubsetGeneSelectionStrategy.ADAPTIVE_WEIGHT -> {
+                if(additionalGeneMutationInfo?.impact != null
+                        && additionalGeneMutationInfo.impact is ArrayGeneImpact
+                        && additionalGeneMutationInfo.impact.sizeImpact.noImprovement.any { it.value < 2 } //if there is recent improvement by manipulating size
+                ){
+                    0.3 // increase probability to mutate size
```
arcuri82, line 119: this is a bit hard to to read. on one side, you have a constant 0.3, on the other `MODIFY_SIZE`.
0.3 should be its own variable (with meaningful name), or a derived from `MODIFY_SIZE`, eg
`x * MODIFY_SIZE` for some x value

core/src/main/kotlin/org/evomaster/core/search/gene/DateGene.kt
```diff
@@ -59,38 +60,28 @@ class DateGene(
...
+        return !onlyValidDates || isValidDate()
+    }
 
-        val selects =  if (impact != null && impact is DateGeneImpact && archiveMutator.applyArchiveSelection()){
-            genes = listOf(
-                    Pair(year, impact.yearGeneImpact),
-                    Pair(month , impact.monthGeneImpact),
-                    Pair(day , impact.dayGeneImpact)
+    override fun adaptiveSelectSubset(internalGenes: List<Gene>, mwc: MutationWeightControl, additionalGeneMutationInfo: AdditionalGeneSelectionInfo): Map<Gene, AdditionalGeneSelectionInfo?> {
+        if (additionalGeneMutationInfo.impact != null && additionalGeneMutationInfo.impact is DateGeneImpact){
```
arcuri82, line 72: i have problems with `Map<Gene, AdditionalGeneSelectionInfo?>`.
a key in a map must be immutable, and `Gene` is not.
there are cases in which you could create such a map with short living inside a block (i think we had an example on this in the past), but not for a public method.
A simple alternative could be to use a list of pairs, eg something like `List<Pair<Gene,AdditionalGeneSelectionInfo?>>`

core/src/main/kotlin/org/evomaster/core/search/gene/DateGene.kt
```diff
@@ -59,38 +60,28 @@ class DateGene(
...
+            val maps = mapOf<Gene, GeneImpact>(
+                    year to additionalGeneMutationInfo.impact.yearGeneImpact,
```
arcuri82, line 74: yep, here `maps` could be as it is short-lived, and we can be sure that no modification is done on any gene

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -50,6 +50,11 @@ abstract class Gene(var name: String) {
      */
     abstract fun copy() : Gene
 
+    /**
+     * weight for mutation
```
arcuri82, line 54: add bit more info: eg, higher the weight, the higher the chances to be selected for mutation

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -79,45 +84,98 @@ abstract class Gene(var name: String) {
...
-                             selection: GeneMutationSelectionMethod,
-                             impact: GeneImpact?,
-                             geneReference: String,
-                             archiveMutator: ArchiveMutator,
-                             evi: EvaluatedIndividual<*>,
-                             targets: Set<Int>){
-        standardMutation(randomness, apc, allGenes)
+     * mutated gene should pass the check if needed, eg, DateGene
+     */
+    open fun mutationCheck() : Boolean = true
```
arcuri82, line 133: I am confused about this one. can you provide more info, eg what is special in DateGene, and what is going to happen when this is false

core/src/main/kotlin/org/evomaster/core/search/gene/MapGene.kt
```diff
@@ -91,21 +91,46 @@ class MapGene<T>(
...
+                        && additionalGeneMutationInfo.impact.sizeImpact.noImprovement.any { it.value < 2 } //if there is recent improvement by manipulating size
+                ){
+                    0.3 // increase probability to mutate size
+                }else MODIFY_SIZE
+            }
+            else ->{
+                MODIFY_SIZE
+
+            }
+        }
```
arcuri82, line 114: you have used exactly the same code before. would be good to somehow have it in its own function (utility class, or interface with defaults, or modify the gene hierarchy) to avoid copy&paste of same code

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
ZhangMan1126: Regarding DateGene, I found that there is a condition check as below. 
```
            do {
                val gene = randomness.choose(listOf(year, month, day))
                gene.standardMutation(randomness, apc, allGenes)
            } while (onlyValidDates && !isValidDate())
```
If this is false (i.e., the mutated internal gene does not satisfy some constraints of the gene), the internal gene will be mutated again.

arcuri82: ok, true. but let's add something like: "In some cases, we must have genes with 'valid' values. For example, a date with month 42 would be invalid. On the one hand, it can still be useful for robustness testing to provide such invalid values in a HTTP call. On the other hand, it would be pointless to try to add it directly into a database, as that SQL command would simply fail without any SUT code involved."

arcuri82:
thx! good work ;) just add that comment I wrote, then you can merge ;)

PR #187, opened 2020-06-12
Fix bugs

ZhangMan1126:
1. select immutable genes to candidates
2. `getValueAsPrintableString` of CycleObjectGene

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/gene/OptionalGene.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/gene/OptionalGene.kt
@@ -52,7 +52,7 @@ class OptionalGene(name: String,
     }
 
     override fun isMutable(): Boolean {
-        return selectable
+        return selectable && gene.isMutable()
     }
 
     override fun copyValueFrom(other: Gene) {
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/evomaster/core/search/gene/OptionalGene.kt
```diff
@@ -52,7 +52,7 @@ class OptionalGene(name: String,
     }
 
     override fun isMutable(): Boolean {
-        return selectable
+        return selectable && gene.isMutable()
     }
```
arcuri82, line 56: i do not think this one is correct.
if an OptionalGene X contains an immutable gene Y, still we can always mutate X, as we can flip its on/off status.
or is there any other issue i m missing here?

ZhangMan1126: then if an OptionalGene X contains an immutable gene Y, only `isActive` can be mutated?

arcuri82: yes, if Y is immutable, the only mutation allowed is to flip `isActive`, which depends on `selectable`

arcuri82: look at how `standardMutation` is implemented. but if that was not clear from the code, maybe it could make sense to add some explanation about it

PR #195, opened 2020-07-03
Epsilon handle

bogdanmarculescu:
Added a small change allowing the [recommendedEpsilon] value for each metric to be changed.

The clustering algorithm uses the recommendedEpsilon of each metric. This is based on the idea that, depending on the metric, the notion of neighbourhood may vary. Now these values can be selected from the options, rather than by editing the code.


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -653,6 +653,18 @@ class EMConfig {
     @Cfg("Generate an executive summary, containing an example of each category of potential fault found")
     var executiveSummary = false
 
+    @Experimental
+    @Cfg("The Distance Metric Last Line may use several values for epsilon." +
+            "During experimentation, it may be useful to adjust these values. Epsilon describes the size of the neighbourhood used for clustering, so may result in different clustering results." +
+            "Epsilon should be between 0.0 and 1.0. If the value is outside of that range, epsilon will use the default of 0.8.")
+    var lastLineEpsilon = 0.8
```
arcuri82, line 660: these constraints should be handled with the `@Min` and `@Max` annotations

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -653,6 +653,18 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("The Distance Metric Error Text may use several values for epsilon." +
+            "During experimentation, it may be useful to adjust these values. Epsilon describes the size of the neighbourhood used for clustering, so may result in different clustering results." +
+            "Epsilon should be between 0.0 and 1.0. If the value is outside of that range, epsilon will use the default of 0.8.")
+    var errorTextEpsilon = 0.8
```
arcuri82, line 666: same as above. might also want to check if `@PercentageAsProbability` makes sense here

core/src/main/kotlin/org/evomaster/core/output/clustering/metrics/DistanceMetricErrorText.kt
```diff
@@ -18,9 +18,12 @@ import javax.ws.rs.core.MediaType
  *
  */
 
-class DistanceMetricErrorText : DistanceMetric<RestCallResult>() {
+class DistanceMetricErrorText(
+        epsilon: Double = 0.6
+) : DistanceMetric<RestCallResult>() {
     private val name = "ErrorText"
-    private val recommendedEpsilon = 0.6
+    private val recommendedEpsilon = if (epsilon in 0.0..1.0) epsilon
```
arcuri82, line 25: if someone provides a wrong input, instead of silently change it into a 0.6, should rather throw an `IllegalArgumentException`.
anyway, constraints in EMConfig will already handle it

core/src/main/kotlin/org/evomaster/core/output/clustering/metrics/DistanceMetricLastLine.kt
```diff
@@ -13,9 +13,12 @@ import org.evomaster.core.problem.rest.RestCallResult
...
  */
 
-class DistanceMetricLastLine : DistanceMetric<RestCallResult>() {
+class DistanceMetricLastLine(
+        epsilon: Double = 0.8
+) : DistanceMetric<RestCallResult>() {
     private val name = "LastLine"
-    private var recommendedEpsilon = 0.8
+    private var recommendedEpsilon = if(epsilon in 0.0..1.0)  epsilon
+                                        else 0.8
```
arcuri82, line 21: same as previous... rather throw an exception

core/src/main/kotlin/org/evomaster/core/output/clustering/metrics/DistanceMetricLastLine.kt
```diff
@@ -29,4 +32,7 @@ class DistanceMetricLastLine : DistanceMetric<RestCallResult>() {
     override fun getName(): String {
         return name
     }
+    fun setRecommendedEpsilon(epsilon: Double){
+        if(epsilon in 0.0..1.0) recommendedEpsilon = epsilon
+    }
```
arcuri82, line 37: throw exception if invalid input

bogdanmarculescu:
I've made the changes. 

If I understand correctly, @PercentageAsProbability is used to activate or deactivate something? Currently, those values can be seen as percentages, but it might differ from distance to distance. Not quite sure what to say about it.

PR #202, opened 2020-10-07
ExperimentalAnnotationSolved

asmab89:
The Experimental Annotation Problem Solution

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -104,6 +104,8 @@ class EMConfig {
                 val text: String,
                 val constraints: String,
                 val enumValues: String,
+                val experimentalValues: String,
+                val validValues: String,
                 val experimental: Boolean
         ) {
             override fun toString(): String {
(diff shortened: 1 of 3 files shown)


docs/options.md
```diff
@@ -33,7 +33,7 @@ There are 3 types of options:
...
 |`outputFolder`| __String__. The path directory of where the generated test classes should be saved to. *Default value*: `src/em`.|
-|`outputFormat`| __Enum__. Specify in which format the tests should be outputted. If left on `DEFAULT`, then the value specified in the _EvoMaster Driver_ will be used. But a different value must be chosen if doing Black-Box testing. *Valid values*: `DEFAULT, JAVA_JUNIT_5, JAVA_JUNIT_4, KOTLIN_JUNIT_4, KOTLIN_JUNIT_5, JS_JEST`. *Default value*: `DEFAULT`.|
+|`outputFormat`| __Enum__. Specify in which format the tests should be outputted. If left on `DEFAULT`, then the value specified in the _EvoMaster Driver_ will be used. But a different value must be chosen if doing Black-Box testing. *Valid values*: ``. *Experimental values*: ``. *Default value*: `DEFAULT`.|
```
arcuri82, line 36: the entries for `*Valid values*:` have disappeared here.
furthermore, if there is no `. *Experimental values*: `, then it should not be printed out.

arcuri82 (changes requested):
hi. generate documentation file is not correct. see comment. furthermore, seems like you missed to commit the changes to the source code

docs/options.md
asmab89: Done

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -168,21 +170,24 @@ class EMConfig {
...
+            var validValues = ""
             val returnType = m.returnType.javaType as Class<*>
-
             if (returnType.isEnum) {
                 val elements = returnType.getDeclaredMethod("values")
                         .invoke(null) as Array<*>
-
                 enumValues = elements.joinToString(", ")
+               val experimentElements= elements.filter{ it is WithExperimentalOptions && it.isExperimental()}
+                val validElements= elements.filter{ it is WithExperimentalOptions && !it.isExperimental()}
```
arcuri82, line 181: the first check is fine, but this latter one is not, as it is not the negation of the first.
ie, if a parameter is not marked with  `WithExperimentalOptions` then all of its enum values are implicitely not experimental, ie:
 `it !is WithExperimentalOptions || !it.isExperimental()`

Furthermore, make sure to run the unit tests before committing, eg `ConfigToMarkdownTest`.
For example, here where is `experimentalValues ` in generating the output? ie the code that does output `Experimental values`

Also, if `experimentalValues.isBlank()`, then  `Experimental values` should not be outputted

arcuri82 (changes requested):
hi, getting there... :) but still some fixes are needed

arcuri82 (approved):

PR #219, opened 2020-11-26
fix a problem when handling cyclotron sut

ZhangMan1126:


bogdanmarculescu:
Seems okay to me.

ZhangMan1126:
@arcuri82 after discussing with Bogdan, we decide to provide a warning for the problem for the moment. In addition, the problem is also clarified as comments.

core/src/main/kotlin/org/evomaster/core/output/oracles/SchemaOracle.kt
```diff
@@ -178,7 +184,21 @@ class SchemaOracle : ImplementedOracle() {
...
+                            with cyclotron sut, a response of get /data/{key}/data is specified as
+                            "responses": {
+                              "200": {
+                                "description": "The data array for a Data Bucket",
+                                "schema": {
+                                  "type": "array"
+                                }
+                              },
+                         */
+                        log.warn("missing type of a response with Array schema {}", call.getName())
```
arcuri82, line 200: in the log, add info on endpoint/verb

arcuri82 (changes requested):

PR #220, opened 2020-11-26
[Postman seeding] Keep encoding of some safe-to-encode characters

AML14:
In an ideal world, URLs should be perfectly URL-encoded. We are not there yet, therefore, assume there could be some characters that need encoding


core/src/main/kotlin/org/evomaster/core/problem/rest/seeding/postman/PostmanParser.kt
```diff
@@ -139,6 +139,18 @@ class PostmanParser(
         }
     }
 
+    private fun getEncodedPath(path: String): String {
+        /*
```
arcuri82, line 143: Hi, this is indeed tricky, and we faced similar problem before.
See for example `RestPath.encode` where `URLEncoder.encode` is used, and there is also a bit of explanation and further links.
Would be good to use the same here

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/seeding/postman/PostmanParser.kt
AML14: The problem here is that we are receiving an input like this in the Postman collection:

`http://localhost:8080/rest/v2/subregion/Australia and New Zealand?fields=field1;field2`

This is the raw URL, and Postman (unfortunately) doesn't encode URLs properly when exporting a Postman collection. The `RestPath.encode` function is used to encode individual parameter keys and values, which is fine. But it cannot be used to encode a whole raw URL.

I've refactored the PostmanParser to obtain the path of the URL from somewhere else, which looks like this:
```json
"path": [
	"api",
	"v1",
	"queryHeaderPath",
	"%7Bprop1=val1,%20val2 and val3;prop2=val4%3Bprop3=val5}"
],
```

Now, the raw URL does not need to be encoded. Although, I'm using the UriUtils class from Spring-Web [1] to decode the path parameter values when generating the EvoMaster requests.

I've added a few more tests to check that everything is OK.

[1] Reference: https://www.baeldung.com/java-url-encoding-decoding

arcuri82:
thx. it's ok.
unfortuntately, it seems for some reasons CircleCI is not runnig on PR from forked reps.
are all you new tests passing? if so, then I can merge

AML14:
> thx. it's ok.
> unfortuntately, it seems for some reasons CircleCI is not runnig on PR from forked reps.
> are all you new tests passing? if so, then I can merge

@arcuri82 Yes, all my new tests are passing. Sorry for the delay in the response, GitHub doesn't notify me of these messages.

PR #221, opened 2020-12-01
Fix all failing tests on macOS Catalina with Spanish locale

AML14:
Additional changes:
- Updated testcontainers to 1.15.0 because some tests were failing due to this: https://github.com/testcontainers/testcontainers-java/issues/3166
- testcontainers 1.15.0 does not include org.jetbrains.annotations as a mvn dependency, which is used in class SqlHandlerInDBTest. Added such dependency.

### DIFF
--- a/client-java/controller/pom.xml
+++ b/client-java/controller/pom.xml
@@ -145,6 +145,11 @@
             <artifactId>testcontainers</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


arcuri82:
Hi. the other PR #222 is including these commits as well, isn't it? so, i can just close this PR, and look at the other?

AML14:
Exactly. PR #222 contains this one.

PR #223, opened 2020-12-02
Adaptivehw example as e2e

ZhangMan1126:
the example is used in the paper to illustrate MIO-WH*.

Performances of Base and MIO-WH* are explained in the comments.
Besides, with the tests, statistics and snapshots are recorded as well. do we need to further provide further analysis on them, e.g., a line plot?


e2e-tests/spring-rest-openapi-v2/src/test/java/com/foo/rest/examples/spring/adaptivehypermutation/AHypermutationRestController.java
```diff
@@ -0,0 +1,141 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Map;
+
+/** automatically created on 2020-10-22 */
+public class AHypermutationRestController extends EmbeddedSutController {
```
arcuri82, line 23: why is this not extending `SpringWithDbController` ?

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v2/src/test/java/com/foo/rest/examples/spring/adaptivehypermutation/AHypermutationRestController.java
ZhangMan1126: thanks
I directly used the controller from a separate project. This is fixed now.

PR #230, opened 2020-12-25
Client dotnet - EmController

amidgol:
This PR contains implementation of `EmController` and some other modifications. Now we are able to run the `EmbeddedEvoMasterController` and then it starts the `EmController`. Then we can call its APIs such as `runSut` and `infoSut`.


client-dotnet/src/Client.Util/SemaphoreLocker.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Client.Util {
+    public class SemaphoreLocker {
```
arcuri82, line 6: add some documentation explaining what this class is for

client-dotnet/src/Controller/Controllers/EmController.cs
```diff
@@ -0,0 +1,280 @@
...
+     
+     However, we want to check it only during testing
+     */
+    //TODO: check
+    private static readonly ICollection<string> connectedClientsSoFar = new SynchronizedCollection<string> ();
+
+    private static readonly SemaphoreLocker _locker = new SemaphoreLocker ();
+
+    //ATTENTION: The relative path is from the SUT
+    private static readonly string htmlWarning = System.IO.File.ReadAllText ("../../../../client-dotnet/src/Controller/Resources/warning.html");
```
arcuri82, line 31: I do not think this is correct. you are referring to a file on the hard-drive,  but that would not exist once the library is published as a DLL, isn't it? so, it should be read from the DLL, once packaged

client-dotnet/src/Controller/Controllers/EmController.cs
```diff
@@ -0,0 +1,280 @@
...
+    private void AssertTrackRequestSource (ConnectionInfo connectionInfo) {
+
+      var res = TrackRequestSource (connectionInfo);
+
+      if (!res) throw new InvalidOperationException ();
+    }
+
+    //Only used for debugging/testing
+    ///<summary>Returns host:port of all clients connected so far</summary>
+    public static HashSet<string> GetConnectedClientsSoFar () => connectedClientsSoFar.ToHashSet ();
```
arcuri82, line 61: instead of `HashSet<string>`, shouldn't it be a `ISet<string>`? ie, type declaration for containers should use the interface, and not the concrete classes

arcuri82 (changes requested):

client-dotnet/src/Client.Util/SemaphoreLocker.cs
amidgol: Some comments added

client-dotnet/src/Controller/Controllers/EmController.cs
amidgol: Yes you are right and it was my mistake. I modified this and now the html file gets copied into the SUT's bin directory after build. I think it is fixed now

client-dotnet/src/Controller/Controllers/EmController.cs
amidgol: Changed to ISet<string>

PR #241, opened 2021-01-28
Dotnet dbhandler

ZhangMan1126:
dbcleaner for dotnet. 
I added SupportedDbType in this cleaner, but we might replace it with DatabaseType which is under the controller-api.


client-dotnet/src/Controller/Controllers/db/DbCleaner.cs
```diff
@@ -0,0 +1,267 @@
...
+            if (connection.State == ConnectionState.Closed)
+            {
+                connection.Open();
+            }
+        }
+
+        // Man: restructure clearDatabase_Postgres here if Andrea agrees with.
+        // for more information about dbcommand and dbconnection
+        // https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/dbconnection-dbcommand-and-dbexception
+        public static void ClearDatabase(int retries, DbConnection connection, string schemaName, List<string> tablesToSkip, SupportedDatabaseType type)
```
arcuri82, line 41: Hi, i guess  we can discuss it on a video call

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/DbCleaner.java
```diff
@@ -157,4 +146,88 @@ In most cases (99.99%), this should not be a problem.
...
+    }
+
+
+    private static String getSchema(SupportedDatabaseType type){
+        switch (type){
+            case H2: return "PUBLIC";
+            case MYSQL: return "db";
+            case POSTGRES: return "public";
+        }
+        throw new IllegalStateException("NOT SUPPORT");
```
arcuri82, line 192: `"Database type " + type + " is not supported"`

arcuri82 (changes requested):

PR #243, opened 2021-01-31
Client dotnet test generation

amidgol:
It contains some modifications, mostly regarding to test generation with dotnet. It's still incomplete


ZhangMan1126 (commented):
thanks. I only have a minor comment for the current implementation

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
```diff
@@ -44,4 +45,23 @@ enum class OutputFormat {
...
+    fun isJavaOrKotlin() = isJava() || isKotlin()
+
+    fun isJUnit5() = false
+
+    fun isJUnit4() = false
+
+    fun isJUnit() = false
+
+    fun isDotnet() = true
+
```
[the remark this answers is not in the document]
amidgol: Sure I'll change this, thanks

client-dotnet/src/Controller.Api/dto/SutInfoDto.cs
```diff
@@ -47,4 +46,17 @@ public class SutInfoDto {
...
+    Note: this enum must be kept in sync with what declared in
+    org.evomaster.core.output.OutputFormat
+ */
+  public enum OutputFormat {
+    JAVA_JUNIT_5,
+    JAVA_JUNIT_4,
+    KOTLIN_JUNIT_4,
+    KOTLIN_JUNIT_5,
+    JS_JEST
+  }
```
arcuri82, line 61: we need to create a new one for dotnet, well, more specifically for CSharp

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
```diff
@@ -17,7 +17,8 @@ enum class OutputFormat {
     JAVA_JUNIT_4,
     KOTLIN_JUNIT_4,
     KOTLIN_JUNIT_5,
-    JS_JEST
+    JS_JEST,
+    DOTNET_CORE_3
```
arcuri82, line 21: hmmm, in .Net there can be different languages, such as F\# for example.
So, maybe we should specify `CSHARP_?`, where ? is the name of the library used for writing the test cases (eg like JUnit for JVM and Jest for JS)

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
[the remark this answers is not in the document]
arcuri82: yep, should not hard-code these things ;)
it can be fine in a quick-check, but should not endup in a PR.
note that the output format can be specified in the driver configuration

e2e-tests/dotnet-rest/test/RestApis.Tests.HelloWorld/Controller/EmbeddedEvoMasterController.cs
```diff
@@ -51,7 +50,7 @@ public class EmbeddedEvoMasterController : EmbeddedSutController {
                 RestApis.HelloWorld.Program.Main (new string[] { ephemeralPort.ToString () });
             });
 
-            await WaitUntilSutIsRunningAsync (ephemeralPort);
+            WaitUntilSutIsRunning (ephemeralPort);
```
arcuri82, line 53: we do not want to wait forever, in case of problems. so, should be an option to state of how long we wait at most (eg 20 seconds) before giving up

arcuri82 (changes requested):
thx! see my comments

e2e-tests/dotnet-rest/test/RestApis.Tests.HelloWorld/Controller/EmbeddedEvoMasterController.cs
amidgol: I added a timeout argument to this method with a default value of 20 seconds. If the SUT didn't start within the specified amount of time, a TimeoutException will be thrown

client-dotnet/src/Controller.Api/dto/SutInfoDto.cs
amidgol: Added

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
amidgol: Thanks for pointing out. I changed it to CSHARP_XUNIT

ZhangMan1126:
Hi, @arcuri82 
@amidgol identifies a problem regarding a sequence to build driver and sut. when compiling .net sut in EMB, there is no .net driver to be download in the network since we do not release it. In this case, we might need to mention to build the .net driver first for the nuget packages to be generated in the local nuget folder. Shall we update this info, i.e., install .net driver, in the build.md?

arcuri82:
Hi @ZhangMan1126 ,
 yes. In the same way that we have instructions for Java (i.e., "mvn install"), we need to provide instructions as well for .Net

client-dotnet/src/Client.Util/Client.Util.csproj
```diff
@@ -5,15 +5,19 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     
     <PackageId>Client.Util</PackageId>
-    <Version>1.0.0-alpha</Version>
+    <Version>1.1.1-SNAPSHOT</Version>
     <Authors>EvoMaster Team</Authors>
     <Company>Kristiania</Company>
   </PropertyGroup>
```
arcuri82, line 11: Hi,
  2 things here:
- in all these packages, should start with EvoMaster for the id, eg., `EvoMaster.Client.Util`. as i guess otherwise it will be a problem to upload those artifacts to NuGet in the future
- remove  `<Company>Kristiania</Company>`

arcuri82 (changes requested):

arcuri82:
Hi,
 also it looks like there are `Conflicting files`. please solve those ;)

client-dotnet/src/Client.Util/Client.Util.csproj
amidgol: Hi, they are fixed now

amidgol:
> Hi,
> also it looks like there are `Conflicting files`. please solve those ;)

PR #248, opened 2021-02-12
not able to import evomaster.org file 

webrollygoal:
not able to import evomaster.org file  into EMDriver ps6 and evomaster.org not getting imported with a new swagger project


arcuri82:
Hi,
  thanks for your interest in EM. But what is this? why are you trying to merge master into a branch?

PR #262, opened 2021-03-17
Fix ahm tests

arcuri82:


core/pom.xml
```diff
@@ -28,7 +28,8 @@
             -->
             <groupId>org.evomaster</groupId>
             <artifactId>evomaster-client-java-controller</artifactId>
-            <scope>test</scope>
+            <!-- Need this module for runWithDeterministicLogger in LoggingUtil -->
+<!--            <scope>test</scope>-->
```
arcuri82, line 32: this should be reverted. as `runWithDeterministicLogger` is only used in the tests, move that function in a new class in test scope. so no need to modify pom file

core/src/main/kotlin/org/evomaster/core/logging/LoggingUtil.kt
```diff
@@ -86,11 +89,22 @@ class LoggingUtil {
          */
         fun runWithDeterministicLogger(lambda: () -> Unit) : String{
```
arcuri82, line 90: move this function somewhere under src/test/kotlin

arcuri82:
`DbDirectIntWithSqlEMTest` seems failing

PR #268, opened 2021-03-24
Db support

ZhangMan1126:
dbcleaner for MySQL, MariaDB and MS SQL Server with java and .net.
For java, test containers are created with genericcontainer.
For .net, test containers are created with pre-configured containers according to db for the moment.

Note that `CIUtils.java` is moved out from `e2e-test-utils` in order to fix a cyclic dependency problem


client-dotnet/src/Controller.Tests/Controllers/db/DbCleanerMySqlTest.cs
```diff
@@ -48,7 +49,7 @@ public async Task InitializeAsync()
 
         public async Task DisposeAsync()
         {
-            DbCleaner.ClearDatabase(_connection, null, DatabaseType.MYSQL);
+            DbCleaner.ClearDatabase(_connection, null, DatabaseType.MYSQL, "db");
             
             // TODO find a proper solution to clean all data in mysql db, instead of dropping db and closing connection
             // SqlScriptRunner.ExecCommand(_connection, "DROP DATABASE db;");
```
arcuri82, line 55: i guess this TODO can be removed?

client-java/controller/pom.xml
```diff
@@ -154,6 +154,39 @@
             <artifactId>mockito-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <!--Need to confirm with Andrea-->
+            <!-- check if this version configuration is needed to add to the root pom.xml-->
+            <version>8.0.23</version>
+            <scope>test</scope>
```
arcuri82, line 163: yep, those should be in root pom file

client-java/controller/pom.xml
```diff
@@ -154,6 +154,39 @@
...
+        </dependency>
+
+
+        <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+            <version>2.7.2</version>
+            <scope>test</scope>
```
arcuri82, line 171: same for this

client-java/controller/pom.xml
```diff
@@ -154,6 +154,39 @@
...
+        </dependency>
+
+        <!--https://docs.microsoft.com/en-us/sql/connect/jdbc/download-microsoft-jdbc-driver-for-sql-server?view=sql-server-ver15-->
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>9.2.1.jre8</version>
+            <scope>test</scope>
```
arcuri82, line 179: same for this

client-java/controller/src/test/java/org/evomaster/client/java/controller/db/DbCleanerSQLServerTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        String host = mssqlserver.getContainerIpAddress();
+        int port = mssqlserver.getMappedPort(PORT);
+        String url = "jdbc:sqlserver://"+host+":"+port;
+
+        /*
+            TODO Man:there exist a problem on creating the connection with CI and GA,
+             but the problem cannot be reproduced on MacOS.
+         */
+        connection = DriverManager.getConnection(url, "SA", PASSWORD);
```
arcuri82, line 56: this is weird. looks like an issue with default user on the DB, but that should be the same due to Docker. i ll try to reproduce on my Windows machine

arcuri82 (changes requested):

client-java/controller/pom.xml
arcuri82: why are those `<version>` still here? 
they should had been declared in `<dependencyManagment>` in root pom file, so child modules do not need to define any `<version>`.

arcuri82 (changes requested):

arcuri82:
@ZhangMan1126  see comments on pom file

client-java/controller/pom.xml
ZhangMan1126: thanks. these are updated with the latest commits.

PR #273, opened 2021-03-30
export headerdto

ZhangMan1126:
configure auth


arcuri82:
Hi,
 why is this?  `AuthenticationDto`  already contains a list of `HeaderDto` for auth

PR #274, opened 2021-04-01
Gql v06

asmab89:


core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -1150,4 +1192,43 @@ class TestCaseWriter {
                 && !printableContent.contains("logged")
                 && !printableContent.contains("""\w+:\d{4,5}""".toRegex()))
     }
+
+    fun getPrintableInputGenes(printableInputGene: MutableList<String>): String {
+        var printableInputGenes = ""
+        for (elt in printableInputGene) {
+            printableInputGenes = "$elt,$printableInputGenes"
+        }
```
arcuri82, line 1200: this can be simplified with something like `printableInputGene.joinToString(",")`  (not sure of the syntax though)

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -1150,4 +1192,43 @@ class TestCaseWriter {
...
+        printableInputGenes = printableInputGenes.substring(0, printableInputGenes.length - 1)
+        printableInputGenes = printableInputGenes.replace("\"", "\\\"")
+        return printableInputGenes
+    }
+
+    fun getPrintableInputGene(inputGenes: List<Gene>): MutableList<String> {
+        val printableInputGene: MutableList<String> = mutableListOf()
```
arcuri82, line 1207: can simplify with: `val printableInputGene = mutableListOf<String>()`

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -1150,4 +1192,43 @@ class TestCaseWriter {
...
+        var mutation = "{${returnGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.BOOLEAN_SELECTION_MODE)}}".replace("{${a.methodName}", "")
+        mutation = mutation.substring(0, mutation.length - 1)
+        return mutation
+    }
+
+    fun getQuery(returnGene: Gene, a: GraphQLAction): String {
+        var query = "{${returnGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.BOOLEAN_SELECTION_MODE)}}"
+        query = query.replace("{${a.methodName}", "", true)
+        query = query.substring(0, query.length - 1)
+        return query
```
arcuri82, line 1230: no need to re-assign query:
```
var query = "{${returnGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.BOOLEAN_SELECTION_MODE)}}"
        .replace("{${a.methodName}", "", true)
        .substring(0, query.length - 1)
```

arcuri82: i m bit concerned about this `replace` calls... what if the same value is repeated inside a string? eg, when `a.methodName` is a generic term. furthermore, should add some comments to specify why this is done

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -257,24 +258,114 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
...
 
-        val invocation = builder.buildPost(bodyEntity)
+            } else {
+                var mutation = getMutation(returnGene, a)
+
+                val bodyEntity = Entity.json("""
+            { "query" : "mutation{    ${a.methodName}  ($printableInputGenes)    $mutation    }","variables":null}
+        """.trimIndent())
+                val invocation = builder.buildPost(bodyEntity)
+                return invocation
```
arcuri82, line 327: this
````
val invocation = builder.buildPost(bodyEntity)
return invocation
```
is repeated a lot of times. could be simplified (eg taken out after the if statements)

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -257,24 +258,114 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
...
+            }
+        }
+        val invocation = builder.buildPost(Entity.json("""
+          """.trimIndent()))
```
arcuri82, line 331: what is this case with empty string defined as multi-line string?

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -257,24 +258,114 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
...
         return invocation
     }
 
+    fun getMutation(returnGene: Gene, a: GraphQLAction): String {
+        var mutation = "{${returnGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.BOOLEAN_SELECTION_MODE)}}".replace("{${a.methodName}", "")
+        mutation = mutation.substring(0, mutation.length - 1)
+        return mutation
```
arcuri82, line 338: simplify with `.substring(0, mutation.length - 1)` instead of re-assigne `mutation`.
and, why the substring? what is the reason? (add comment)

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -257,24 +258,114 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
...
+    }
+
+    fun getQuery(returnGene: Gene, a: GraphQLAction): String {
+        var query = "{${returnGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.BOOLEAN_SELECTION_MODE)}}"
+        query = query.replace("{${a.methodName}", "", true)
+        query = query.substring(0, query.length - 1)
+        return query
```
arcuri82, line 345: see previous comments

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -257,24 +258,114 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
...
+    }
+
+    fun getPrintableInputGenes(printableInputGene: MutableList<String>): String {
+        var printableInputGenes = ""
+        for (elt in printableInputGene) {
+            printableInputGenes = "$elt,$printableInputGenes"
+        }
```
arcuri82, line 352: see previous comments

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -142,7 +142,22 @@ open class ObjectGene(name: String, val fields: List<out Gene>, val refType: Str
                 }.joinToString(","))
                 buffer.append("}")
             }
-        } else {
+        }
+        else if (mode == GeneUtils.EscapeMode.GQL_INPUT_MODE) {
+
```
arcuri82, line 147: add comment to specify what this is for

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
asmab89: Done.

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
asmab89: Done

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
asmab89: done

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
asmab89: Removed

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
asmab89: done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
asmab89: Added

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
asmab89: I agree. I did not find another solution. I tried to remove the name (return param name) from the action builder it throws the exception: empty name.
Comments->done

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
asmab89: .substring-> did not know how to simplify it 
add comment -> done

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
asmab89: .replace -> done.
.substring-> did not know how to simplify it

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
arcuri82: why not handling it inside the `getValueAsPrintableString`? could either be based on `mode`, or can add new param to that function

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
asmab89: 1.why not handling it inside the getValueAsPrintableString? -> if I put it inside: getValueAsPrintableString I will lost all the objects names inside the return object gene since it is printed recursively.
2.could either be based on mode, or can add new param to that function->I did not understand those two propositions

PR #276, opened 2021-04-06
Js handle ternary

ZhangMan1126:
Js handle ternary operator

### DIFF
--- a/client-js/evomaster-client-js/src/instrumentation/babel-plugin-evomaster.ts
+++ b/client-js/evomaster-client-js/src/instrumentation/babel-plugin-evomaster.ts
@@ -133,6 +133,40 @@ export default function evomasterPlugin(
             Look like is checking for some types, which in theory should always be pure, like literals.
             But very unclear on its features... eg, would it handle as pure "!false" ???
             TODO need to check... furthermore we do not care if throwing exception
+
+            TODO: Man
+            also check the source code of isPure() from ../@babel/traverse/lib/scope/index.js
+
+            isPureish() seems not fully applicable, we might need to modify it.
+
+            found the source code of isPureish() from ../@babel/types/lib/validators/generated/index.js
+
+                function isPureish(node, opts) {
+                  if (!node) return false;
+                  const nodeType = node.type;
+
+                  if (nodeType === "Pureish"
+                    || "FunctionDeclaration" === nodeType
+                    || "FunctionExpression" === nodeType
+                    || "StringLiteral" === nodeType
+                    || "NumericLiteral" === nodeType
+                    || "NullLiteral" === nodeType
+                    || "BooleanLiteral" === nodeType
+                    || "ArrowFunctionExpression" === nodeType
+                    || "ClassExpression" === nodeType
+                    || "ClassDeclaration" === nodeType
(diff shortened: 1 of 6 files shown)


client-js/evomaster-client-js/src/instrumentation/babel-plugin-evomaster.ts
```diff
@@ -190,6 +225,67 @@ export default function evomasterPlugin(
...
+            Note that we do not further replace 'test' here.
+            if it is related to condition, it will be replaced by other existing replacement and
+            additional branch will be added there.
+
+            From trello
+            where ternary needs to create 2 objectives:
+                - 1 for for when it is executed/called  (h=1)
+                - 1 for when no exception (h=0.5 and then h=1 if and onyl if () => B did not throw exception)
+            TODO: check with Andrea, I am not clear about the second point,
+                do we need to handle 'consequent' and 'alternate' differently?
```
arcuri82, line 263: can add `Andrea: no, they should be treated the same`

client-js/evomaster-client-js/src/instrumentation/babel-plugin-evomaster.ts
```diff
@@ -190,6 +225,67 @@ export default function evomasterPlugin(
...
+                in addition, do we need to handle 'throw exception' with different h?
+                how about itself is a throw exception expression, e.g., A? B: (throw e)
+                if needed, we might need to add a variable for a check with 't.isThrowStatement()'
+         */
```
arcuri82, line 267: very interesting point! didn't think about it.
if it is a `throw` statement, can have `h=1`. can u add a test for it?

client-js/evomaster-client-js/src/instrumentation/heuristic/HeuristicsForBooleans.ts
```diff
@@ -332,4 +333,31 @@ export default class HeuristicsForBooleans {
         return res;
     }
 
+    public static handleTernary(f: () => any, fileName: string, line: number, index: number) {
+
+        // TODO: Man what is this for? do I need this?
+        HeuristicsForBooleans.lastEvaluation = null;
```
arcuri82, line 339: it was explained in the paper. but i guess we should put it in a (private) function (e.g., `resetLastEvaluation`), with some comments like: `Make sure that nested evaluations of && and || do not use unrelated previous computation`.

I am not sure if it is really needed here, but can leave it for now

client-js/evomaster-client-js/src/instrumentation/heuristic/HeuristicsForBooleans.ts
```diff
@@ -332,4 +333,31 @@ export default class HeuristicsForBooleans {
...
+            How about the f() itself is an throw exception expression
+         */
+        try{
+            res = f();
+            ExecutionTracer.updateObjective(id, 1);
+        }catch (e) {
+            ExecutionTracer.updateObjective(id, 0.5);
+            //Man: might throw exception again
+            throw e;
+            //res = e;
```
arcuri82, line 355: yep, must guarantee the same semantic

arcuri82 (changes requested):

PR #290, opened 2021-04-20
graphql db support

ZhangMan1126:
graphql db support
e2e tests


core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLFitness.kt
```diff
@@ -44,7 +44,7 @@ class GraphQLFitness : HttpWsFitness<GraphQLIndividual>() {
         val cookies = getCookies(individual)
 
         //TODO
-        //doInitializingActions(individual)
+        doInitializingActions(individual)
```
arcuri82, line 47: remove TODO

arcuri82 (changes requested):

PR #292, opened 2021-04-21
Split default

bogdanmarculescu:
Making the split the default option. 

For tests that don't generate all 3 types of files (faults, successes, and other) it may require turning the option off (the WsTestBase might run into trouble compiling files that are not there).


core/src/main/kotlin/org/evomaster/core/output/oracles/SchemaOracle.kt
```diff
@@ -103,8 +122,8 @@ class SchemaOracle : ImplementedOracle() {
 
         //this differs between kotlin and java
         when{
-            format.isJava() ->lines.add(".that($variableName, $json_ref.getMap(\"\").keySet().containsAll(Arrays.asList($referenceKeys)))")
-            format.isKotlin() -> lines.add(".that($variableName, $json_ref.getMap<Any, Any>(\"\").keys.containsAll(Arrays.asList($referenceKeys)))")
+            format.isJava() ->lines.add(".that($variableName, ($json_ref as LinkedHashMap<*,*>).keys.containsAll(Arrays.asList($referenceKeys)))")
+            format.isKotlin() -> lines.add(".that($variableName, ($json_ref as LinkedHashMap<*,*>).keys.containsAll(Arrays.asList($referenceKeys)))")
         }
```
arcuri82, line 127: this seems brittle... ie, referifng to concrete class `LinkedHashMap` instead of the interface `Map`

core/src/main/kotlin/org/evomaster/core/output/oracles/SchemaOracle.kt
```diff
@@ -115,13 +134,13 @@ class SchemaOracle : ImplementedOracle() {
...
-                    lines.add(".containsAll($json_ref.getMap(\"\").keySet()))")
+                    lines.add(".containsAll(($json_ref as LinkedHashMap<*,*>).keys))")
                 }
             }
             format.isKotlin() -> {
                 lines.add(".that($variableName, listOf<Any>($referenceOptionalKeys)")
                 lines.indented {
-                    lines.add(".containsAll($json_ref.getMap<Any, Any>(\"\").keys))")
+                    lines.add(".containsAll(($json_ref as LinkedHashMap<*,*>).keys))")
                 }
```
arcuri82, line 144: same as previous comment

core/src/main/kotlin/org/evomaster/core/output/oracles/SchemaOracle.kt
```diff
@@ -158,6 +177,50 @@ class SchemaOracle : ImplementedOracle() {
         }
     }
 
+    fun supportedObject(obj: LinkedTreeMap<*,*>, call: RestCallAction): Boolean{
+
```
arcuri82, line 181: use `Map` instead.
furthermore, add JavaDocs to this function

core/src/main/kotlin/org/evomaster/core/output/oracles/SchemaOracle.kt
```diff
@@ -229,9 +292,35 @@ class SchemaOracle : ImplementedOracle() {
...
+        //val referenceObject = objectGenerator.getNamedReference(expectedObject)
+
+        var supported = true
+
+        if (res.getBodyType()?.isCompatible(MediaType.APPLICATION_JSON_TYPE) == true){
+            val actualObject = Gson().fromJson(res.getBody(), Object::class.java)
+            if  (actualObject is LinkedTreeMap<*,*>)
+                supported = supportedObject(actualObject, call)
+            else if (actualObject is ArrayList<*>
+                    && (actualObject as ArrayList<*>).isNotEmpty()
```
arcuri82, line 312: use interfcases, eg `List`, instead of concrete classes (which is brittle)

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/adaptivehypermutation/AHypermutationAWHTest.java
```diff
@@ -111,6 +111,8 @@ public void testRunMIOAWH() throws Throwable {
                 budget,
                 true,
                 (args) -> {
+                    args.add("--testSuiteSplitType");
+                    args.add("NONE");
```
arcuri82, line 116: adding this on every single E2E is not viable, nor scalable.
should rather be default in `WsTestBase`, with option to use other modes if needed

arcuri82 (changes requested):

PR #294, opened 2021-04-23
handling assertion with response

ZhangMan1126:
handle array and object in array
handle null


core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -688,6 +675,33 @@ class TestCaseWriter {
         }
     }
 
+    private fun handleAdditionalFieldValues(stringKey: String, resContentsItem: Any?): List<Pair<String, String>>?{
```
arcuri82, line 678: please add some JavaDocs, especially about what are the pairs of strings returned

arcuri82 (changes requested):

PR #302, opened 2021-04-29
generate tests even with problems

ZhangMan1126:
TODO handling string with backslash eg., rxHhwHeR\ in gql

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import io.swagger.v3.oas.models.OpenAPI
 import org.evomaster.core.EMConfig
 import org.evomaster.core.logging.LoggingUtil
+import org.evomaster.core.output.formatter.MismatchedFormatException
 import org.evomaster.core.output.formatter.OutputFormatter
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
```diff
@@ -1162,7 +1163,12 @@ class TestCaseWriter {
 
         val bodyEntity = GraphQLUtils.generateGQLBodyEntity(call, configuration.outputFormat)
 
-        val body = if (bodyEntity!=null) OutputFormatter.JSON_FORMATTER.getFormatted(bodyEntity.entity) else {
+        val body = if (bodyEntity!=null) {
+            if (OutputFormatter.JSON_FORMATTER.isValid(bodyEntity.entity))
+                OutputFormatter.JSON_FORMATTER.getFormatted(bodyEntity.entity)
+            else
+                bodyEntity.entity
```
arcuri82, line 1170: issue a warning log here

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/TestCaseWriter.kt
ZhangMan1126: added. thanks

PR #303, opened 2021-04-30
Working on sorting out the executive summary.

bogdanmarculescu:
executive summary is only available for the clustering option.

I've also removed the summary only option. The clustering is needed to get the summary, so throwing away the results of the clustering to get only the summary seems a bit weird.


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -668,15 +673,15 @@ class EMConfig {
     enum class TestSuiteSplitType {
         NONE,
         CLUSTER,
-        SUMMARY_ONLY,
         CODE
     }
```
arcuri82, line 677: what is the difference between CLUSTER and CODE? add comments here to clarify

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -668,15 +673,15 @@ class EMConfig {
...
 
     @Cfg("Instead of generating a single test file, it could be split in several files, according to different strategies")
     var testSuiteSplitType = TestSuiteSplitType.CODE
 
-    @Experimental
-    @Cfg("Generate an executive summary, containing an example of each category of potential fault found")
+    @Cfg("Generate an executive summary, containing an example of each category of potential fault found." +
+                    "NOTE: This option is only meaningful when used in conjuction with clustering. " +
+                    "This is achieved by turning the option --testSuiteSplitType to CLUSTER")
     var executiveSummary = false
```
arcuri82, line 686: why is this not `true` be default? 
and what would be the effects of having `CLUSTER` without the `executiveSummary`?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
bogdanmarculescu: I'm hesitant to have it true by default, because it seems easy to change the split type, but forget to change the summary. If the split is CLUSTER, but the executive summary is off, then the normal test suites (faults, successes, and other) are generated. Having executive summary on only adds the "fault_representatives".

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
bogdanmarculescu: Will do.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: would that mean that `CLUSTER` without `executiveSummary` is the same as `CODE`?

bogdanmarculescu: Selecting CODE only, just separates the tests by the code, and conducts no clustering (and attaches no cluster labels). CLUSTER conducts clustering and attaches the labels, in addition to the split. At the moment, they both generate the same files.

The main difference is that CLUSTER includes partial oracles in the definition of "failed".  If a partial oracle allows for something to be labeled as "failed" even if it has a 200 code (for example, a GET after DELETE), then that test case would be under the "faults" file according to CLUSTERING, and under the "successes"  according to CODE. 

I'll try to make my reasoning clear in the documentation.

arcuri82: ok. add these explanations in the comments.
then, `CLUSTER` and `executiveSummary` should be on by default. When `executiveSummary` is on but not `CLUSTER`, then issue a log warning instead of crashing EM

arcuri82:
hi.
  CI are both failing. please look at them.
also merge latest from master, as i merged my main refactoring of TestCaseWriter

bogdanmarculescu:
Found a small bug that caused a test to fail. Re-running tests locally now to make sure it's all good. I'll merge the latest master changes too.

PR #307, opened 2021-05-07
Resource ah db

ZhangMan1126:
sql on resource handling


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1071,6 +1071,15 @@ class EMConfig {
     @Min(0.0)
     var maxSqlInitActionsPerResource = 0
 
+    @Experimental
+    @Cfg("When generating resource using SQL (e.g., sampler or mutator), how many new rows (max) to generate for the specific resource each time")
+    var employSqlNumResourceStrategy = SqlInitResourceStrategy.NONE
```
arcuri82, line 1076: is this Cfg description correct for this variable?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1071,6 +1071,15 @@ class EMConfig {
...
+
+    enum class SqlInitResourceStrategy{
+        NONE,
+        RANDOM,
+        DPC
```
arcuri82, line 1081: add comment on what is DPC

core/src/main/kotlin/org/evomaster/core/problem/util/ParamUtil.kt
```diff
@@ -30,6 +38,69 @@ class ParamUtil {
...
+
+        /**
+         * bind values between [call] and [dbActions]
+         * [candidates] presents how to map [call] and [dbActions] at Gene-level
+         * [forceBindParamBasedOnDB] is an option to bind params of [call] based on [dbActions]
+         */
+        fun bindCallWithDBAction(
+            call: RestResourceCalls,
+            dbActions: MutableList<DbAction>,
+            candidates: MutableMap<RestAction, MutableList<ParamGeneBindMap>>,
```
arcuri82, line 50: this needs to be refactored. RestAction is a mutable object, isn't it? if so, should not be used as a key in a map, especially as input to a public function

core/src/main/kotlin/org/evomaster/core/problem/util/ParamUtil.kt
```diff
@@ -30,6 +38,69 @@ class ParamUtil {
...
+            resourceCluster : Map<String, RestResourceNode>,
```
arcuri82, line 51: what is the String key here representing?

core/src/main/kotlin/org/evomaster/core/search/EvaluatedIndividual.kt
```diff
@@ -252,17 +254,21 @@ class EvaluatedIndividual<T>(val fitness: FitnessValue,
...
-        Lazy.assert{mutatedGenes.mutatedIndividual != null}
-
         Lazy.assert {
-            mutatedGenes.mutatedIndividual != null
-            tracking != null
+            mutatedGenes.mutatedIndividual != null && tracking != null
         }
 
+        if(previous.getSizeOfImpact(false) != mutated.getSizeOfImpact(false)){
+            log.info("impacts should be same before updating")
```
arcuri82, line 262: shouldn't this be a warn? is this a possible bug in EM? or something that could happen in regular scenario?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/util/ParamUtil.kt
ZhangMan1126: actually, the `keys` of `candiates` are rest actions in `call`. here I want to bind the values based on the map provided by `candidates`.

core-it/src/test/kotlin/org/evomaster/core/problem/rest/service/resource/ResourceTestBase.kt
```diff
@@ -306,45 +306,44 @@ abstract class ResourceTestBase : ExtractTestBaseH2(), ResourceBasedTestInterfac
...
         assert(individual.getResourceCalls()[1].getResourceNode().getName() == first.getName())
         assert(individual.getResourceCalls()[0].getResourceNode().getName() == second.getName())
 
         val previousIndividual = individual.copy() as RestIndividual
         val delSpec = MutatedGeneSpecification()
-        structureMutator.mutateRestResourceCalls(individual, RestResourceStructureMutator.MutationType.DELETE, delSpec)
-        assert(delSpec.mutatedPosition.size == 1)
-        assert(delSpec.removedGene.isNotEmpty() || previousIndividual.getResourceCalls()[delSpec.mutatedPosition.first()].actions.flatMap { it.seeGenes() }.isEmpty())
+        structureMutator.mutateRestResourceCalls(individual,  RestResourceStructureMutator.MutationType.DELETE, delSpec)
+        assert(delSpec.getRemoved(true).mapNotNull { it.resourcePosition }.toSet().size == 1)
```
arcuri82, line 327: all these assertions seem expensive, as operations like `mapNotNull` and `toSet` creates new collections. these should be `Lazy` evaluated

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceCluster.kt
```diff
@@ -0,0 +1,83 @@
+package org.evomaster.core.problem.rest.resource
+
+import org.evomaster.client.java.controller.api.dto.database.operations.DataRowDto
+import org.evomaster.core.EMConfig
+import org.evomaster.core.database.SqlInsertBuilder
+import org.evomaster.core.database.schema.Table
+import org.evomaster.core.problem.rest.RestCallAction
+import org.evomaster.core.search.Action
+
+class ResourceCluster {
```
arcuri82, line 10: add comment to clarify what this class is for.

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceCluster.kt
```diff
@@ -0,0 +1,83 @@
...
+     * key is table name
+     * value is the table
+     */
+    private val tables : MutableMap<String, Table> = mutableMapOf()
+
+
+    /**
+     * init resource nodes based on [actionCluster] and [sqlInsertBuilder]
+     */
+    fun intResourceCluster(actionCluster : Map<String, Action>, sqlInsertBuilder: SqlInsertBuilder? = null, config: EMConfig) {
```
arcuri82, line 35: `int` or `init`?

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceCalls.kt
```diff
@@ -41,29 +62,104 @@ class RestResourceCalls(
...
+     * @return the mutable SQL genes and they do not bind with any of Rest Actions
+     *
+     * */
+    private fun seeMutableSQLGenes() : List<out Gene> = getResourceNode().getMutableSQLGenes(dbActions, getRestTemplate(), is2POST)
+
+    fun repairGenesAfterMutation(mutatedGene: MutatedGeneSpecification?, cluster: ResourceCluster){
+
+        mutatedGene?: log.warn("repair genes of resource call ({}) with null mutated genes", getResourceNode().getName())
+
+        val boundBaseGenes = if (mutatedGene?.didStructureMutation() == true) longestPath().seeGenes() else mutatedGene?.getMutated(true)?.mapNotNull { it.gene } ?: longestPath().seeGenes()
```
arcuri82, line 94: the concept of dealing with bounded-genes should be independent of REST. so likely this code needs to be refactored out in a way that can be used in different problems. eg, maybe inside the mutator operator (which already does some post-processing if i remember correctly)

arcuri82: also, didn't we discuss about having `BoundedGene` as a new gene?

arcuri82 (changes requested):

core-it/src/test/kotlin/org/evomaster/core/problem/rest/service/resource/ResourceTestBase.kt
ZhangMan1126: now it is replaced by `distinctBy`

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/ResourceCluster.kt
ZhangMan1126: the comment is added.

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceCalls.kt
ZhangMan1126: yes. this would be fixed with another branch.

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -36,8 +36,36 @@ abstract class Individual(trackOperator: TrackOperator? = null, index : Int = DE
...
+
+        /**
+         * actions which are not SQL-related actions
+         */
+        NO_SQL
+    }
 
+    /**
+     * @return actions based on the specified [filter]
+     */
```
arcuri82, line 68: add a TODO specifying it needs to be implemented

core-it/src/test/kotlin/org/evomaster/core/problem/rest/service/resource/ResourceTestBase.kt
arcuri82: hi. sorry, there was a misunderstanding... :)
i didn't see it was in `test` folder.
`assert` is to define invariants in the SUT (eg, under `main` folder).
in tests, use the `assert*` from JUnit, eg, `assertEquals`, and not the function `assert` from Kotlin.
when invariant are expensive, to do not slow down the SUT, we use `Lazy.assert`, which is not needed in the tests

PR #308, opened 2021-05-11
Looking up the path can be problematic. In the case of language tools…

bogdanmarculescu:
When looking up the supported responses (and respective code), the OpenAPI specification sometimes excludes some prefixes. 

In the case of Language tools, this was "/v2". 
Other APIs may have different prefixes that need exclusion. 

The approach at the moment is to try to get the path as is. If null, try to remove common prefixes until a non-null path is found (or we run out of common prefixes). 

This should solve the problem for the moment, though a long term solution might still be needed.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
@@ -10,8 +10,12 @@ import org.evomaster.core.problem.httpws.service.HttpWsCallResult
 import org.evomaster.core.problem.rest.RestIndividual
 import org.evomaster.core.search.EvaluatedAction
 import org.evomaster.core.search.EvaluatedIndividual
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 abstract class ImplementedOracle {
+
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
```diff
@@ -75,9 +75,13 @@ abstract class ImplementedOracle {
...
      * OpenAPI standards.
+     *
+     * The same applies where the prefix is "v2" (e.g. language tools).
      */
+
     fun retrievePath(objectGenerator: ObjectGenerator, call: RestCallAction): PathItem? {
         return objectGenerator.getSwagger().paths.get(call.path.toString()) ?:
-        objectGenerator.getSwagger().paths.get(call.path.toString().removePrefix("/api"))
+        objectGenerator.getSwagger().paths.get(call.path.toString().removePrefix("/api")) ?:
+        objectGenerator.getSwagger().paths.get(call.path.toString().removePrefix("/v2"))
```
arcuri82, line 85: I m a bit skeptical about these adhoc hacks.
the handling of "basePath" has changed between V2 and V3 versions of OpenApi, but it is still info available from the schema, eg, see addActionsFromSwagger

arcuri82 (changes requested):

PR #310, opened 2021-05-14
impact warning

ZhangMan1126:
only add mutated genes in history if the action exists for httpwsindividual


core/src/main/kotlin/org/evomaster/core/search/service/mutator/StandardMutator.kt
```diff
@@ -258,16 +259,18 @@ open class StandardMutator<T> : Mutator<T>() where T : Individual {
 
 
             additionInfo!!.effectiveHistory.addAll(effective.mapNotNull {
-                if (it.individual.seeActions(isDb).isEmpty())
-                    ImpactUtils.findMutatedGene(it.individual.seeGenes(), gene, includeSameValue)
-                else
+                if (it.individual.seeActions(isDb).isEmpty()){
+                    if(it.individual is HttpWsIndividual) null
```
arcuri82, line 263: i m not sure i understand the reasoning here... if something is `HttpWsIndividual`, then we do not collect history? but shouldn't the mutation of genes be independent from the type of individual? (as more types can be added in the future)

arcuri82 (changes requested):

ZhangMan1126:
to make it more generic, `hasAnyAction() ` is added in the individual

PR #321, opened 2021-06-01
Adjusted the path lookup. 

bogdanmarculescu:
There should be no multiple paths at all anymore. However, I've left the code handling multiple paths (to make the thing more resilient and prevent any crashes).

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
@@ -8,6 +8,7 @@ import org.evomaster.core.output.ObjectGenerator
 import org.evomaster.core.output.TestCase
 import org.evomaster.core.problem.rest.RestCallAction
 import org.evomaster.core.problem.httpws.service.HttpWsCallResult
+import org.evomaster.core.problem.rest.RestActionBuilderV3
 import org.evomaster.core.problem.rest.RestIndividual
 import org.evomaster.core.search.EvaluatedAction
 import org.evomaster.core.search.EvaluatedIndividual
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/evomaster/core/output/oracles/ImplementedOracle.kt
```diff
@@ -85,8 +87,16 @@ abstract class ImplementedOracle {
      */
 
     fun retrievePath(objectGenerator: ObjectGenerator, call: RestCallAction): PathItem? {
+        val serverUrl = objectGenerator.getSwagger().servers[0].url
+        val basePath: String = try {
+            URI(serverUrl).path.trim()
+        } catch (e: URISyntaxException){
+            LoggingUtil.uniqueWarn(log, "Invalid URI used in schema to define servers: $serverUrl")
+            ""
+        }
```
arcuri82, line 96: this is copied&pasted from `addActionsFromSwagger`.
this piece of code should be refactored out in a static function, and then use such function in both places to avoid code-duplication

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -728,4 +723,14 @@ object RestActionBuilderV3 {
         }
     }
 
+    fun getBasePathFromURL(serverUrl: String): String {
```
arcuri82, line 726: this should take as input the `swagger` object, and not the `serverUrl`.
point is that choosing `.servers[0]` is now an hack, as in the future might want to support other indices as well

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -723,7 +723,8 @@ object RestActionBuilderV3 {
         }
     }
 
-    fun getBasePathFromURL(serverUrl: String): String {
+    fun getBasePathFromURL(swagger: OpenAPI): String {
+        val serverUrl = swagger.servers[0].url
```
arcuri82, line 727: the comment:
`TODO would need more general approach, as different HTTP servers could
            have different base paths`
should be moved here

arcuri82 (changes requested):

PR #325, opened 2021-06-10
copy refactor 

ZhangMan1126:
add `StructuralElement` which ables  to get e.g., `children` and `parenet`
refactor `copy()` of `gene`, `param`, `action`, `RestResourceCall`, and `Individual` which inherit `StructuralElement`.
refactor `copyValueFrom()`  `randomize` under `gene`
add `repair()` under `gene`
add `randomize` under `action`

manipulate binding genes under `gene`

clean `refCache` and `dtoCache` in `RestActionBuilderV3`.kt

skip `TaintMultiEMTest` on CI due to _Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000616680000, 1243611136, 0) failed; error='Cannot allocate memory' (errno=12)_


.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs: # a collection of steps
       # do the actual build
       - run:
           command: mvn clean install --fae
-          no_output_timeout: 20m
+          no_output_timeout: 1h
```
arcuri82, line 41: why this?

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -15,12 +16,28 @@ import org.evomaster.core.search.tracer.TrackOperator
  * to a RESTful API, SQL operations on a database or WireMock setup)
  *
  */
-abstract class Individual(trackOperator: TrackOperator? = null, index : Int = DEFAULT_INDEX) : TraceableElement(trackOperator, index){
+abstract class Individual(override var trackOperator: TrackOperator? = null,
+                          override var index: Int = Traceable.DEFAULT_INDEX,
+                          children: List<out StructuralElement>
+) : Traceable, StructuralElement(children){
```
arcuri82, line 22: please provide some javadocs/comments for each of these input parameters

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -15,12 +16,28 @@ import org.evomaster.core.search.tracer.TrackOperator
...
+
+    override var evaluatedResult: EvaluatedMutation? = null
+
+    override var tracking: TrackingHistory<out Traceable>? = null
```
arcuri82, line 27: `evaluatedResult` and `tracking`? why here? we do have already a class for `EvaluatedIndividual` . if really needed here, add comments to explain it

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -0,0 +1,135 @@
...
+        child.parent = this
+    }
+
+    open fun addChildren(children : List<StructuralElement>){
+        initChildren(children)
+    }
+
+    /**
+     * make a deep copy on the content
+     */
```
arcuri82, line 46: add bit more explanation on why/how this is different from `copy`

arcuri82 (changes requested):

.circleci/config.yml
ZhangMan1126: CI fails due to 'Too long with no output (exceeded 20m0s): context deadline exceeded' (see https://app.circleci.com/pipelines/github/EMResearch/EvoMaster/2078/workflows/4dc9592d-fdc8-4757-8999-dd2add437c31/jobs/2149)

arcuri82: and? if a test hangs, then build should fail. moving from 20m to 60m just makes it fail 40m later...
do we have any case of a test that would run for more than 20m?

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
arcuri82: what does it mean `do not recommend`? are we tracking in both places? what would be the case of `evaluatedIndividual` not been tracked, but `individual` is?

arcuri82: or are we not tracking in `evaluatedIndividual`?

arcuri82 (changes requested):

.circleci/config.yml
ZhangMan1126: I changed this back, let's see whether it will fail.

PR #328, opened 2021-06-16
referType in RestActionBuilderV3.kt

ZhangMan1126:
fix missing referType in ObjectGene
add tests for openAPI v2 and v3

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
@@ -110,7 +110,7 @@ object RestActionBuilderV3 {
      * Create an [ObjectGene] based on schema info of a DTO, given in the format
      * "name: {...}"
      */
-    fun createObjectGeneForDTO(name: String, dtoSchema: String) : Gene{
+    fun createObjectGeneForDTO(name: String, dtoSchema: String, referenceTypeName: String?) : Gene{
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -110,7 +110,7 @@ object RestActionBuilderV3 {
      * Create an [ObjectGene] based on schema info of a DTO, given in the format
      * "name: {...}"
      */
-    fun createObjectGeneForDTO(name: String, dtoSchema: String) : Gene{
+    fun createObjectGeneForDTO(name: String, dtoSchema: String, referenceTypeName: String?) : Gene{
```
arcuri82, line 113: add comment to specify what `referenceTypeName` is, and what it is needed for

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
ZhangMan1126: hi, Based on the OpenAPI, the `name` of `ObjectGene` in RestAction is _body_. For the object gene from DTO, I am not sure whether the `name` of DTO can be always recognized as `refType`. please see the TODO at line 170 in `AbstractRestFitness.kt`.  if it is always true, I might remove refereneTypeName for `createObjectGeneForDTO`

PR #329, opened 2021-06-16
Basic assertions fix

bogdanmarculescu:
Added some tests checking that the "--enableBasicAssertions" option works properly.

There was a small bug where the semicolon (if the option was turned off) was added in a comment, leading to compilation errors (fixed). 

The tests check that the tests are correctly generated for Kotlin and Java (I focused more on Java, since that is where the bugs was). 

I also check that the basic assertions are being generated appropriately (marked by the ".assertThat()" marker. 

The actual contents of the assertions are not being evaluated at the moment (only their presence).


arcuri82:
after the merge of js-assertions branch, this one has now conflicts that must be resolved

PR #332, opened 2021-06-23
Assertions update

bogdanmarculescu:
Added the Kotlin version of for-assertions-api. Also consolidated several assertions related EM Tests into a single one, that includes generating Java and Kotlin code, and turns basic assertions on and off.


e2e-tests/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/assertions/AssertionEMTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        val outputFolderName = "AssertionEM"
+        val className = ClassName("org.foo.AssertionEMOff")
+
+        testRunEMGeneric(false, className)
+
+        val assertion = Files.lines(Paths.get(outputFolderPath(outputFolderName)
+                + "/"
+                + className.getBytecodeName()
+                + ".kt")).noneMatch { l: String -> l.contains(".assertThat()") }
+
```
arcuri82, line 42: this code is copied&pasted several times. should be refactored in a common function in the superclass, or test utility class

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/assertions/AssertionEMTest.kt
bogdanmarculescu: I've included it as a separate function in the current test class. Would it be used enough to require a separate test utility class?

PR #334, opened 2021-06-25
Refactor resource call

ZhangMan1126:
refactor the resource-based solution
gene bindings within resource handing or dependent resources handing
add option to skip failed sql commands in the generated tests


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/operations/InsertionResultsDto.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class InsertionResultsDto {
+
+    /**
+     * primary key generated with the insertions
+     */
```
arcuri82, line 12: specify exactly what are the `key` and `value` in this map

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/operations/InsertionResultsDto.java
```diff
@@ -0,0 +1,19 @@
...
+    public Map<Long, Long> idMapping = new HashMap<>();
+
+    /**
+     * whether the insertion at the index executed successfully
+     */
+    public List<Boolean> executionResults = new ArrayList<>();
```
arcuri82, line 18: specify that the list is ordered by index... of what exactly? the actions in the tests? or a specific series of SQL commands?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/database/DbAction.kt
```diff
@@ -52,6 +52,10 @@ class DbAction(
         }
     }
 
+    /**
+     * whether the insertion was executed successfully
+     */
+    var insertExecutedSuccessfully = false
```
arcuri82, line 58: didn't we discuss that this should not be in the action itself, but on the result? 
or was this supposed to be refactored at a later stage (but, in this, it misses a TODO comment)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/database/DbAction.kt
ZhangMan1126: I will refactor this later. now the TODO is added.

PR #349, opened 2021-07-19
Js pure analysis

ZhangMan1126:


client-js/evomaster-client-js/tests/unit/instrumentation/babel-code-test.ts
```diff
@@ -400,5 +400,163 @@ test("ternary throw", () => {
 });
 
 
+test("purity analysis 'and' and 'or' with literal/binary expression", () => {
```
arcuri82, line 403: can you add tests to check out non-pure cases like:

assignment: eg, `if(x=42)`
increment/decrement: eg, `if(x++)`

arcuri82 (changes requested):

client-js/evomaster-client-js/tests/unit/instrumentation/babel-code-test.ts
ZhangMan1126: tests are added.

PR #352, opened 2021-07-20
Handle stopping action results 

ZhangMan1126:
the problem is identified by js case study which has the timeout action.
**Note that the branch is from the pure analysis.**
**Please handle it after pure analysis.**

In addition, fix a problem in js test writer and add a test

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
@@ -594,8 +594,8 @@ abstract class HttpWsTestCaseWriter : WebTestCaseWriter() {
             val k = when{
                 //TODO should do check for when there are spaces in the field name
                 format.isJavaOrKotlin() -> if(fieldPath.isEmpty()) "" else "'$fieldPath'."
-                format.isJavaScript() -> if(fieldPath.isEmpty()) "" else ".$fieldPath"
-                format.isCsharp()  -> if(fieldPath.isEmpty()) "" else ".$fieldPath"
+                format.isJavaScript() -> if(fieldPath.isEmpty()) "" else "${if(fieldPath.startsWith("["))"" else "."}$fieldPath"
+                format.isCsharp()  -> if(fieldPath.isEmpty()) "" else "${if(fieldPath.startsWith("["))"" else "."}$fieldPath"
                 else -> throw IllegalStateException("Format not supported yet: $format")
             }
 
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -594,8 +594,8 @@ abstract class HttpWsTestCaseWriter : WebTestCaseWriter() {
             val k = when{
                 //TODO should do check for when there are spaces in the field name
                 format.isJavaOrKotlin() -> if(fieldPath.isEmpty()) "" else "'$fieldPath'."
-                format.isJavaScript() -> if(fieldPath.isEmpty()) "" else ".$fieldPath"
-                format.isCsharp()  -> if(fieldPath.isEmpty()) "" else ".$fieldPath"
+                format.isJavaScript() -> if(fieldPath.isEmpty()) "" else "${if(fieldPath.startsWith("["))"" else "."}$fieldPath"
```
arcuri82, line 597: what is this for? can you add comment to clarify (eg with example)?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestFitness.kt
```diff
@@ -74,6 +74,12 @@ open class RestFitness : AbstractRestFitness<RestIndividual>() {
 
             if (a is RestCallAction) {
                 ok = handleRestCall(a, actionResults, chainState, cookies, tokens)
+                /*
+                    the action might be stopped due to e.g., timeout (see [handleRestCall]),
+                    but the property of [stopping] is not handle.
+                    we can also handle the property inside [handleRestCall]
+                 */
+                actionResults.filterIsInstance<RestCallResult>()[i].stopping = !ok
```
arcuri82, line 82: no, this is not what `stopping` was used for.
see `AbstractRestFitness.handleSaveLocation`

a failed call (eg due to timeout) does NOT prevent making the following calls in the test

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestFitness.kt
ZhangMan1126: yes. I found that it is the only place where to handle the property `stopping`. but in the loop of executing rest action, if the `ok` is false, the following actions are not executed (see lines 75 and 88 in `RestFitness`).  do we add another property to control if the following actions from the current were executed?

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
ZhangMan1126: I added a test `testTestWithObjectAssertion` in `TestCaseWriterTest`. 
for example,  `expect(Object.keys(res_0.body[0]).length).toBe(0);` (see line 1294) would be `expect(Object.keys(res_0.body**.**[0]).length).toBe(0);` with the previous implementation.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestFitness.kt
arcuri82: hi. hmmm... you might be right. but i need to think a bit more about this

PR #355, opened 2021-07-26
Gql utility function get stats

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
```diff
@@ -0,0 +1,46 @@
...
+     *        ... on InterfaceObject1 {
+     *           field
+     *        }
+     *        ... on InterfaceObjectN {
+     *          field
+     *        }
+     *     }
+     */
+    const val INTERFACE_BASE_TAG = "#BASE#"
+    const val INTERFACE_TAG = "#INTERFACE#"
```
arcuri82, line 31: what is the difference between `INTERFACE_BASE_TAG` and `INTERFACE_TAG`?

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
```diff
@@ -0,0 +1,46 @@
...
+    const val SCALAR_TAG = "scalar"
+    const val OBJECT_TAG = "object"
+    const val UNION_STRING_TAG = "union"
+    const val INTERFACE_STRING_TAG = "interface"
+    const val ENUM_TAG = "enum"
+    const val LIST_TAG = "list"
+    const val INPUT_OBJECT_TAG = "input_object"
+    /**/
```
arcuri82, line 39: why some `TAG` are in uppercase surrounded by `##` (eg `#BASE#`), whereas others are in lower case (eg `list`).
be consistent.
if those have completely different use in the code, then do not use the same `_TAG` suffix

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
```diff
@@ -7,12 +7,34 @@ import org.evomaster.core.problem.graphql.param.GQReturnParam
...
 import org.evomaster.core.search.gene.*
 import org.slf4j.LoggerFactory
+import java.util.*
 import javax.ws.rs.client.Entity
 
 object GraphQLUtils {
 
     private val log = LoggerFactory.getLogger(GraphQLUtils::class.java)
 
+    val history: Deque<String> = ArrayDeque<String>()
```
arcuri82, line 17: a public mutable static state? this needs to be refactored

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
asmab89: eg:
let s take this schema:
```
interface Book {
  title: String!
}

type Textbook implements Book {
  title: String!
  courses: String!
}

type ColoringBook implements Book {
  title: String!
  colors: [String!]!
}

type Query {
  books: [Book!]!
}
```
And this Query:
```
{
  books {
    title
    ... on Textbook {
      courses # Only present in ColoringBook
    }
    ... on ColoringBook {
      colors # Only present in ColoringBook
    }
            }
}
```
INTERFACE_BASE_TAG: is used to annotate the interface type Book (so we can construct all the interface fields, like: title ) and
INTERFACE_TAG: is used to annotate the query field name books (which has the type: interface) (so we can construct ... on Textbook and ... on ColoringBook)

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
asmab89: Indeed they are different. those surrounded by `## (eg #BASE#)`  are for a print function and the others `(in lower case(eg list))` are frequently used strings. 
Suffix changed.

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
asmab89: Did not understand this one.

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
arcuri82: add this explanations as comments

asmab89: Done

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
asmab89: Already removed.

asmab89:
Done

PR #363, opened 2021-08-13
Dotnet instrumentation

amidgol:


client-dotnet/src/EvoMaster.Instrumentation/Heuristic/TruthnessUtils.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        if(!Double.IsFinite(v) || v == Double.MaxValue){
+            return 1d;
+        }
+
+        //normalization function from old ICST/STVR paper
+        double normalized = v / (v + 1d);
+
+        //assert normalized >= 0 && normalized <= 1; TODO
```
arcuri82, line 24: in C#, can't this be handled with `Debug.Assert` ?

client-dotnet/src/EvoMaster.Instrumentation/StaticState/ObjectiveRecorder.cs
```diff
@@ -0,0 +1,261 @@
...
+     * However, it is also important to notice that which classes
+     * are loaded depends on what is executed.
+     * We can force the loading of all classes, but usually that
+     * is not a good idea, as static initializers might have
+     * side effects.
+     * However, we can do that at the end of the search once we are
+     * done.
+     * This can be useful to calculate how many targets we have missed.
+     */
+        private static readonly HashSet<string> AllTargets = new HashSet<string>();
```
arcuri82, line 41: is this set thread safe?

client-dotnet/src/EvoMaster.Instrumentation/StaticState/ObjectiveRecorder.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        /**
+     * Key -> id of an objective/target
+     * <br>
+     * Value -> a mapped unique id in numeric format
+     * <br>
+     * Note: we need this mapping to reduce the id size,
+     * as to reduce TCP bandwidth consumption when communicating
+     * with the EvoMaster process
+     */
```
arcuri82, line 53: please fix indentation of (all) comments

client-dotnet/src/EvoMaster.Instrumentation/StaticState/ObjectiveRecorder.cs
```diff
@@ -0,0 +1,261 @@
...
+                throw new ArgumentException("Empty target name");
+            }
+
+            AllTargets.Add(target);
+        }
+
+        /**
+     * @return a coverage value in [0,1]
+     */
+        public static double? ComputeCoverage(string prefix)
```
arcuri82, line 137: `double?` means nullable, isn't it? why so then?

client-dotnet/test/EvoMaster.Instrumentation.Tests/Examples/Triangle/LineCovTcTest.cs
```diff
@@ -0,0 +1,24 @@
+using EvoMaster.Instrumentation.Examples.Triangle;
+using EvoMaster.Instrumentation.Staticstate;
+using Xunit;
+
+namespace EvoMaster.Instrumentation.Tests.Examples.Triangle
+{
+    public class LineCovTcTest
+    {
+        [Fact]
+        public void TestLineCov()
```
arcuri82, line 10: where is the instrumentation of the DLL executed? we ll need to discuss it

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/staticstate/ObjectiveRecorder.java
```diff
@@ -83,7 +83,6 @@
      */
```
arcuri82, line 83: why modifying this file??? or was it by mistake?

arcuri82 (changes requested):

client-dotnet/src/EvoMaster.Client.Util/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Collections.Generic;
+
+namespace EvoMaster.Client.Util.Extensions
+{
```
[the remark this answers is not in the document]
amidgol: It is done by the code beautifier of the IDE. However it can be as you described but it should be done manually for all of the files.

client-dotnet/src/EvoMaster.Instrumentation/Heuristic/TruthnessUtils.cs
amidgol: I handled this with Trace.Assert which works in release mode unlike Debug.Assert which works in debug mode. Is it ok?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/staticstate/ObjectiveRecorder.java
amidgol: Sorry it was by mistake

client-dotnet/src/EvoMaster.Client.Util/ConcurrentHashSet.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace EvoMaster.Client.Util
+{
```
arcuri82, line 7: as previously discussed, { should not be on their own empty line :)

client-dotnet/src/EvoMaster.Client.Util/ConcurrentHashSet.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class ConcurrentHashSet<T> : ICollection<T>, IReadOnlyCollection<T>
```
arcuri82, line 8: what is the need for this? 
why not using that `ConcurrentDictionary` directly?

client-dotnet/src/EvoMaster.Client.Util/Extensions/DictionaryExtensions.cs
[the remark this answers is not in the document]
arcuri82: yep, deactivate/change that config in the IDE, and re-update all these files

client-dotnet/src/EvoMaster.Instrumentation-Shared/ClassName.cs
```diff
@@ -0,0 +1,104 @@
...
+            if (name.EndsWith(".class"))
+            {
+                name = name.Substring(0, name.Length - ".class".Length);
+            }
+
+            if (name.EndsWith(".java"))
+            {
+                name = name.Substring(0, name.Length - ".java".Length);
+            }
+
```
arcuri82, line 62: the code above is specific for Java. i guess should need something equivalent for .Net. 
also need to check if this is even needed in .Net

client-dotnet/src/EvoMaster.Instrumentation-Shared/ClassName.cs
```diff
@@ -0,0 +1,104 @@
...
+        public string GetBytecodeName()
+        {
+            return _bytecodeName;
+        }
+
+        /// <summary>Eg, foo.bar.Hello</summary>
+        public string GetFullNameWithDots()
+        {
+            return _fullNameWithDots;
+        }
```
arcuri82, line 91: ie, continuing from previous comment: does .Net make a difference between class names in the bytecode and class name in source code (use of "/" vs ".")? or is everything based on just the namespace?

client-dotnet/src/EvoMaster.Instrumentation/Coverage/MethodReplacement/DistanceHelper.cs
```diff
@@ -0,0 +1,219 @@
...
+
+    public static readonly double H_NOT_NULL = 0.1d;
+
+    public static readonly double H_REACHED_BUT_EMPTY = H_REACHED_BUT_NULL;
+
+    public static readonly double H_NOT_EMPTY = H_NOT_NULL;
+
+
+    //2^16=65536, max distance for a char
+    public static readonly int MAX_CHAR_DISTANCE = 65_536;
```
arcuri82, line 15: does this apply to .Net as well? if so, add a comment, eg "Like on JVM, also in .Net a char..."

arcuri82 (changes requested):

client-dotnet/src/EvoMaster.Client.Util/ConcurrentHashSet.cs
amidgol: .NET doesn't have a concurrent HashSet. What I have written is nothing but a wrapper around the ConcurrentDictionary. The value is byte and always zero. I though that it might be meaningless to use a key,value structure here.

arcuri82:
@amidgol  thx
@ZhangMan1126  can u have a look at this PR? after that, I ll review it as well

client-dotnet/src/EvoMaster.Instrumentation/Instrumentator.cs
```diff
@@ -0,0 +1,131 @@
...
+using EvoMaster.Instrumentation.StaticState;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Cecil.Rocks;
+
+namespace EvoMaster.Instrumentation {
+    public class Instrumentator {
+        private MethodReference _probe;
+
+        public void Instrument(string assembly, string destination) {
```
ZhangMan1126, line 14: add a bit comments about the params

client-dotnet/src/EvoMaster.Instrumentation/Instrumentator.cs
```diff
@@ -0,0 +1,131 @@
...
+                    method.Body.OptimizeMacros(); //Change back Br opcodes to Br.s if possible
+                }
+            }
+
+            if (destination.Length > 1 && destination[^1] == '/') {
+                destination = destination.Remove(destination.Length - 1, 1);
+            }
+
+            module.Write($"{destination}/InstrumentedSut.dll");
+            Console.WriteLine($"Instrumented File Saved at \"{destination}\"");
```
ZhangMan1126, line 97: may use log instead of console print

client-dotnet/src/EvoMaster.Instrumentation_Shared/ObjectiveNaming.cs
```diff
@@ -0,0 +1,192 @@
...
+            var m0 =
+                CacheSuccessCall.ComputeIfAbsent(className,
+                    c => new ConcurrentDictionary<int, IDictionary<int, string>>()); //TODO: capacity 10_000
+            var
+                m1 = m0.ComputeIfAbsent(line, l => new ConcurrentDictionary<int, string>()); //TODO: capacity 10
+            return m1.ComputeIfAbsent(index, i =>
+                SUCCESS_CALL + "_at_" + ClassName.Get(className).GetFullNameWithDots() +
+                "_" + PadNumber(line) + "_" + index);
+
+//        string name = SUCCESS_CALL + "_at_" + ClassName.get(className).getFullNameWithDots() +
```
ZhangMan1126, line 106: may remove not needed commented code

PR #365, opened 2021-08-16
Assertion bugfix

bogdanmarculescu:
Updated the API for assertions.


arcuri82:
thx. looking at the code: when you generate the Kotlin versions, do you as well compile the generated tests to make sure they are syntactically valid? (it does not seem the kotlin compiler is called in those tests... or did i miss it?)

arcuri82:
@bogdanmarculescu  furthermore, when we have checks like:
`assertHasAtLeastOne(solution, HttpVerb.GET, 200, "/api/assertions/data", "{\"a\":42,\"c\":[1000,2000,3000],\"d\":{\"e\":66,\"f\":\"bar\",\"g\":{\"h\":[\"xvalue\",\"yvalue\"]}},\"i\":true,\"l\":false}")`
that is not checking if those do appear in the generated tests, but just in the HTTP repsonse.
like the tests for JS and C#, need to check in generated tests as well

PR #367, opened 2021-08-18
Js auth token

ZhangMan1126:
- enable loginjson auth for js
- e2e test


client-js/integration-tests/tests/login-token/args.txt
```diff
@@ -0,0 +1 @@
+--testSuiteSplitType=NONE
```
arcuri82, line 1: what is this file?

ZhangMan1126: it is added by mistake. now removed.

PR #374, opened 2021-08-30
handle async when using arrow function

ZhangMan1126:
handle async property of replaced arrow function for logic expression and ternary.


client-js/evomaster-client-js/src/instrumentation/babel-plugin-evomaster.ts
```diff
@@ -153,6 +153,20 @@ export default function evomasterPlugin(
         branchCounter++;
     }
 
+    function doesContainExpression(node: Expression) : boolean{
```
arcuri82, line 156: can u add a comment to explain it? i mean, does it check if something is `asyn`? (and if yes, Asyn should be somehow in the name)

arcuri82 (changes requested):

PR #379, opened 2021-09-03
handle cookie login

ZhangMan1126:
handle cookie login with a different service for ind1 
add e2e tests for cookie login

### DIFF
--- a/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/CookieLoginDto.java
+++ b/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/CookieLoginDto.java
@@ -35,10 +35,6 @@ public class CookieLoginDto {
      */
     public String loginEndpointUrl;
 
-    /**
-     * Whether the [loginEndpointUrl] is a complete URL
-     */
-    public Boolean fullUrl;
 
     public enum HttpVerb{
         GET, POST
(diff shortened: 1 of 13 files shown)


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/CookieLoginDto.java
```diff
@@ -35,6 +35,10 @@
      */
     public String loginEndpointUrl;
 
+    /**
+     * Whether the [loginEndpointUrl] is a complete URL
+     */
+    public Boolean fullUrl;
```
arcuri82, line 41: i don't think there is a need for such a boolean. we can simply analyze the content of `loginEndpointUrl` (eg, by parsing it in a URL object), and see if it is a full absolute URL, or a relative one (ie, no protocol/host/port info)

core/src/main/kotlin/org/evomaster/core/problem/httpws/service/HttpWsFitness.kt
```diff
@@ -292,7 +292,7 @@ abstract class HttpWsFitness<T>: FitnessFunction<T>() where T : Individual {
             }
 
             val response = try {
-                client.target(baseUrl + cl.loginEndpointUrl)
+                client.target(if (cl.fullUrl) cl.loginEndpointUrl else (baseUrl + cl.loginEndpointUrl))
```
arcuri82, line 295: when the loginUrl is relative, should have a check if it starts with "/".
also, need to check if `baseUrl` is not ending already with "/" when concatenating the two

arcuri82 (changes requested):

arcuri82 (approved):

PR #384, opened 2021-09-08
Fix unresolved initclass

ZhangMan1126:
fix cookie login em for updated e2e openapi v3


arcuri82:
hi. thx, but no longer needed. i had a hotfix while in the meeting with Asma :)

PR #386, opened 2021-09-10
Bb schema parsing

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLBlackBoxFitness.kt
```diff
@@ -65,4 +64,16 @@ class GraphQLBlackBoxFitness : GraphQLFitness() {
          */
         return null
     }
+
+    override fun getGraphQLError(
```
arcuri82, line 68: maybe we need a better name... in BB we still have Errors, it is just that we miss the info on the last-executed-line.
Maybe rename with something like `getGraphQLErrorWithLineInfo`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/service/GraphQLBlackBoxFitness.kt
asmab89: Done

PR #388, opened 2021-09-19
Tmp handle ind1

ZhangMan1126:
- supports new column types of MySQL for ind1
- add a config to specify the maximum number of existing data to be sampled for DB handling

- add `min` and `max` for `NumberGene`
- handle `DoubleGene` and `FloatGene` mutation by considering its `min` and `max` if they exist
- handle `precision` for Double and Float
- extract size and precision for a decimal (currently decimal is  handled as `FloatGene`)
- tests are added


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/ColumnDto.java
```diff
@@ -22,4 +22,10 @@
 
     public boolean isUnsigned = false;
 
+    /**
+     * precision of number
+     * -1 means that the precision is unspecified or inapplicable
+     */
```
arcuri82, line 28: do not check for specific values. have something more general like `a negative number (e.g., -1) means...`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1432,6 +1432,13 @@ class EMConfig {
     var skipFailureSQLInTestFile = false
 
 
+    @Experimental
+    @Cfg("Specify a maximum number of existing data in the database to sample when SQL handling is enabled. " +
+            "Note that 0 means all existing data would be sampled")
+    @Min(0.0)
+    var maximumExistingDataToSampleInDb = 0
+
```
arcuri82, line 1440: hmmm... but what is one wants no data to sample (ie 0)?
maybe better like no data is sampled if the value is negative, with -1 being the default

core/src/main/kotlin/org/evomaster/core/database/DbActionGeneBuilder.kt
```diff
@@ -420,7 +421,15 @@ class DbActionGeneBuilder {
             checkNotEmpty(column.enumValuesAsStrings)
             EnumGene(name = column.name, data = column.enumValuesAsStrings.map { it.toFloat() })
         } else {
-            FloatGene(column.name)
+            if (column.precision != -1){
```
arcuri82, line 424: rather check if not negative, ie `>=0`

core/src/main/kotlin/org/evomaster/core/database/schema/Column.kt
```diff
@@ -36,7 +36,12 @@ data class Column(
 
         val databaseType: DatabaseType,
 
-        val isUnsigned : Boolean = false
+        val isUnsigned : Boolean = false,
+
+        /**
+         * -1 means that the precision is unspecified or inapplicable
```
arcuri82, line 42: see previous comments

core/src/main/kotlin/org/evomaster/core/search/gene/DoubleGene.kt
```diff
@@ -49,21 +61,39 @@ class DoubleGene(name: String,
...
-            1 -> value + (getDelta(randomness, apc) * randomness.nextGaussian())
-            //to reduce precision, ie chop off digits after the "."
-            2 -> BigDecimal(value).setScale(randomness.nextInt(15), RoundingMode.HALF_EVEN).toDouble()
-            else -> throw IllegalStateException("Regression bug")
+        /*
+           TODO min/max for double
+           Man: update a bit by considering min and max,
+           NEED a check by Andrea
+        */
+        var gaussianDelta = randomness.nextGaussian()
```
arcuri82, line 69: make sure we have some unit tests for this, and the different branches are covered.

also, once computation is finished, you can have `assert` checks to verifiy the final results in within min/max

core/src/main/kotlin/org/evomaster/core/search/gene/FloatGene.kt
```diff
@@ -46,24 +61,36 @@ class FloatGene(name: String,
...
+            res += (if (gaussianDelta>0) 1.0f else -1.0f) * getMinimalDelta()!!
         }
 
-        value = k.toFloat()
+        value = if (max != null && res > max) max
+                else if (min != null && res < min) min
+                else getFormattedValue(res)
 
         return true
     }
```
arcuri82, line 90: hmmm... here code is similar to Double. is there any way in which code could be re-used? (ie avoiding duplication)

core/src/main/kotlin/org/evomaster/core/search/gene/NumberGene.kt
```diff
@@ -25,4 +33,36 @@ abstract class NumberGene<T : Number>(name: String, var value: T) : Gene(name, m
...
+        return when(randomness.choose(strategies)){
+            ModifyStrategy.SMALL_CHANGE-> value + min(1, maxRange) * delta
+            ModifyStrategy.LARGE_JUMP -> value + specifiedJumpDelta * delta
+            ModifyStrategy.REDUCE_PRECISION -> BigDecimal(value).setScale(randomness.nextInt(15), RoundingMode.HALF_EVEN).toDouble()
+        }
+
+    }
+
+    /**
+     * @return formatted value
```
arcuri82, line 49: this comment just states the name of the method... :)
you can clarify what a "formatted value" is

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/DoubleGene.kt
ZhangMan1126: there is a new unit test `NumberGeneTest.kt`. 

I will add e2e test.

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -418,5 +418,11 @@ abstract class Gene(var name: String, children: List<out StructuralElement>) : S
     open fun repair(){
         //do nothing
     }
+
+    /**
+     * @return whether the gene is valid
+     *  based on any specialized rule for different types of genes if there exist
+     */
```
arcuri82, line 425: can u give an example here in the comments of a case of an invalid gene?

core/src/main/kotlin/org/evomaster/core/search/gene/NumberGene.kt
```diff
@@ -25,4 +30,15 @@ abstract class NumberGene<T : Number>(name: String, var value: T) : Gene(name, m
             value.toFloat()
 
 
+    /**
+     * @return whether the [value] is between [min] and [max] if they are specified
+     */
+    override fun isValid() : Boolean{
+        if (max != null && value.toDouble() > max!!.toDouble())
+            return false
```
arcuri82, line 38: how is that possible? shouldn't this prevented in the sampler/mutator? 
or is this method just used for debugging / assertion writing?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/NumberGene.kt
ZhangMan1126: it is only used for testing purposes to check whether the value is [min, max] when they are specified.

arcuri82: in that case, add such clarification in the comments of the `isValid` declaration in super-class

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/DbCleaner.java
```diff
@@ -85,31 +89,47 @@ public static void clearDatabase(Connection connection, List<String> tablesToSki
...
      * @param statement is to execute the SQL command
      * @param schema specify the schema of data to clean. if [schema] is empty, all data will be cleaned.
      * @param singleCommand specify whether to execute the SQL commands (e.g., truncate table/tables) by single command
      * @param doDropTable specify whether to drop tables which is only for MySQL and MariaDB now.
      * @throws SQLException are exceptions during sql execution
      */
-    private static void cleanDataInTables(List<String> tablesToSkip, Statement statement, DatabaseType type, String schema, boolean singleCommand, boolean doDropTable) throws SQLException {
+    private static void cleanDataInTables(List<String> tableToSkip, List<String> tableToClean, Statement statement, DatabaseType type, String schema, boolean singleCommand, boolean doDropTable) throws SQLException {
+        if (tableToSkip != null && tableToClean != null)
+            throw new IllegalArgumentException("tableToSkip and tableToClean cannot be configured at the same time.");
```
arcuri82, line 131: should also check that those list are not empty

arcuri82 (changes requested):

PR #391, opened 2021-09-22
fix gql error related statistic bugs

asmab89:
add tests for it

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
@@ -153,13 +153,15 @@ class FitnessValue(
     }
 
     fun gqlErrors(idMapper: IdMapper, withLine : Boolean): List<String>{
-        return targets.keys
+        // GQLErrors would be 0.5 when it is initialed, so we count it when it is covered.
+        return targets.filter { it.value.distance == MAX_VALUE }.keys
                 .filter { idMapper.isGQLErrors(it, withLine) }
                 .map { idMapper.getDescriptiveId(it) }
     }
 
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
```diff
@@ -153,13 +153,15 @@ class FitnessValue(
     }
 
     fun gqlErrors(idMapper: IdMapper, withLine : Boolean): List<String>{
-        return targets.keys
+        // GQLErrors would be 0.5 when it is initialed, so we count it when it is covered.
```
arcuri82, line 156: avoid hardcoding values in comments (as those might change in the future). eg, instead of `be 0.5`, write something less brittle like `be >0`.

btw, what about the bug in the counting of queries/mutations?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
asmab89: There was three bugs in calculating statistics:
In distinctActions (the counting of queries/mutations), gqlNoErrors and gqlErrors all of them are fixed now.

arcuri82: was distinctActions  fixed in this branch? or in another one?

asmab89: It was fixed in another branch but I close it to avoid the repeated fix since when Man fix the gqlNoErrors and gqlErrors fix also distinctActions here.

PR #392, opened 2021-10-02
Exp script

arcuri82:


arcuri82:
@ZhangMan1126 hi. besides reviewing if i missed anything, can u add the scripts for the papers u were main author? (i could not find them in the other reps)

ZhangMan1126:
Hi @arcuri82 , scripts for gecco, emse, ssbse, tomse are added. the script is a bit different from the `exp.py`, but it provides descriptions about params to be configured.

PR #401, opened 2021-11-02
Sql handling

ZhangMan1126:
update interpolateSqlString with JSqlParser
remove comments
as investigated, no need to handle prepared statements of zebra

### DIFF
--- a/client-java/instrumentation/pom.xml
+++ b/client-java/instrumentation/pom.xml
@@ -148,6 +148,10 @@
                                     <pattern>dk.brics.automaton</pattern>
                                     <shadedPattern>shaded.dk.brics.automaton</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.github.jsqlparser</pattern>
+                                    <shadedPattern>shaded.com.github.jsqlparser</shadedPattern>
+                                </relocation>
                             </relocations>
                         </configuration>
                     </execution>
--- a/client-java/instrumentation/pom.xml
+++ b/client-java/instrumentation/pom.xml
@@ -34,6 +34,11 @@
             <artifactId>asm-commons</artifactId>
         </dependency>
 
+        <!--  handle sql command      -->
+        <dependency>
+            <groupId>com.github.jsqlparser</groupId>
+            <artifactId>jsqlparser</artifactId>
+        </dependency>
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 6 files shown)


client-java/instrumentation/pom.xml
```diff
@@ -34,6 +34,11 @@
             <artifactId>asm-commons</artifactId>
         </dependency>
 
+        <!--  handle sql command      -->
+        <dependency>
+            <groupId>com.github.jsqlparser</groupId>
+            <artifactId>jsqlparser</artifactId>
+        </dependency>
```
arcuri82, line 41: if you add dependencies to `instrumentation` those must be shaded. see bottom of pom file for `maven-shade-plugin`

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/PreparedStatementClassReplacement.java
```diff
@@ -145,11 +190,9 @@ private static String handlePreparedStatement(PreparedStatement stmt) {
...
         }
 
-        /*
-            handle db middleware, ie, zebra
-         */
+        // just check whether there exist any new update
         if (fullClassName.startsWith("com.dianping.zebra")){
-            sql = extractSqlFromZebraPreparedStatement(stmt);
+            throw new IllegalArgumentException("unsupported type for zebra: " + fullClassName);
         }
```
arcuri82, line 196: why throwing an exception here?  wasn't this checked with `checkZebraPreparedStatementWrapper`?  or should rather be an `assert` here?

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/PreparedStatementClassReplacement.java
ZhangMan1126: all zebra prepared statements should be handled before this line (see 155).
if this line is reached, it means that there exist new types of prepared statements, then throw the exception with the info.

client-java/instrumentation/pom.xml
ZhangMan1126: thanks! updated

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/PreparedStatementClassReplacement.java
arcuri82: ok. can u add this clarification as a code comment?

ZhangMan1126: the comment is modified with more info now. thanks.

PR #402, opened 2021-11-03
Issue 396 fix

bogdanmarculescu:
A fix for the issue. TCs with only 200 calls should not appear in the "_faults" file if expectations are not active.


arcuri82:
unfortunately the handling of `default` seems still broken. for example, in `features-service` we still create tests like:

```
@Test
    public void test_16_with500() throws Exception {
        ExpectationHandler expectationHandler = expectationHandler();
        
        String location_productName = "";
        
        ValidatableResponse res_0 = given().accept("*/*")
                .post(baseUrlOfSut + "/products/mDNayE_5")
                .then()
                .statusCode(201)
                .assertThat()
                .body(isEmptyOrNullString());
        location_productName = res_0.extract().header("location");
        assertTrue(isValidURIorEmpty(location_productName));
        
        expectationHandler.expect(ems)
            /*
             Note: No supported codes appear to be defined. https://swagger.io/docs/specification/describing-responses/.
             This is somewhat unexpected, so the code below is likely to lead to a failed expectation
            */
            .that(sco, Arrays.asList().contains(res_0.extract().statusCode()));
```

where the schema does defines `default` behavior:
```
 "/products/{productName}": {
      "get": {
        "operationId": "getProductByName",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "productName",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "$ref": "#/definitions/Product"
            },
            "headers": {}
          }
        }
      },
      "post": {
        "operationId": "addProduct",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "productName",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "default": {
            "description": "successful operation"
          }
        }
      },
```

PR #405, opened 2021-11-04
ignore init sql

ZhangMan1126:
ignore initialization sql in sql heuristics


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/EMController.java
```diff
@@ -507,6 +509,8 @@ public Response executeDatabaseCommand(DatabaseCommandDto dto, @Context HttpServ
 
         try {
 
+            ExecutionTracer.executingInitSql = true;
```
arcuri82, line 512: unfortunately, I just realized this approach does not work with External Drivers :(  
we need to discuss this

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/EMController.java
ZhangMan1126: ok.

arcuri82:
thanks. unfortunately we do not have (yet) proper E2E scaffolding for external drivers :(  just make sure it seems working with some manual testing using actual external drivers (eg from EMB)

client-java/controller/src/main/java/org/evomaster/client/java/controller/ExternalSutController.java
```diff
@@ -364,6 +364,13 @@ public void setKillSwitch(boolean b) {
         ExecutionTracer.setKillSwitch(b);// store info locally as well, to avoid needing to do call to fetch current value
     }
 
+    @Override
+    public void setExecutingInitSql(boolean executingInitSql) {
+        checkInstrumentation();
+        serverController.setExecutingInitSql(executingInitSql);
+        // Man: why there are two set, need to check
+        ExecutionTracer.setKillSwitch(executingInitSql);
```
arcuri82, line 372: :)

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/EMController.java
```diff
@@ -507,6 +508,8 @@ public Response executeDatabaseCommand(DatabaseCommandDto dto, @Context HttpServ
 
         try {
 
+            sutController.setExecutingInitSql(true);
```
arcuri82, line 511: hmmm, i have been wondering, is this even necessary? all the SQL commands here are executed in `SqlScriptRunner`, which is not instrumented. so in theory they commands should be not recorded, regardless of this code, isn't it?

arcuri82 (changes requested):

ZhangMan1126:
checked it with rest-news using external driver. SQL for initialization could be ignored correctly.

PR #417, opened 2021-11-19
Add dependency.

asmab89:


e2e-tests/spring-graphql/pom.xml
```diff
@@ -49,6 +49,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-data-jpa</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-validation</artifactId>
+            <version>2.5.0</version>
```
arcuri82, line 55: recall [https://github.com/EMResearch/EvoMaster/blob/master/docs/for_developers.md#maven-dependency-version](). version declaration should be in root pom. if u are not sure how Maven work, we can chat about this next meeting

arcuri82 (changes requested):

e2e-tests/pom.xml
```diff
@@ -32,6 +32,11 @@
                 <scope>test</scope>
                 <type>test-jar</type>
             </dependency>
+            <dependency>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-starter-validation</artifactId>
+                <version>2.5.0</version>
+            </dependency>
```
arcuri82, line 39: this is not the main root pom file. furthermore, the version should be `${springboot.version}` as all other dependencies

asmab89: Done

e2e-tests/spring-graphql/pom.xml
asmab89: Done

PR #434, opened 2022-01-14
RPC

ZhangMan1126:
main features:
- enable evomaster with RPC SUT
- handle constraints with javax.validation.contraints
- extend genes with constraints
- extend map with enum key
- handle auth info,  eg with RPC endpoints or specified values in requests
- enable customized seeds, also enable them during sampling and mutation
- test generations with driver (kotlin and java)
- test generations with native instances (only java)
- assertions based on responses (only java)
- max size of assertions for data in collection 
- handle the auth with an interface provided by SutHandler

others
- unit tests for RPC problems
- RPC e2e tests for authsetup, branches, customization (seeding), db, hypermutation, taint, taintignorecase, taintinvalid, testability, thriftexception


arcuri82:
hi.
 thx. looks like there is conflicts. please resolve them, and then add me again as a reviewer

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
```diff
@@ -79,6 +79,67 @@ default void setupForGeneratedTest(){}
...
+     * This method must be blocking until the SUT is initialized.
+     * </p>
+     *
+     * <p>
+     * This method is only required when the problem is RPC for the moment,
+     * otherwise return null
+     * </p>
+     *
+     * TOCHECK with Andrea if it is a proper place where we put this method
+     * this method is required
```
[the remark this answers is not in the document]
arcuri82: aslo, why returning an `Object`, instead of something like:
`default <T,T extends X> T getRPCClient(Class<X> x)`? 
(not sure on the syntax here on GitHub editor)

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
```diff
@@ -79,6 +79,67 @@ default void setupForGeneratedTest(){}
...
+
+    /**
+     * <p>
+     * execute an RPC endpoint with evomaster driver
+     * </p>
+     *
+     * TOCHECK with Andrea if it is a proper place where we put this method
+     * this method is required
+     */
+    default void extractRPCSchema(){}
```
arcuri82, line 131: but this does not return anything, ie, `void`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/EMController.java
```diff
@@ -196,6 +202,29 @@ public Response getSutInfo(@Context HttpServletRequest httpServletRequest) {
             GraphQlProblem p = (GraphQlProblem) info;
             dto.graphQLProblem = new GraphQLProblemDto();
             dto.graphQLProblem.endpoint= removePrefix(p.getEndpoint(), baseUrlOfSUT);
+        } else if(info instanceof RPCProblem){
+            dto.rpcProblem = new RPCProblemDto();
+            // extract RPCSchema
+            sutController.extractRPCSchema();
+            Map<String, InterfaceSchema> rpcSchemas = sutController.getRPCSchema();
+            if (rpcSchemas == null){
```
arcuri82, line 210: `|| rcpSchemas.isEmpty()` ?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -57,6 +70,22 @@
      */
     private final List<ExtraHeuristicsDto> extras = new CopyOnWriteArrayList<>();
 
+    /**
+     * a map of interface schemas for RPC service under test
+     */
+    private final Map<String, InterfaceSchema> rpcInterfaceSchema = new HashMap<>();
```
arcuri82, line 76: to avoid issues with non-determinism, prefer `LinkedHashMap` over `HashMap`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -57,6 +70,22 @@
...
+
+    /**
+     * a map of local auth setup schemas for RPC service under test
+     */
+    private final Map<Integer, LocalAuthSetupSchema> localAuthSetupSchemaMap = new HashMap<>();
```
arcuri82, line 81: what is the key here?

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/CodeJavaGenerator.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * a set of util for generating instance creation and assertion with java in tests
+ * in order to make endpoint invocation
+ */
+public class CodeJavaGenerator {
```
arcuri82, line 12: why was this necessary? i mean, why need of Java code on the driver? 
can't the API be called via reflection?

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/schema/params/ByteBufferParam.java
```diff
@@ -0,0 +1,98 @@
...
+            setValue(dto.stringValue.getBytes());
+    }
+
+    @Override
+    protected void setValueBasedOnValidInstance(Object instance) {
+        setValue((ByteBuffer) instance);
+    }
+
+    @Override
+    public List<String> newInstanceWithJava(boolean isDeclaration, boolean doesIncludeName, String variableName, int indent) {
```
arcuri82, line 63: so, this is about the discussion we had, in which currently the test case is generated on the Driver side instead of the Core?

client-java/controller/src/test/java/com/thrift/example/real/facebook/fb303/FacebookService.java
```diff
@@ -0,0 +1,9791 @@
+/**
```
arcuri82: this is nearly 10 000 lines of code, which was automatically generated. fine.
does Thrift require a special tool to generate this code? or could it be called via a Maven plugin?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -340,6 +340,10 @@ class EMConfig {
                     "The option has been deactivated for this run, to prevent a crash.")
             //throw IllegalArgumentException("The option to turn on Executive Summary is only meaningful when clustering is turned on (--testSuiteSplitType CLUSTERING).")
         }
+
+        if (enablePureRPCTestGeneration && outputFormat != OutputFormat.DEFAULT && !outputFormat.isJava()){
```
arcuri82: what is "pure" in this contex?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1537,6 +1542,26 @@ class EMConfig {
...
+    @Experimental
+    @Cfg("Whether to enable extra targets for responses, e.g., regarding nullable response, having extra targets for whether it is null")
+    var enableRPCExtraResponseTargets = false
+
+    @Experimental
+    @Cfg("Whether to enable customized responses indicating business logic")
+    var enableRPCCustomizedResponseTargets = false
+
+    @Experimental
+    @Cfg("Whether to generate RPC tests which is independent from EM driver.")
```
arcuri82: what does this mean? no WBT test case can be independent from EMDriver, eg due to start/stop/reset

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1537,6 +1542,26 @@ class EMConfig {
...
+    var enablePureRPCTestGeneration = false
+
+    @Experimental
+    @Cfg("Whether to generate RPC Assertions based on instance")
+    var enableRPCAssertionWithInstance = false
```
arcuri82: what does it mean "based on instance"?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1537,6 +1542,26 @@ class EMConfig {
...
+
+    @Experimental
+    @Cfg("Specify a maximum number of data in a collection to be asserted in generated tests." +
+            "Note that a zero or negative value means all data in the collection should be asserted.")
+    var maxAssertionForDataInCollection = -1
```
arcuri82: shouldn't `0` rather mean that no data is going to be asserted?

arcuri82 (changes requested):

arcuri82:
Hi. thx! i did't manage to double-check carefully every single file... :( let me know if I missed any important part you wanted my feedback on

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/ActionResponseDto.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     *  an original response with json format
+     */
+    public String jsonResponse;
+
+    /**
+     * code for processing the action
+     */
+    public List<String> testScript;
```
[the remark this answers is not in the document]
ZhangMan1126: as we discussed, in generated test scripts, the action could be involved by either 1) `executeRPCAction` (a method in SutHandler) or 2) `client.endpoint(args)`. For 1), we could process it from `core` side, while for 2), now it is handled from `driver side`. then we need such attributes in `DTO` to send generate scripts from `driver` to `core`. I will add such clarification in code as well.

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
ZhangMan1126: it will be used in the generated tests if the invocation of RPC endpoints in tests is proceeded with `executeRPCAction` method in driver.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/CustomizedRequestValueDto.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * a set of fields and corresponding values in request
+     */
+    public List<KeyValuePairDto> combinedKeyValuePairs;
+
+    /**
+     * a key with a set of candidate values
+     */
+    public keyValuesDto keyValues;
```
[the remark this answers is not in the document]
ZhangMan1126: `combinedKeyValuePairs` represents a candidate for a set of dependent keys and `keyValues` represents a set of candidates for one key. the clarification is added in the code as well

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/CustomizedCallResultCode.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * service error
+     */
+    SERVICE_ERROR,
+
+    /**
+     * other error scenarios, eg user error
+     */
+    OTHERWISE_ERROR
```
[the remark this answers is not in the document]
ZhangMan1126: updated

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/JsonAuthRPCEndpointDto.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * a list of the payload to send, as stringified JSON object
+     */
+    public List<String> jsonPayloads;
+
+    /**
+     * a list of the class name of the json object
+     */
+    public List<String> classNames;
```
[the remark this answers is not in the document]
ZhangMan1126: add clarification as below:
'jsonPayloads' are used to specify values of params for RPC endpoints
here are needed to explicitly specify what param types are with their full class name

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/keyValuesDto.java
```diff
@@ -0,0 +1,23 @@
+package org.evomaster.client.java.controller.api.dto;
+
+
+import java.util.List;
+
+/**
+ * dto is used to specify multiple seeds for a field.
+ * the field could be regarded as independent [CustomizedRequestValueDto#keyValues]
+ */
+public class keyValuesDto {
```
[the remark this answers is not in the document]
ZhangMan1126: updated

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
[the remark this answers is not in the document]
ZhangMan1126: each interface could have a corresponding type for its client,  (it might not always be true, as T extends X or X extends T, see native thrift client). Also in the driver, the RPC problem is specified with `Map<String, Object> mapOfInterfaceAndClient` for interfaces to client instances. then here we return `Object`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/EMController.java
ZhangMan1126: thanks. updated

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
ZhangMan1126: clarification is added as below in code:
 * - key is the index of the auth info which is specified in the driver
 * - value is extracted local auth setup schema

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/CodeJavaGenerator.java
ZhangMan1126: this class is not related to endpoint invocation
it is only used to generate test scripts and assertion scripts in generated tests.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: if we employ `driver`, ie. `executeRPCAction` to invoke endpoints in generated tests, or we generate it as `client.endpoint(args instance)`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: update it to `Whether to generate RPC endpoint invocation which is independent from EM driver.`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: update it to `Whether to generate RPC Assertions based on response instance`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
ZhangMan1126: it could be, then it means that only `size` of collection would be asserted.

ZhangMan1126:
Hi, just summarize some changes on genes
`ArrayGene` and `MapGene` are updated with `minSize`, `MapGene` supports enum type, and new gene `SeededGene`
Since now we support javax.validation. then based on the example in industrial case study, I found that a string might be specified with `Max` or `Min`. As checked with industrial partner, it is for value, not for length of the string, then regarding this case, I add a `LongGene` into `specializationGenes` of the StringGene (see `RPCEndpointsHandler`).

for specifying auth info, besides RPC endpoint with json inputs and customized value in fields, we now also support to implement `handleLocalAuthenticationSetup` in`SutHandler`. this is because for industrial case study, only such local handling could enable auth setup, I will bring you more info if you want. however, are you ok with additional method in `SutHandler`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
ZhangMan1126: updated

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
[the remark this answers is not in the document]
ZhangMan1126: we could pass a `Class<?>` instead of `String`, before I thought that there might have dependency or class loader problem. is it ok with `string`, then use `Class.forName` to get class with string?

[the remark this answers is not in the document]
arcuri82: hmm... i see. i guess it is ok for now

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
arcuri82: in that case, shouldn't `executeRPCAction` chech if  `extractRPCSchema` has been called, and, if not, call it?

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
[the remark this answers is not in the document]
arcuri82: i guess this will become more clear when we are going to generate tests using the interfaces directly

client-java/controller/src/test/java/com/thrift/example/real/facebook/fb303/FacebookService.java
arcuri82: ok, thx

arcuri82:
additional methods in `SutHandler` are fine, as long as they are needed in the generated tests.

disabling `TestabilityEMTest` and `DbBaseEMTest` is bit worrisome. we need to understand what is going on.
Also, an old related task: https://trello.com/c/dJ7X7w1Y/192-update-seems-showing-further-sql-issue-need-to-study-side-effects-of-tt-on-dbbaseemtest

client-java/controller/src/main/java/org/evomaster/client/java/controller/SutHandler.java
ZhangMan1126: we could do that. `executeRPCAction` is also used during the search. during the search, `extractRPCSchema` would be invoked by `getSutInfo`, also inside `extractRPCSchema`, there is a check if the schema is extracted. if we move `extractRPCSchema` (totally doable), such check would be invoked every time during the search. 

currently, we keep the test generation with driver because we need it to process e2e tests since the other solution (test generated on the driver) only supports Java not kotlin.
if further we support kotlin, `extractRPCSchema` could be removed from `SutHandler`.

ZhangMan1126:
the two tests pass on local. failing message on CI could be found here https://github.com/EMResearch/EvoMaster/runs/4725265779?check_suite_focus=true

For DbBaseEM, one of tests fail.
For Testability, it needs a further check. 
I will create a Trello card for this.

PR #436, opened 2022-01-17
handle failing sql heuristics for select 1

ZhangMan1126:


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/SqlHandler.java
```diff
@@ -149,7 +149,11 @@ public ExecutionDto getExecutionDto() {
 
         buffer.stream()
                 .forEach(sql -> {
-                    if (isSelect(sql) || isDelete(sql) || isUpdate(sql)) {
+                    /*
+                        for `SELECT 1`, the dist is always 0, might skip to compute heuristics for it
+                        check with Andrea
+                     */
+                    if (!isSelectOne(sql) && (isSelect(sql) || isDelete(sql) || isUpdate(sql))) {
```
arcuri82, line 156: hi. should be fine. i guess any operation with no `WHERE` should be skipped. but i guess `SELECT 1` might be the only one

client-java/controller/src/test/java/org/evomaster/client/java/controller/internal/db/ParserUtilsTest.java
```diff
@@ -42,4 +47,17 @@ public void testEmptyInsertValue(){
 
         assertTrue(parsed);
     }
+
+    @Test
+    public void testSelectOne() throws JSQLParserException {
+        String[] sqls = {"SELECT 1", "SELECT 1;", "Select    1;", "Select 1 ; "};
+        for (String sql: sqls){
```
arcuri82, line 54: this could be refactored with a parameterized test. see use of `@ParameterizedTest`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/SqlHandler.java
arcuri82: could add this clarification as a comment

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/SqlHandler.java
ZhangMan1126: With the latest commit, I add a comment regarding the method ie, `getDistances`. is it correct?

client-java/controller/src/test/java/org/evomaster/client/java/controller/internal/db/ParserUtilsTest.java
ZhangMan1126: thanks. done

PR #442, opened 2022-01-19
updating exp.py script to support .Net

arcuri82:


ZhangMan1126 (approved):
thanks! I do not have any further comments. I could run exp with generated scripts.

PR #444, opened 2022-01-21
multidimensional postgres arrays

jgaleotti:


core/src/test/resources/sql_schema/postgres_array_types.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE ArrayTypes (
+                            nonArrayColumn integer NOT NULL,
+                            arrayColumn integer[] NOT NULL,
+                            matrixColumn integer[][] NOT NULL,
+                            spaceColumn integer[][][] NOT NULL,
+                            manyDimensionsColumn integer[][][][] NOT NULL,
+                            exactSizeArrayColumn integer[3] NOT NULL,
+                            exactSizeMatrixColumn integer[3][3] NOT NULL
```
arcuri82, line 8: would be good to have at least one case of something different from integer, like a text array

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlMultidimensionalArrayGene.kt
```diff
@@ -0,0 +1,348 @@
...
+import org.evomaster.core.search.service.Randomness
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+/**
+ * https://www.postgresql.org/docs/14/arrays.html
+ * A multidimensional array representation.
+ * When created, all dimensions have length 0
+ */
+class SqlMultidimensionalArrayGene<T>(
```
arcuri82, line 18: @ZhangMan1126  can you look at this class?
when we add new genes (which might happen again in the future) we need to make sure that we handle all of their features, including adaptive hypermutation. 
for example, does this implementation handle it? if not, what is missing? likely we would need to write some documentation and/or test for when we add new genes

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlMultidimensionalArrayGene.kt
ZhangMan1126: @arcuri82 it does not enable the adaptive hypermutation, we need to add corresponding impacts and override methods of `Gene`. I will write such documentation. where do we put the documentation? part of README (eg for developers) or comments in `Gene`?

arcuri82: thx. let's have it in `Gene`. then, @jgaleotti can try to follow such documentation, and see if he can implement this gene correctly

ZhangMan1126: the documentation is added in the 32ecdc5 commit. let me know if anything is unclear @jgaleotti

core/src/test/resources/sql_schema/postgres_array_types.sql
jgaleotti: there is another test case using string arrays. The table is StringArrayTable.

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlMultidimensionalArrayGene.kt
jgaleotti: ok. I will have a look at it.

arcuri82:
@jgaleotti thx
@ZhangMan1126 can u look at the changes regarding hypermutation, if they are ok? thx

ZhangMan1126:
yes. I will check it. @arcuri82

core/src/main/kotlin/org/evomaster/core/search/impact/impactinfocollection/ImpactUtils.kt
```diff
@@ -69,6 +69,7 @@ class ImpactUtils {
                 is SqlPrimaryKeyGene -> SqlPrimaryKeyGeneImpact(id, gene)
                 is SqlForeignKeyGene -> SqlForeignKeyGeneImpact(id)
                 is SqlAutoIncrementGene -> GeneImpact(id)
+                is SqlMultidimensionalArrayGene<*> -> ArrayGeneImpact(id)
```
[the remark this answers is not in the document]
jgaleotti: Thanks. I added the new SqlMultidimensionalArrayGeneImpact instead of using ArrayGeneImpact.

arcuri82:
hi @ZhangMan1126 when u have time, please look at @jgaleotti changes. if you approve them, then write a comment here, and then i ll take over with reviewing this PR

ZhangMan1126 (commented):
@jgaleotti thanks.
@arcuri82 , I do not have further comments regarding adaptive mutation.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/ColumnDto.java
```diff
@@ -23,10 +23,16 @@ public class ColumnDto {
     public boolean isUnsigned = false;
 
     public boolean isEnumeratedType = false;
+
+    public boolean isCompositeType = false;
```
arcuri82, line 27: can add some comments on what a composite type is

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/CompositeTypeColumnDto.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * The name of the composite type
+     */
+    public String name;
+
+    /**
+     * Returns if the column is a composite or a built-in type
+     */
+    public boolean isCompositeType;
```
arcuri82, line 16: bit confusing... so a `CompositeTypeColumnDto` might have `isCompositeType` false???

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/CompositeTypeColumnDto.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * The name of the composite type
+     */
+    public String type;
+
+    public int size;
+
+    public boolean nullable;
+
+    public boolean isUnsigned = false;
```
arcuri82, line 27: why this redundant `= false`?  and, even if having it, shouldn't as well `nullable` have it to make it consistent?

arcuri82 (changes requested):
thx! just have a couple of minor comments

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/CompositeTypeColumnDto.java
jgaleotti: you might have something like this
CREATE TYPE myDomain AS (
  integerColumn integer NOT NULL
);
In this case, "integerColumn" is indeed a composite column but "integer" is not a composite type.

In contrast,
CREATE TYPE nestedType AS (
  myDomainColumn MyDomain NOT NULL
);
In this case, "myDomainColumn" is a composite type column but type "MyDomain", therefore isCompositeType==true. 

I agree the field name is confusing, instead of "isCompositeType" we could use "columnHasNoBuiltInType" ?

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/CompositeTypeColumnDto.java
jgaleotti: I will remove the =false to keep consistence.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/CompositeTypeColumnDto.java
arcuri82: yep, i guess that could be an option. can also add that example as a comment in the code to clarify it

jgaleotti: done!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/schema/ColumnDto.java
jgaleotti: I added an extract of the postgres documentation

PR #460, opened 2022-02-08
Handle customize constraints

ZhangMan1126:
allow user to specify customized not null constraints specified by annotations
fix some problems in test generations 
extract cause of undeclared throwable exception if the cause is not null
update a solution of handling field of dto specified with old or latest version of native thrift
handle dto which has superclass and generic type
smart db clean by removing accessed table


ZhangMan1126:
@arcuri82 the branch could be merged later, but it might be better to review it first before the RPC experiment starts.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/rpc/RPCExceptionInfoDto.java
```diff
@@ -26,4 +26,9 @@
      * a dto of the exception defined by the RPC service
      */
     public ParamDto exceptionDto;
+
+    /**
+     * if the exception is a cause of java.lang.UndeclaredThrowableException
+     */
```
arcuri82, line 32: why is this special?

should then rather provide directly the cause name?

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
```diff
@@ -357,6 +360,29 @@ public static void execScript(Connection conn, String script) throws SQLExceptio
         }
     }
 
+    public static List<String> extractSql(String script){
```
arcuri82, line 363: add some JavaDoc

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
```diff
@@ -357,6 +360,29 @@ public static void execScript(Connection conn, String script) throws SQLExceptio
...
+        String[] commands = script.split(";");
+        return Arrays.stream(commands).filter(s-> !s.replaceAll("\n","").isEmpty()).map(s-> s+";").collect(Collectors.toList());
+    }
+
+    /**
+     *
+     * @param commands
+     * @return
+     */
+    public static  Map<String, Set<String>> extractSqlTableMap(List<String> commands){
```
arcuri82, line 373: add more JavaDoc. eg, what it needs in input, and what does it return (what is the key in the map? what are the values in the Set? etc)

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -70,6 +74,23 @@
...
+     * track all tables accessed in a test
+     */
+    private final List<String> accessedTables = new CopyOnWriteArrayList<>();
+
+
+    /**
+     * a map of table to fk target tables
+     */
+    private final Map<String, List<String>> fkMap = new LinkedHashMap<>();
+
```
arcuri82, line 87: should use thread-safe maps here

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -70,6 +74,23 @@
...
+
+    /**
+     * a map of table to a set of commands which are to insert data into the db
+     */
+    private final Map<String, Set<String>> tableInitSqlMap = new LinkedHashMap<>();
+
```
arcuri82, line 93: see previous comment

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -270,36 +296,155 @@ public final ExtraHeuristicsDto computeExtraHeuristics() {
...
+    private Optional<String> findInCollectionIgnoreCase(String name, Collection<String> list){
+        return list.stream().filter(i-> i.equalsIgnoreCase(name)).findFirst();
+    }
+
+    private Optional<? extends Map.Entry<String, ?>> findInMapIgnoreCase(String name, Map<String, ?> list){
+        return list.entrySet().stream().filter(x-> x.getKey().equalsIgnoreCase(name)).findFirst();
+    }
+
+    public void cleanAccessedTables(){
```
arcuri82, line 339: if public method, add JavaDoc to it. also, should be on top, and not after other private methods

arcuri82: also, I guess should be marked with `final`

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/CodeJavaGenerator.java
```diff
@@ -21,6 +21,12 @@
      */
     private final static String GET_CLIENT_METHOD = "getRPCClient";
 
+
+    public static String handleClasNameWithGeneric(String fullName, List<String> genericTypes){
```
arcuri82, line 25: `Class`? (ie, missing a `s`)

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -551,12 +551,16 @@ class TestSuiteWriter {
 
 
         lines.block {
+
             if (format.isJavaScript()) {
+                //TODO add resetDatabase
                 addStatement("await $controller.resetStateOfSUT()", lines)
             } else if (format.isJavaOrKotlin()) {
+                addStatement("$controller.resetDatabase()", lines)
```
arcuri82, line 559: so, are we always adding `resetDatabase()`? shouldn't that depend on whether we handle or not a smart reset? (ie, if not, it should not be there). or is this handled inside `resetDatabase()`? clarify

arcuri82: but, even in this latter case, it would be confusing to see a call that does nothing in the generated tests

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/rpc/RPCExceptionInfoDto.java
ZhangMan1126: we handled the detailed info about the cause with this DTO, 
and this property is used to indicate if the exception is UndeclaredThrowableException and the DTO is for its cause.

the doc is updated.

PR #462, opened 2022-02-09
add filters for configs and suts

ZhangMan1126:
update `exp.py`


scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+
+
+EXP_ID = "evomaster"
+
+
+
+if len(sys.argv) < 9 or len(sys.argv) > 11:
+    print(
+        "Usage:\n<nameOfScript>.py <cluster> <baseSeed> <dir> <minSeed> <maxSeed> <maxActions> <minutesPerRun> <nJobs> <configFilter?> <sutFilter? eg a,b>")
+    exit(1)
```
arcuri82, line 92: can remove the example `eg a,b`

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+# that are not finished withing the specified time.
+MINUTES_PER_RUN = int(sys.argv[7])
+
+# How many scripts M we want the N jobs to be divided into.
+# Note: on cluster we can at most submit 400 scripts.
+# Also not that in the same .sh script there can be experiments only for a single SUT.
+NJOBS = int(sys.argv[8])
+
+# Specify a string to filter CONFIGS to be included
+# None or `all` represents all CONFIGS should be included
```
arcuri82, line 131: `an optional string`

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+# Default is None
+CONFIGFILTER = None
+if len(sys.argv) > 9:
+    CONFIGFILTER = str(sys.argv[9])
+
+# Specify a string to filter suts to be included based on their names
+# None or `all` represents all Suts should be included
+# Default is None
```
arcuri82, line 139: specify here as a comment that this can be a `,` separated list of names, like `a,b`

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+NJOBS = int(sys.argv[8])
+
+# Specify a string to filter CONFIGS to be included
+# None or `all` represents all CONFIGS should be included
+# Default is None
+CONFIGFILTER = None
+if len(sys.argv) > 9:
+    CONFIGFILTER = str(sys.argv[9])
+
+# Specify a string to filter suts to be included based on their names
```
arcuri82, line 137: `an optional string`

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+        # .Net
+        Sut("cs-rest-ncs",1,DOTNET_3),
+        Sut("cs-rest-scs",1,DOTNET_3),
+        Sut("sampleproject",1,DOTNET_3),
+        Sut("menu-api",1,DOTNET_3)
+]
+
+if SUTFILTER is not None and SUTFILTER.lower() != "all":
+    filteredsut = list(filter(lambda x: x.name.lower() in (f.lower() for f in SUTFILTER.split(",")), SUTS))
+    if len(filteredsut) > 0:
```
arcuri82, line 209: should give an error if user put a SUT name that does not exist

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+
+
+def createJobs():
+
+    CONFIGS = getConfigs()
+
+    ## filter config if specified
+    if CONFIGFILTER is not None and CONFIGFILTER.lower() != "all":
+            filtered = list(filter(lambda x: x.filterKey is None or x.filterKey.lower() == CONFIGFILTER.lower(), CONFIGS))
+            if len(filtered) > 0:
```
arcuri82, line 578: should give error if asking for filter tag that is undefined

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+    print("Total budget: " + str(CPUS * sum(state.waits) / 60) + " hours")
+
+
+
+############################################################################
+### Custom
+### Following will need to be changed based on what kind of experiments
+### we want to run.
+############################################################################
+
```
arcuri82, line 655: based on the comment above, are the following (eg `ParameterSetting`) something the user should change?
or rather the comment `Custom...` should be moved down this file?

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+class ParameterSetting:
+    def __init__(self, name, values):
+        self.name = name
+        self.values = values
+        self.count = len(self.values)
+
+    def pvalue(self, index):
```
arcuri82, line 662: add comment specifying what a `pvalue` is

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+
+
+############################################################################
+### Custom
+### Following will need to be changed based on what kind of experiments
+### we want to run.
+############################################################################
+
+class ParameterSetting:
+    def __init__(self, name, values):
```
arcuri82, line 657: add comment stating this `name` is the same name used in the EM parameters

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+
+class Config:
+    def __init__(self, settings, filterKey=None):
+        self.filterKey = filterKey
+        self.settings = settings
+        self.numOfSettings = 1
+        for s in self.settings:
+            self.numOfSettings *= s.count
+
+    def genAllSettings(self):
```
arcuri82, line 675: what is `gen` here? add comments (and/or more verbose name)

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+            return lst
+        for i in range(len(lst)):
+            if lst[i] < self.settings[i].count-1:
+                lst[i] = lst[i] + 1
+                return lst
+            else:
+                lst[i] = 0
+        return None
+
+    def setting(self, index):
```
arcuri82, line 699: what is this function for? add comment

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+        if index >= self.count:
+            exit("a setting at the index "+ str(index) + " does not exist")
+        return self.values[index]
+
+
+### NOTE there is no need to update this method
```
arcuri82, line 705: in this case, remove this comment, and move the function before the `# Custom` comment that starts the are the user is supposed to modify manually

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+        self.values = values
+        self.count = len(self.values)
+
+    def pvalue(self, index):
+        if index >= len(self.values):
+            exit("a value at the index "+ str(index) + " does not exist")
+        return (self.name, self.values[index])
+
+class Config:
+    def __init__(self, settings, filterKey=None):
```
arcuri82, line 668: specify that `settings` is an array of `ParameterSetting` objects

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+    try:
+        float(input)
+    except ValueError:
+        return False
+    return True
+
+def getConfigs():
+
+    # array of configuration objects. We will run experiments for each of
+    # these configurations
```
arcuri82, line 735: add comment to clarify that each Config object will have a list of `ParameterSetting` objects

scripts/expX2.py
```diff
@@ -0,0 +1,763 @@
...
+   ALGO_MIO = ParameterSetting("algorithm",["MIO"])
+   ALGO_RANDOM = ParameterSetting("algorithm",["RANDOM"])
+
+   PR5 = ParameterSetting("probOfRandomSampling",[0.5])
+   PR = ParameterSetting("probOfRandomSampling",[0.1, 0.2])
+
+   foo = Config([ALGO_MIO, PR5], "foo")
+   bar = Config([ALGO_RANDOM, PR], "bar")
+   CONFIGS.append(foo)
+   CONFIGS.append(bar)
```
arcuri82, line 748: this example can be left (it is good!), but comment it out

arcuri82 (changes requested):

scripts/expX2.py
arcuri82: rather than issuing error if none exists, should issue error if any does not exist.

arcuri82 (changes requested):

scripts/expX2.py
```diff
@@ -207,11 +207,12 @@ def __init__(self, name, timeWeight, platform):
...
 if SUTFILTER is not None and SUTFILTER.lower() != "all":
-    filteredsut = list(filter(lambda x: x.name.lower() in (f.lower() for f in SUTFILTER.split(",")), SUTS))
-    if len(filteredsut) > 0:
+    specifiedsut = SUTFILTER.split(",")
+    filteredsut = list(filter(lambda x: x.name.lower() in (f.lower() for f in specifiedsut), SUTS))
+    if len(filteredsut) > 0 and len(filteredsut) == len(specifiedsut):
         SUTS = filteredsut
     else:
-        print("ERROR: None of specified SUT exists")
+        print("ERROR: cannot find all specified SUTs. Please check them again.")
```
arcuri82, line 215: would be good to print out the actual SUTs that are misspelled/missing

arcuri82 (changes requested):

arcuri82 (approved):

PR #465, opened 2022-02-17
Wiremock

seran:


e2e-tests/wiremock/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+    <parent>
+        <artifactId>evomaster-e2e-tests</artifactId>
+        <groupId>org.evomaster</groupId>
+        <version>1.3.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>wiremock</artifactId>
+
+    <properties>
```
ZhangMan1126, line 14: is necessary to use JDK 11 for this module? if not, might remove this

e2e-tests/wiremock/src/main/java/com/foo/wiremock/WireMockApplication.java
```diff
@@ -0,0 +1,14 @@
+package com.foo.wiremock;
+
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+
+public class WireMockApplication {
```
ZhangMan1126, line 9: might create an E2E that contains two services, ie, `X` and `Y`. then one endpoint of `X` is to make a call to an endpoint of `Y`.

e2e-tests/wiremock/src/main/test/com/foo/wiremock/WireMockBasicTest.java
```diff
@@ -0,0 +1,53 @@
...
+                        aResponse()
+                        .withHeader("Content-Type", "text/plain")
+                        .withStatus(200)
+                        .withBody("Working")));
+    }
+
+    @Test
+    public void testServiceStatus() {
+
+        given()
```
ZhangMan1126, line 45: here, instead of directly invoking the mocked server, might call the service `X`

e2e-tests/wiremock/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+
+    <dependencies>
```
ZhangMan1126, line 19: could configure these dependencies in the `dependencyManagement` of the root `pom.xml`, then here do not need to refer to the version info

ZhangMan1126 (changes requested):

e2e-tests/wiremock/src/test/java/org/evomaster/wiremock/WireMockBasicTest.java
```diff
@@ -0,0 +1,58 @@
...
+                        aResponse()
+                                .withHeader("Content-Type", "text/plain")
+                                .withStatus(200)
+                                .withBody("Working")));
+    }
+
+    @Test
+    public void testServiceStatus() {
+
+        given()
```
ZhangMan1126, line 44: could you add a bit of description about this test?

e2e-tests/wiremock/src/test/java/org/evomaster/wiremock/WireMockBasicTest.java
```diff
@@ -0,0 +1,58 @@
...
+                .get("http://localhost:" + wireMockServer.port() +"/api/call")
+                .then()
+                .assertThat()
+                .statusCode(200)
+                .body(is("Working"));
+    }
+
+    @Test
+    public void testApplicationCall() {
+        WireMockApplication wireMockApplication = new WireMockApplication();
```
ZhangMan1126, line 55: in this test, there is no assertion. is it correct?
In addition, could you add a bit of description about this test?

e2e-tests/wiremock/src/main/java/com/foo/wiremock/WireMockApplication.java
```diff
@@ -0,0 +1,18 @@
+package com.foo.wiremock;
+
+
+import static org.hamcrest.Matchers.is;
+import static io.restassured.RestAssured.given;
+
+public class WireMockApplication {
+
+    public void callApi(int port) {
```
ZhangMan1126, line 9: add a bit description about this endpoint, eg, call another service Y which could be mocked in the manual test.
In addition, instead of asserting the call, should the endpoint return some info (eg, `Working`) which is received from the call?

ZhangMan1126 (changes requested):

e2e-tests/pom.xml
```diff
@@ -21,6 +21,7 @@
         <module>spring-graphql</module>
         <module>spring-rest-mysql</module>
         <module>spring-rpc</module>
+        <module>wiremock</module>
```
arcuri82, line 24: no need to create its own module. should be just a new package under `spring-rest-openapi-v2` (if writing in Java) or `v3` (if in Kotlin). see for example `e2e-tests/spring-rest-openapi -v2/src/test/java/org/evomaster/e2etests/spring/examples/db`

e2e-tests/wiremock/pom.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>evomaster-e2e-tests</artifactId>
+        <groupId>org.evomaster</groupId>
+        <version>1.3.1-SNAPSHOT</version>
```
arcuri82, line 8: update to latest version

e2e-tests/wiremock/src/main/java/com/foo/wiremock/WireMockApplication.java
```diff
@@ -0,0 +1,15 @@
+package com.foo.wiremock;
+
+
+import static org.hamcrest.Matchers.is;
+import static io.restassured.RestAssured.given;
+
+public class WireMockApplication {
+
+    public String callApi(int port) {
```
arcuri82, line 9: this is just a function.
should have an actual REST API, in which this fucntion is behind a GET endpoint for example

e2e-tests/wiremock/src/main/java/com/foo/wiremock/WireMockApplication.java
```diff
@@ -0,0 +1,15 @@
...
+        return given()
+                .when()
+                .get("http://localhost:" + port +"/api/call")
+                .body().asString();
+    }
```
arcuri82, line 14: should return something simple, like a string X.
then, here, should have a trivial check like `if(x.equals("42"))`, and return `OK` if so

pom.xml
```diff
@@ -698,6 +699,12 @@
                 <version>${thrift.libthrift.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>com.github.tomakehurst</groupId>
+                <artifactId>wiremock-jre8</artifactId>
+                <version>${wiremock.version}</version>
```
arcuri82, line 705: should have the `test` scope, and moved close to the other test library declarations

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/wiremock/WireMockManualTest.java
```diff
@@ -0,0 +1,41 @@
...
+                .get(baseUrlOfSut + "/api/wiremock/equalsFoo/foo")
+                .then()
+                .statusCode(200)
+                .body("valid", is(true));
+
+    }
+
+    @Test
+    public void testExternalCall() {
+        given().accept(ContentType.JSON)
```
ZhangMan1126, line 30: add a bit description about the `false` valid in body.

e2e-tests/spring-rest-openapi-v2/src/main/java/com/foo/rest/examples/spring/wiremock/WireMockRest.java
```diff
@@ -0,0 +1,60 @@
...
+            produces = MediaType.APPLICATION_JSON
+    )
+    public StringsResponseDto externalCall(
+            @PathVariable("key") String key
+    ) {
+        StringsResponseDto stringsResponseDto = new StringsResponseDto();
+
+        RestTemplate restTemplate = new RestTemplate();
+
+        //
```
ZhangMan1126, line 48: remove `\\` or add a bit of description about the behavior of the external services

ZhangMan1126 (commented):
thanks, Seran! I just have minor comments, after those are addressed, you could request Andrea's review.

arcuri82:
hi.
 @seran when u address Man's comments, also make sure to fix conflicts in `pom.xml`

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/wiremock/WireMockTestBase.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+
+public class WireMockTestBase extends SpringTestBase {
+
+    private static WireMockServer wireMockServer;
+
+    @BeforeAll
+    public static void initClass() throws Exception {
+        // For the moment port is set to 10101
```
arcuri82, line 19: do not use hardcoded port numbers, but rather ephemeral ones

e2e-tests/spring-rest-openapi-v2/src/main/java/com/foo/rest/examples/spring/wiremock/WireMockRest.java
```diff
@@ -0,0 +1,63 @@
...
+        StringsResponseDto stringsResponseDto = new StringsResponseDto();
+
+        RestTemplate restTemplate = new RestTemplate();
+
+        /**
+         * The URL below is hard coded with the port of Wiremock.
+         * When the key is alpha characters echo endpoint will
+         * respond with the given value, otherwise Not Found!
+        */
+        String uri = "http://localhost:52768/api/echo/" + key;
```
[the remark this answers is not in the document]
arcuri82: eg, once we start WireMock on ephemeral port, need to see if can replace hardcoded port to point to such ephemeral one

e2e-tests/spring-rest-openapi-v2/src/main/java/com/foo/rest/examples/spring/wiremock/WireMockRest.java
```diff
@@ -0,0 +1,63 @@
...
+
+        ResponseEntity<String> response = restTemplate.getForEntity(uri, String.class);
+
+        if (response.getBody().equals(key)) {
```
arcuri82, line 56: make  this basic example simpler. instead of having to bind input key to wiremock response, use something hardcoded. eg, `response.getBody().equals("foo")`

arcuri82 (approved):

PR #466, opened 2022-02-18
Fix smart dbreset

ZhangMan1126:
fix a problem on missing init data if the accessed table is empty
`resetDatabase` with specific tables

### DIFF
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
@@ -361,6 +361,22 @@ public class SqlScriptRunner {
 
     /**
      *
+     * @param conn a connection to db
+     * @param script represents a sql script
+     * @param tablesToInsert represents if only insert table which exists in this list
+     * @throws SQLException if the execution of the command fails
+     */
+    public static void execScript(Connection conn, String script, List<String> tablesToInsert) throws SQLException {
+        List<String> commands = extractSql(script);
+        for (String command : commands){
+            if(shouldExecuteInsert(command, tablesToInsert)){
+                execCommand(conn, command+";");
+            }
+        }
+    }
+
+    /**
+     *
      * @param script is a SQL script
      * @return a list of SQL commands based on the script
      */
@@ -422,6 +438,14 @@ public class SqlScriptRunner {
         statement.close();
 
         return queryResult;
+    }
+
+    private static boolean shouldExecuteInsert(String command, List<String> tablesToInsert){
+        if (!ParserUtils.isInsert(command)) return true;
+        if (tablesToInsert == null || tablesToInsert.isEmpty()) return false;
+        Insert stmt = (Insert) ParserUtils.asStatement(command);
+        Table table = stmt.getTable();
+        return table!= null && tablesToInsert.stream().anyMatch(t-> t.equalsIgnoreCase(table.getName()));
     }
 
 }
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
@@ -363,10 +363,13 @@ public class SqlScriptRunner {
      *
      * @param conn a connection to db
      * @param script represents a sql script
-     * @param tablesToInsert represents if only insert table which exists in this list
+     * @param tablesToInsert represents insertions are executed only on the tables in this list.
+     *                       insertions for other tables will be skipped.
+     *                       if it is null or empty, nothing will be inserted.
      * @throws SQLException if the execution of the command fails
      */
     public static void execScript(Connection conn, String script, List<String> tablesToInsert) throws SQLException {
+        if (tablesToInsert == null || tablesToInsert.isEmpty()) return;
         List<String> commands = extractSql(script);
         for (String command : commands){
             if(shouldExecuteInsert(command, tablesToInsert)){
--- a/core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
+++ b/core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
@@ -638,6 +638,25 @@ class SqlInsertBuilder(
         }
     }
 
+    /**
+     * extract tables with additional fk tables
+     * @param tables to check
+     * @param all is a complete set of tables with their fk
+     */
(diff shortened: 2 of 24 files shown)


client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
```diff
@@ -359,6 +359,22 @@ public static void execScript(Connection conn, String script) throws SQLExceptio
         }
     }
 
+    /**
+     *
+     * @param conn a connection to db
+     * @param script represents a sql script
+     * @param tablesToInsert represents if only insert table which exists in this list
```
arcuri82, line 366: i don't understand this comment... need to better clarify it

arcuri82: do u mean something like `insertions are executed only on the tables in this list. insertions for other tables will be skipped. `?

arcuri82: also, what should happen if `tablesToInsert` is null or empty? should throw an IllegalArgumentException?

core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
```diff
@@ -638,6 +638,26 @@ class SqlInsertBuilder(
         }
     }
 
+    /**
+     * extract tables with additional fk tables
+     * @param tables to check
+     * @param all is a complete set of tables with their fk
+     */
+    fun extractFkTable(tables: Set<String>, all: MutableSet<String>){
+        tables.forEach { t->
```
arcuri82, line 647: i guess here we should return the `all` set, as immutable `Set`. also, to deal with recursion, the input `all` could have a default `= mutableSet<String>()`

core/src/main/kotlin/org/evomaster/core/output/service/RPCTestCaseWriter.kt
```diff
@@ -72,7 +72,7 @@ class RPCTestCaseWriter : WebTestCaseWriter() {
             if (rpcCallAction.response!=null && varName != null){
                 if (rpcCallResult.hasResponse())
                     handleAssertions(lines, varName, rpcCallResult)
-                else{
+                else if(!rpcCallResult.isExceptionThrown()){
```
arcuri82, line 75: what shall happen if exection is thrown?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -131,6 +136,32 @@ class TestSuiteWriter {
         return lines.toString()
     }
 
+    private fun handleResetDatabaseInput(solution: Solution<*>): String{
+        if (!config.outputFormat.isJavaOrKotlin())
+            throw IllegalStateException("DO NOT SUPPORT resetDatabased for "+ config.outputFormat)
+
+        val accessedTable = mutableSetOf<String>()
+        solution.individuals.forEach { e->
+            if (e.individual is ApiWsIndividual){
```
arcuri82, line 145: add `//TODO will need to be refactored when supporting Web Frontend`

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsSampler.kt
```diff
@@ -81,4 +81,11 @@ abstract class ApiWsSampler<T> : Sampler<T>() where T : Individual {
 
     abstract fun initSqlInfo(infoDto: SutInfoDto)
 
+    override fun extractFkTables(tables: Set<String>): Set<String> {
+        if(sqlInsertBuilder == null || tables.isEmpty()) return tables
+        val allSet = mutableSetOf<String>()
+        sqlInsertBuilder!!.extractFkTable(tables, allSet)
```
arcuri82, line 87: once fixed the other comment, no need to create `allSet` as a variable here

core/src/main/kotlin/org/evomaster/core/search/service/Sampler.kt
```diff
@@ -107,4 +107,11 @@ abstract class Sampler<T> : TrackOperator where T : Individual {
             EMConfig.MaxTestSizeStrategy.DPC_DECREASING -> apc.getExploratoryValue(config.maxTestSize, config.dpcTargetTestSize)
         }
     }
+
+    /**
+     * extract tables with additional FK tables
+     */
+    open fun extractFkTables(tables: Set<String>): Set<String>{
+        return tables
```
arcuri82, line 115: by default, shouldn't rather throw an exception?

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/db/SqlScriptRunner.java
ZhangMan1126: thanks. doc is updated. 
if it is `null` or `empty`, nothing will be inserted now. This function is mainly used in `resetDatabase` in generated tests.

core/src/main/kotlin/org/evomaster/core/output/service/RPCTestCaseWriter.kt
ZhangMan1126: if an exception is thrown, we will provide the info about what the exception is and where the exception happens as before and put the invocation under the try-catch. here I exclude the assertion for the response because with ind1 case study, the response is quite uncertain.

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
ZhangMan1126: thanks, added

core/src/main/kotlin/org/evomaster/core/database/SqlInsertBuilder.kt
ZhangMan1126: updated

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsSampler.kt
ZhangMan1126: updated

core/src/main/kotlin/org/evomaster/core/search/service/Sampler.kt
ZhangMan1126: updated

PR #468, opened 2022-03-03
Fix ind1

ZhangMan1126:
fix problems in array and map mutation with specified constraints
add constraints handling for decimalMax and decimalMin

### DIFF
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/JavaXConstraintHandler.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/JavaXConstraintHandler.java
@@ -35,9 +35,9 @@ public class JavaXConstraintHandler {
             case SIZE: solved = handleSize(namedTypedValue, annotation); break;
             case PATTERN: solved = handlePattern(namedTypedValue, annotation); break;
             case DECIMAL_MAX:
-            case MAX: solved = handleMax(namedTypedValue, annotation); break;
+            case MAX: solved = handleMax(namedTypedValue, annotation, supportType); break;
             case DECIMAL_MIN:
-            case MIN: solved = handleMin(namedTypedValue, annotation); break;
+            case MIN: solved = handleMin(namedTypedValue, annotation, supportType); break;
         }
 
         if (!solved){
@@ -149,7 +149,7 @@ public class JavaXConstraintHandler {
         return true;
     }
 
-    private static boolean handleMax(NamedTypedValue namedTypedValue, Annotation annotation){
+    private static boolean handleMax(NamedTypedValue namedTypedValue, Annotation annotation, JavaXConstraintSupportType supportType){
 
          /*
             based on
@@ -158,9 +158,16 @@ public class JavaXConstraintHandler {
             null elements are considered valid.
          */
         Long max = null;
+        Boolean inclusive = true;
         try {
             // TODO might change long to BigDecimal
-            max = (Long) annotation.annotationType().getDeclaredMethod("value").invoke(annotation);
+            if (supportType == JavaXConstraintSupportType.DECIMAL_MAX){
+                String maxStr =  (String) annotation.annotationType().getDeclaredMethod("value").invoke(annotation);
+                max = Long.valueOf(maxStr);
+                inclusive = (Boolean) annotation.annotationType().getDeclaredMethod("inclusive").invoke(annotation);
+            }else
+                max = (Long) annotation.annotationType().getDeclaredMethod("value").invoke(annotation);
+
         } catch (NoSuchMethodException | InvocationTargetException |IllegalAccessException e) {
             throw new RuntimeException("ERROR: fail to process max");
         }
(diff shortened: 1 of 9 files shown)


client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/JavaXConstraintHandler.java
```diff
@@ -20,20 +20,24 @@
...
+        switch (supportType){
+            case NOT_NULL: solved = handleNotNull(namedTypedValue); break;
+            case NOT_EMPTY: solved = handleNotEmpty(namedTypedValue); break;
+            case NOT_BLANK: solved = handleNotBlank(namedTypedValue); break;
+            case SIZE: solved = handleSize(namedTypedValue, annotation); break;
+            case PATTERN: solved = handlePattern(namedTypedValue, annotation); break;
+            case DECIMAL_MAX:
+            case MAX: solved = handleMax(namedTypedValue, annotation, supportType); break;
+            case DECIMAL_MIN:
+            case MIN: solved = handleMin(namedTypedValue, annotation, supportType); break;
```
arcuri82, line 40: are these all supported cases?
should we have a `default` throwing an exception if reached?

arcuri82: or log it?

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/JavaXConstraintHandler.java
```diff
@@ -42,170 +46,186 @@ public static void handleParam(NamedTypedValue namedTypedValue, Annotation annot
...
+            }else
+                max = (Long) annotation.annotationType().getDeclaredMethod("value").invoke(annotation);
 
-            try {
-                return (String) annotation.annotationType().getDeclaredMethod("regexp").invoke(annotation);
-            } catch (NoSuchMethodException | InvocationTargetException |IllegalAccessException e) {
-                throw new RuntimeException("ERROR: fail to process regexp");
-            }
+        } catch (NoSuchMethodException | InvocationTargetException |IllegalAccessException e) {
+            throw new RuntimeException("ERROR: fail to process max");
```
arcuri82, line 172: when rethrowing an exception, add its output in the string message, or add it as a cause as last parameter in this constructor

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -96,7 +97,12 @@ class TestSuiteWriter {
 
         //catch any sorting problems (see NPE is SortingHelper on Trello)
         val tests = try {
-            testSuiteOrganizer.sortTests(solution, config.customNaming)
+            // NEED to check with Andrea, skip to sort RPC for the moment
```
arcuri82, line 100: add `TODO` on this comment

arcuri82: yes, in future we need to handle test suite split for all types, including RPC ang GQL

core/src/main/kotlin/org/evomaster/core/search/gene/ArrayGene.kt
```diff
@@ -50,13 +50,17 @@ class ArrayGene<T>(
...
 
         if (maxSize != null && elements.size > maxSize!!) {
-            throw IllegalArgumentException(
-                "ArrayGene "+name+": More elements (${elements.size}) than allowed ($maxSize)")
+//            throw IllegalArgumentException(
+//                "ArrayGene "+name+": More elements (${elements.size}) than allowed ($maxSize)")
+
+            log.warn("ArrayGene "+name+": More elements (${elements.size}) than allowed ($maxSize)")
+
         }
```
arcuri82, line 64: not sure about this...
if it happens it means it is a bug in EM, which we must fix.
if we rather log it instead of throwing an exception, we might just miss this issue.

core/src/main/kotlin/org/evomaster/core/search/gene/MapGene.kt
```diff
@@ -34,12 +34,16 @@ class MapGene<K, V>(
...
                 "MapGene "+name+": minSize (${minSize}) is greater than maxSize ($maxSize)")
         }
 
         if (maxSize != null && elements.size > maxSize!!) {
-            throw IllegalArgumentException(
+//            throw IllegalArgumentException(
+//                "MapGene "+name+": More elements (${elements.size}) than allowed ($maxSize)")
+            log.warn(
                 "MapGene "+name+": More elements (${elements.size}) than allowed ($maxSize)")
         }
```
arcuri82, line 48: same as previous comment

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/JavaXConstraintHandler.java
ZhangMan1126: it is handled at line 26 by checking whether we support the annotation with its simple name. 
To be safe, I will add a default to have a further check.

PR #478, opened 2022-03-19
fix a bug which leads to ocvn crash

ZhangMan1126:
the update was checked with EMB

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceManageService.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceManageService.kt
@@ -216,7 +216,9 @@ class ResourceManageService {
                     previousDbActions = bindWith?.flatMap { it.seeActions(ActionFilter.ONLY_SQL) as List<DbAction>} ?: listOf())
 
                 if(!created){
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/service/ResourceManageService.kt
```diff
@@ -216,7 +216,9 @@ class ResourceManageService {
                     previousDbActions = bindWith?.flatMap { it.seeActions(ActionFilter.ONLY_SQL) as List<DbAction>} ?: listOf())
 
                 if(!created){
-                    LoggingUtil.uniqueWarn(log, "it is unlikely to prepare a resource for $resourceKey with SQL/REST action")
+                    // a resource might be impossible to create with sql or endpoint, then use info instead of warning
+                    log.info("it is unlikely to prepare a resource for $resourceKey with SQL/REST action")
```
arcuri82, line 220: why impossible with SQL? due to limitations of EM? Need to discuss.
if this is expected, maybe better to use `debug` instead of `info`

arcuri82 (changes requested):

PR #480, opened 2022-03-22
Wiremock emb usecases

seran:


docs/loopback.md
```diff
@@ -0,0 +1,14 @@
+# Multiple loopback interface for tests
+
+WireMock tests require different loopback address other than 127.0.0.1.
+
+By default, in macOS, this address is not available to enable it run 
+the following command.
+
+`sudo ifconfig lo0 alias 127.0.0.2`
```
arcuri82, line 8: does this need to be done for every single 127.x.x.x address?
in that case, might want to create a script to do that (eg a loop over all the 256^3 addresses)

client-java/instrumentation/src/main/resources/keepInstrumentationList.txt
```diff
@@ -1,4 +1,3 @@
 //list of classes/packages to include regardless of skipList.
 //this is useful when you want to ban a domain org.foo but still allow
 //some of its subpackages, like org.foo.bar
-
```
arcuri82, line 4: any reason for modyfing this file?

arcuri82 (changes requested):

docs/loopback.md
arcuri82: note: can just leave a comment about it. no need to write the script for this PR

client-java/instrumentation/src/main/resources/keepInstrumentationList.txt
seran: Did some experiments, previous commit captured the changes. In this reverted it back to the original version.

docs/loopback.md
seran: I'll add the script in the next commit.

PR #485, opened 2022-04-01
WireMock EMB method replacement

seran:


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/URLClassReplacement.java
```diff
@@ -0,0 +1,46 @@
...
+    )
+    public static URLConnection openConnection(URL caller) throws java.io.IOException {
+        /**
+         * If the protocol is HTTP/S and the remote is not localhost, then the
+         * request will be remapped to WireMock instance running in a different
+         * loopback.
+         * */
+        URLConnection urlConnection = null;
+        if ((caller.getProtocol().equals("http") || caller.getProtocol().equals("https")) && !caller.getHost().equals("localhost")) {
+                String url = "http://127.0.0.2:8080" + caller.getPath() + ((caller.getQuery() != null) ? "?" + caller.getQuery() : "");
```
ZhangMan1126, line 36: if it is fixed, it might be better to declare it as a constant somewhere.

ZhangMan1126 (commented):
thanks @seran. I just have a minor comment.

@arcuri82 the URL method replacement works. do we need to include the part for sending the url info (such as) to core? if not, after the Seran address the minor comments, it is ready for you to review.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/URLClassReplacement.java
seran: This is a basic implementation, just to test the idea. Have to improve this implementation with the upcoming tasks. Will send the information to core using `additional info` with the next task.

arcuri82: @ZhangMan1126 @seran  Hi. features do not need to be fully completed  to be merged into master. as long as the build does not fail, we should merge ongoing work at least once every 2 weeks. so, `sending the url info (such as) to core` can be done in another branch after this

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/AdditionalInfo.java
```diff
@@ -65,6 +65,21 @@ public StatementDescription(String line, String method) {
...
     private final Map<String, Deque<StatementDescription>> lastExecutedStatementStacks = new ConcurrentHashMap<>();
 
+    /**
+     * To keep track of external service hosts called by the SUT.
+     * When captured mock hostname will be empty, till the WireMock instance
+     * initiated.
+     */
+    private final Set<ExternalServiceInfo> externalServices = new CopyOnWriteArraySet<>();
+
+    public void addExternalService(ExternalServiceInfo hostInfo) {
```
arcuri82, line 75: move these public setter/getter after the private fields (ie, be consistent with all the other getters/setters location)

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/AdditionalInfo.java
```diff
@@ -65,6 +65,21 @@ public StatementDescription(String line, String method) {
...
+        externalServices.add(hostInfo);
+    }
+
+    public Set<ExternalServiceInfo> getExternalServices() {
+        return externalServices;
```
arcuri82, line 80: should be in a `Collections.unmodifiableSet`. see other getters

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExternalServiceInfo.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Contains the remote hostname
+     */
+    private final String remoteHostname;
+
+    private final String protocol;
+
+    private final int remotePort;
+
+    private final String mockHostname;
```
arcuri82, line 19: what is this? add comment

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExternalServiceInfo.java
```diff
@@ -0,0 +1,51 @@
...
+public class ExternalServiceInfo implements Serializable {
+
+    /**
+     * Contains the remote hostname
+     */
+    private final String remoteHostname;
+
+    private final String protocol;
+
+    private final int remotePort;
```
arcuri82, line 17: what if the port is not specified? maybe this could be a nullable `Integer`?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExternalServiceInfo.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final String mockHostname;
+
+    private final int mockHostPort;
+
+    public ExternalServiceInfo(String protocol, String remoteHostname, int remotePort, String mockHostname, int mockHostPort) {
```
arcuri82, line 23: as u are using this class inside a `Set`, then must define `equals/hashcode` methods. IDE can generate those for you

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExternalServiceInfo.java
seran: Most case I've observed there is a port. If there is nothing we can use `-1`.

PR #499, opened 2022-04-27
Big integer decimal

ZhangMan1126:
handle min, max, minInclusive, maxInculsive, precision and scale for number genes
add BigIntegerGene
add BIgDecimalGene
add NumericStringGene
handle constraints, such as Digits, Positive(OrZero), Negative(orZeror), Null, AssertFase, AssertTue
fix some problems reported by ind1

add unit tests accordingly

update e2e tests for big numbers and handling negative floatingpoint number
add e2e tests for numericstring gene


core/src/main/kotlin/org/evomaster/core/search/gene/IntegralNumberGene.kt
```diff
@@ -0,0 +1,11 @@
+package org.evomaster.core.search.gene
+
+abstract class IntegralNumberGene<T: Number> (
```
arcuri82, line 3: what does `Integral` means here?

core/src/main/kotlin/org/evomaster/core/search/gene/NumericStringGene.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.evomaster.core.search.impact.impactinfocollection.value.OptionalGeneImpact
+import org.evomaster.core.search.impact.impactinfocollection.value.numeric.NumericStringGeneImpact
+import org.evomaster.core.search.service.AdaptiveParameterControl
+import org.evomaster.core.search.service.Randomness
+import org.evomaster.core.search.service.mutator.MutationWeightControl
+import org.evomaster.core.search.service.mutator.genemutation.AdditionalGeneMutationInfo
+import org.evomaster.core.search.service.mutator.genemutation.SubsetGeneSelectionStrategy
+import java.math.BigDecimal
+
+class NumericStringGene(
```
arcuri82, line 13: explain this gene. write something like this is a number in all aspects, but in its phenotype it is going to be outputted as a string

core/src/main/kotlin/org/evomaster/core/search/gene/NumericStringGene.kt
```diff
@@ -0,0 +1,114 @@
...
+    name: String,
+    /**
+     * inclusive
+     *
+     * note that precision would represent maxLength
+     */
+    val minLength: Int,
```
arcuri82, line 20: if this `minLength` constraints is needed, shouldn't it be inside `BigDecimalGene`? otherwise, how can we ensure a mutation on this latter will not break the constraint?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/NumericStringGene.kt
ZhangMan1126: For BigDecimalGene, there is no lower bound for the digits. 
I add a TODO for this String if it exists.
It is a bit tricky.
It depends on how the string would be further dealt with, eg, we could fill 0 at left. 
or we need to control min and max.

arcuri82 (changes requested):

PR #504, opened 2022-05-05
enable info collected during sut boot-time

ZhangMan1126:
manipluate executing action state
collect and restore objective info (such as targets and external service info) during the boot-time (only for JVM)
tests for checking whether the boot-time targets are covered with the exported targets file.


core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
```diff
@@ -146,6 +147,29 @@ class FitnessValue(
...
     }
 
+    /**
+     * this method is to report the union results with targets at boot-time
+     *
+     * @return first is boot-time (negative means that the boot-time info is unavailable)
+     *         second is search time, and
+     *         third is a union of boot-time and search time
+     */
+    fun unionWithBootTimeCoveredTargets(prefix: String?, idMapper: IdMapper, bootTimeInfoDto: BootTimeInfoDto?, unavailableBootTime: Int = -1): Triple<Int, Int, Int>{
```
arcuri82, line 157: the use of `Triple` is fine in a local scope, but not when returned by a public method, as it makes code harder to read. eg., how to understand what `first` and `third` are referring to semantically.
create a new class with meaningful names to replace this `Triple`

arcuri82, line 157: the use of `Triple` is fine in a local scope, but not when returned by a public method, as it makes code harder to read. eg., how to understand what `first` and `third` are referring to semantically.
create a new class with meaningful names to replace this `Triple`

e2e-tests/spring-rest-openapi-v3/src/main/kotlin/com/foo/rest/examples/spring/openapi/v3/boottimetargets/BootTimeTargetsRest.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val startupTime  = LocalDateTime.now()
+    private var startupInfo : String
+
+    init {
+        startupInfo = "${branchTarget()}:$startupTime"
+        println("startupInfo:$startupInfo")
+    }
+
+    companion object{
+        // TODO might need a further check, targets for lines 24 and 25 seems to be skipped
```
arcuri82, line 23: yep, indeed, we do not instrument class initializers

arcuri82, line 23: yep, indeed, we do not instrument class initializers

arcuri82 (changes requested):

seran (commented):
Thank you @ZhangMan1126. I have no comments.

core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
```diff
@@ -167,7 +166,7 @@ class FitnessValue(
                     duplicatedcounter++
             }
         }
-        return Triple(bootTime.size, searchTime, bootTime.size + searchTime - duplicatedcounter)
+        return TargetStatisticTriple(bootTime.size, searchTime, bootTime.size + searchTime - duplicatedcounter)
```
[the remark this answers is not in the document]
ZhangMan1126: maybe it only happens for class due to class load? 
to ensure that we avoid the repeated ones, we might still need to have a check on them

core/src/main/kotlin/org/evomaster/core/search/TargetStatisticTriple.kt
```diff
@@ -0,0 +1,21 @@
...
+     */
+    val bootTime: Int,
+    /**
+     * achieved by search
+     */
+    val searchTime: Int,
+    /**
+     * a total of unique targets covered at the end
+     */
+    val total: Int
```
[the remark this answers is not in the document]
arcuri82: @ZhangMan1126 hi. thx. wondering if this is related to handling of `firstTimeEncountered`, which might not have been reset, maybe?

arcuri82 (approved):

PR #513, opened 2022-05-13
WireMock initialisation for external services

seran:


core/src/test/kotlin/org/evomaster/core/problem/external/ExternalServiceUtilsTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Tests the next IP address generator. If IP address is out of range from
+     * loopback range should throw exception as well as if IP address is given
+     * in wrong format.
+     */
+    @Test
+    fun testNextAddress() {
+        assertEquals("127.1.100.2", nextIPAddress("127.1.100.1"))
+        assertNotEquals("127.0.0.1", nextIPAddress("127.0.0.0"))
```
ZhangMan1126, line 21: might add another assertion about what it should be

core/src/test/kotlin/org/evomaster/core/problem/external/ExternalServiceUtilsTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        assertThrows(Exception::class.java) {
+            nextIPAddress("127.255.255.254")
```
ZhangMan1126, line 24: is it due to being out of range? since we check if the ip is avaiable, we might allow the loop, eg, from the begining. how do you thint?

ZhangMan1126 (commented):
many thanks, Seran! just a few comments.

core/src/test/kotlin/org/evomaster/core/problem/external/ExternalServiceUtilsTest.kt
seran: Loopback address ranges from 127.0.0.0 to 127.255.255.255. Usually highest number of a range considered as broadcast address, to avoid unexpected conflicts this address is intentionally skipped. That's why this check is there.

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceUtils.kt
```diff
@@ -0,0 +1,95 @@
...
+                    for (j in i + 1 until tokens.size) {
+                        tokens[j] = "0"
+                    }
+                    break
+                }
+            }
+        }
+        var ip = String.format("%s.%s.%s.%s", tokens[0], tokens[1], tokens[2], tokens[3])
+        if (tokens[0].toInt() != 127) {
+            throw Exception("Next available IP address is out of usable range")
```
arcuri82, line 36: rather throw an IllegalStateException

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceInfo.kt
```diff
@@ -1,10 +1,15 @@
...
 
+import com.github.tomakehurst.wiremock.WireMockServer
+
 class ExternalServiceInfo(
         val protocol: String,
         val remoteHostname: String,
         val remotePort: Int) {
 
+    // TODO: Need to refactor
+    private lateinit var wireMockServer : WireMockServer
```
arcuri82, line 11: not sure should be done here in a class called `Info`. Maybe create a new class to deal with the actual storage and handling of WM instances

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceUtils.kt
```diff
@@ -0,0 +1,95 @@
...
+     * This will generate random IP address for loopback range.
+     *
+     * Note: there is a chance for randomly generated IP address from different
+     * execution to end-up closer to each other. Although the likelihood is low, but
+     * can happen. As the result when creating the next IP address from the last used,
+     * there will be negligible amount of small impact on the speed of the execution.
+     */
+    fun generateRandomIPAddress() : String {
+        val (p1, p2, p3) = Triple(
+            (0..255).shuffled().first(),
```
arcuri82, line 59: this introduce an un-controlled source of non-determinism.
should rather pass in as a parameter an instance of `org.evomaster.core.search.service.Randomness` (which is a singleton), and use that for shuffling (might need new method there)

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceInfo.kt
arcuri82: or, if keeping here, likely need new name for this class

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceInfo.kt
```diff
@@ -14,6 +19,10 @@ class ExternalServiceInfo(
         }
     }
 
+    fun assignWireMockServer(wm: WireMockServer) {
+        wireMockServer = wm
```
arcuri82, line 23: why not done in constructor? leaving potentially uninitialized `lateinit` fields is risky

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+     * Will return the next available IP address from the last know IP address
+     * used for external service.
+     */
+    private fun getNextAvailableAddress() : String {
+        val nextAddress: String = nextIPAddress(lastIPAddress)
+
+        if (isAddressAvailable(nextAddress, WIREMOCK_PORT)) {
+            return nextAddress
+        } else {
+            throw Exception(nextAddress.plus(" is not available for use"))
```
arcuri82, line 61: throw IllegalStateException

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+    }
+
+    /**
+     * Returns the default IP address for external service initialisation.
+     * If the respective address and port is not available it'll throw an
+     * exception.
+     */
+    private fun getDefaultAddress() : String {
+        if (isAddressAvailable("127.0.0.10", WIREMOCK_PORT)) {
+            return "127.0.0.10"
```
arcuri82, line 72: the "default" in EMConfig is different, ie, 127.0.0.2. is there any reason for this?

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+                .bindAddress(address)
+                .port(WIREMOCK_PORT)
+                .extensions(ResponseTemplateTransformer(false)))
+        wm.start()
+
+        // to prevent from the 404 when no matching stub below stub is added
+        wm.stubFor(get(urlMatching("/.*"))
+            .atPriority(2)
+            .willReturn(aResponse()
+                .withStatus(200)
```
arcuri82, line 142: if not found, shouldn't rather be a 404?

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+    /**
+     * Will initialise WireMock instance on a given IP address for a given port.
+     */
+    private fun initWireMockServer(address: String): WireMockServer {
+        val wm = WireMockServer(
+            WireMockConfiguration()
+                .bindAddress(address)
+                .port(WIREMOCK_PORT)
+                .extensions(ResponseTemplateTransformer(false)))
+        wm.start()
```
arcuri82, line 136: as we start servers, we ll need a method to shut them all down (eg, to be called at end of the search)

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+
+    /**
+     * Update the DNS cache with a different IP address for a given host to enable spoofing.
+     * If there is an entry already it'll skip from adding to the cache.
+     */
+    private fun updateDNSCache(host : String, address : String) {
+        val entry = DnsCacheManipulator.getDnsCache(host)
+        if (entry != null) {
+            DnsCacheManipulator.setDnsCache(host, address)
+        }
```
arcuri82, line 156: what if null? we should at least log error

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+    }
+
+    /**
+     * Clean up the DNS cache
+     */
+    fun cleanDNSCache() {
+        DnsCacheManipulator.clearDnsCache()
+    }
+
+    companion object {
```
arcuri82, line 166: companion objects should be on top of file, not bottom

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
```diff
@@ -1,26 +1,169 @@
...
+        private const val WIREMOCK_PORT : Int = 8080
```
arcuri82, line 167: why this? shouldn't be based on the port called from the SUT?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
seran: This was changed during a test, failed to revert back.

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceInfo.kt
seran: Original I had this class name as `ExternalService`, later we changed it to `Info` because it was confusing with `ExternalServices`.

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
seran: Most cases SUT port will be 80 or 443. To bind on those ports we need root privileges. Let me check this again, if it's working, I'll make the changes to use SUT port.

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServices.kt
seran: This is temporary. Will improve this code with upcoming changes.

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalService.kt
```diff
@@ -0,0 +1,13 @@
+package org.evomaster.core.problem.external.service
+
+import com.github.tomakehurst.wiremock.WireMockServer
+
+class ExternalService (
+    val externalServiceInfo: ExternalServiceInfo,
+    val wireMockServer: WireMockServer
```
arcuri82, line 7: i guess this should be `private`?

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceHandler.kt
```diff
@@ -145,14 +155,22 @@ class ExternalServices {
...
      * Update the DNS cache with a different IP address for a given host to enable spoofing.
      * If there is an entry already it'll skip from adding to the cache.
      */
     private fun updateDNSCache(host : String, address : String) {
         val entry = DnsCacheManipulator.getDnsCache(host)
-        if (entry != null) {
+        if (entry == null) {
             DnsCacheManipulator.setDnsCache(host, address)
+        } else {
+            log.warn("$host already has a DNS cache entry.")
```
arcuri82, line 173: naming is bit confusing, ie, `update` that gives a warning if already set. 
rather call it `bindDNSCache`, and return a boolean (no warn) to see if bound or not (ie already present)

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceUtils.kt
```diff
@@ -9,10 +11,15 @@ object ExternalServiceUtils {
 
     private val log: Logger = LoggerFactory.getLogger(ExternalServiceUtils::class.java)
 
+    @Inject
+    lateinit var randomness: Randomness
+
     /**
      * This method provides the next IP address from the given value for
      * loopback range. If generated IP address is not in the range, this
      * will throw an Exception.
```
arcuri82, line 20: not an `Exception`, but an `IllegalArgumentException`

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceUtils.kt
```diff
@@ -31,36 +38,53 @@ object ExternalServiceUtils {
...
-            (0..255).shuffled().first(),
-            (0..254).shuffled().first(),
+            randomness.randomIPBit(),
+            randomness.randomIPBit(),
+            randomness.randomIPBit(),
         )
-        return String.format("127.%s.%s.%s", p1, p2, p3)
+        var ip = String.format("127.%s.%s.%s", p1, p2, p3)
+
+        if (isReservedIP(ip)) {
```
arcuri82, line 84: there is a tiny probability that this can fail second time. should be inside a loop

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -452,4 +452,10 @@ class Randomness {
 
         return k
     }
+
+    fun randomIPBit() : Int {
+        val k = (0..255).shuffled().first()
```
arcuri82, line 457: this is still non-deterministic, as `.shuffled()` is not run on the current instance of `random`.
on a collection, you need to call `.shuffle(random)`

arcuri82 (changes requested):

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/Action.java
```diff
@@ -18,11 +15,17 @@ public class Action implements Serializable {
...
-    public Action(int index, Collection<String> inputVariables) {
+    /**
+     * A map of hostname and WireMock IP to mock external service calls.
+     */
+    private final Map<String, String> externalServiceMapping;
+
+    public Action(int index, Collection<String> inputVariables, Map<String, String> externalServiceMapping) {
         this.index = index;
         this.inputVariables = Collections.unmodifiableSet(new HashSet<>(inputVariables));
+        this.externalServiceMapping = externalServiceMapping;
```
arcuri82, line 28: should use `Collections.unmodifiable` to make it immutable, like `inputVariables`

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/URLClassReplacement.java
```diff
@@ -37,8 +38,38 @@ public static URLConnection openConnection(URL caller) throws java.io.IOExceptio
...
+            if (ExecutionTracer.hasExternalMapping(caller.getHost())) {
+                String ip  = ExecutionTracer.getExternalMapping(caller.getHost());
+
+                // Usage of ports below 1024 require root privileges to run
+                String url = "http://" + ip + ":" + caller.getPort() + caller.getPath();
+
+                URL newURL = new URL(url);
+                return newURL.openConnection();
+            } else {
+                throw new UnknownHostException("There is no WireMock initiated for this hostname");
```
arcuri82, line 71: can leave a TODO here, stating the message will likely need to be changed in the future.
if host is not reachable, exeption is thrown, and could lead to 500 with the stack-trace in response. in such case, assertions would be generated on response, which will then fail when running tests as the messages would be different when running without instrumentation. we ll likely see this problem when we will run the generated tests, and think of a solution then

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1673,6 +1669,34 @@ class EMConfig {
...
+        RANDOM
+    }
+
+    @Cfg("Specify a method to select the first external service spoof IP address.")
+    @Experimental
+    var externalServiceIPSelectionStrategy = ExternalServiceIPSelectionStrategy.NONE
+
+    @Cfg("User provided external service IP.")
+    @Experimental
+    var externalServiceIP : String = "127.0.0.2"
```
arcuri82, line 1699: can add a Regex constraint here

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceHandler.kt
```diff
@@ -0,0 +1,168 @@
...
+     */
+
+    /**
+     * Contains the information about each external calls made
+     */
+    private val externalServiceInfos: MutableList<ExternalServiceInfo> = mutableListOf()
+
+    private val externalServices: MutableList<ExternalService> = mutableListOf()
+
+    private var lastIPAddress : String = ""
```
arcuri82, line 34: add comments for this

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceHandler.kt
```diff
@@ -0,0 +1,168 @@
...
+     * Contains hostname against to WireMock instance mapping
+     */
+    private val externalServiceMapping: MutableMap<String, String> = mutableMapOf()
+
+    @Inject
+    private lateinit var randomness: Randomness
+
+    @Inject
+    private lateinit var config : EMConfig
+
```
arcuri82, line 46: put the `@Inject` variables at beginning of list of variables, ie move up before `externalServiceInfos`

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceHandler.kt
```diff
@@ -0,0 +1,168 @@
...
+            if (!externalServiceMapping.containsKey(externalServiceInfo.remoteHostname)) {
+                val ip = getIP(externalServiceInfo.remotePort)
+                lastIPAddress = ip
+                val wm : WireMockServer = initWireMockServer(ip, externalServiceInfo.remotePort)
+
+                externalServices.add(ExternalService(externalServiceInfo, wm))
+                externalServiceMapping[externalServiceInfo.remoteHostname] = ip
+            }
+        }
+        externalServiceInfos.add(externalServiceInfo)
```
arcuri82, line 64: what if already present? maybe instead of lists should use maps to keep those 2 lists in sync (maybe indexed on hostname)

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceHandler.kt
```diff
@@ -0,0 +1,168 @@
...
+                .bindAddress(address)
+                .port(port)
+                .extensions(ResponseTemplateTransformer(false)))
+        wm.start()
+
+        // to prevent from the 404 when no matching stub below stub is added
+        wm.stubFor(get(urlMatching("/.*"))
+            .atPriority(2)
+            .willReturn(aResponse()
+                .withStatus(200)
```
arcuri82, line 162: hmmm... leave a TODO. we ll need to discss what should be the default behavior when WM not configured yet. eg, maybe 500 or 404 could be better than 200. we can discuss it later on (ie, for now just leave a TODO)

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceUtils.kt
```diff
@@ -0,0 +1,118 @@
...
+            socket = Socket()
+            socket.connect(InetSocketAddress(address, port), 1000)
+            false
+        } catch (e: ConnectException) {
+            true
+        } catch (e: SocketTimeoutException) {
+            true
+        } catch (e: IOException) {
+            true
+        } finally {
```
arcuri82, line 108: can just have a single `catch` with those types concatenated with `|`

arcuri82 (changes requested):

PR #519, opened 2022-05-20
Gql bb v2

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
```diff
@@ -53,7 +53,15 @@ object GqlConst {
...
     const val ROOT = "root"
+    const val QUERY_ROOT = "query_root"
+    const val ROOTQUERY = "rootquery"
+    const val ROOTQUERYTYPE = "rootquerytype"
+    const val DOTAQUERY = "dotaquery"
+
     const val MUTATION = "mutation"
+    const val ROOTMUTATIONTYPE = "rootmutationtype"
+    const val ROOTMUTATION = "rootmutation"
+    const val DOTAMUTATION = "dotamutation"
```
arcuri82, line 65: please add some comments/documentation for these constants. eg, what is the difference between `QUERY_ROOT` and `ROOTQUERY`?

core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -70,6 +83,22 @@ class IntrospectiveQuery {
...
         val body = response.readEntity(String::class.java)
 
+        //TODO parse this body here to see if it has any "errors" field and no "data"
+
+        val jackson = ObjectMapper()
+
+        val node: JsonNode = try {
+            jackson.readTree(body)
+        } catch (e: JsonProcessingException) {
+            return null
```
arcuri82, line 93: this should not be ignored.
if what returned is not a JSON, should log it as an error and throw a `SutProblemException` with a meaningful error message

core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -70,6 +83,22 @@ class IntrospectiveQuery {
...
+        }
+
+        val withErrors= node.findPath("errors")
+
+       if (!withErrors.isEmpty){
+            throw SutProblemException("Failed to retrieve GraphQL schema: response contains error.")
```
arcuri82, line 99: add here the message of the error. could simply output `body`

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
```diff
@@ -49,7 +49,10 @@ object GqlConst {
...
     const val LIST = "list"
     const val INPUT_OBJECT = "input_object"
-    /**/
+    /*
+    *Those are entry points of GraphQL query and mutation
+    * Queries and mutations names are defined by developers, so they can have different names
+    * eg: query, rootquery ...  */
     const val QUERY = "query"
     const val QUERY_TYPE = "querytype"
     const val ROOT = "root"
```
arcuri82, line 58: names could be anything. cannot hardcode here.
need to check based on actual type in schema, NOT the name.
see https://graphql.org/learn/schema/

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
arcuri82: create a test in which we have something like `foo: Query`

asmab89: Done

core/src/main/kotlin/org/evomaster/core/problem/graphql/GqlConst.kt
arcuri82: did u add the documentation? or, if those are not used anymore, should be removed from the code

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
```diff
@@ -208,6 +208,8 @@ object GraphQLUtils {
      * graph: a map with the name of the node as a key and its fields and edges as values,
      * history: used in the cycles management,
      * objectFieldsHistory: a set used in the management of already constructed nodes
+
+     * TODO Remove isRoot(), and extract the names directly from the schema
```
arcuri82, line 212: is this TODO still valid?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
asmab89: Done.

PR #520, opened 2022-05-20
Refactor db specification

arcuri82:


arcuri82:
thx. i already updated EMB, in another PR :)

PR #524, opened 2022-05-27
Bb url call swagger

ZhangMan1126:
handle url for call swagger for bb

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
@@ -8,6 +8,7 @@ import io.swagger.v3.oas.models.media.MediaType
 import io.swagger.v3.oas.models.media.ObjectSchema
 import io.swagger.v3.oas.models.media.Schema
 import io.swagger.v3.oas.models.parameters.Parameter
+import org.evomaster.core.Lazy
 import org.evomaster.core.logging.LoggingUtil
 import org.evomaster.core.parser.RegexHandler
 import org.evomaster.core.problem.api.service.param.Param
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -784,4 +785,24 @@ object RestActionBuilderV3 {
...
+        // if the url does not start with baseUrl (i.e., not from SUT), then there might be no point to execute this rest action
+        if (!url.startsWith(baseUrl)) return null
+
+        val uri = URI(url)
+//        Lazy.assert { "${uri.scheme}://${uri.host}:${uri.port}" == baseUrl }
+
+        val path = RestPath("${uri.scheme}://${uri.host}:${uri.port}${uri.path}".removePrefix(baseUrl.removeSuffix("/")))
+        val query : MutableList<Param> = uri.query?.split("&")?.map { q->
+            val keyValue = q.split("=")
+            Lazy.assert { keyValue.size == 2 }
```
arcuri82, line 802: this is not fully correct: see [https://datatracker.ietf.org/doc/html/rfc3986#section-3.4]()

rather than crash here, should issue a warning that non-key-pairs are not currently supported.
also , could have an empty string value in the form of `key=`, ie, no char after the `=`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -796,11 +796,23 @@ object RestActionBuilderV3 {
...
+            else {
+                /*
+                    key-value pair is not restricted for query
+                    eg, /v2/api-docs?foo is considered as valid
+                    see https://datatracker.ietf.org/doc/html/rfc3986#section-3.4
+                 */
+                log.warn("Do not support the get a RestAction with the url $url")
+                return null
+            }
         }?.toMutableList()?: mutableListOf()
```
arcuri82, line 816: maybe `filter` out the `null` elements in this list?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
arcuri82: ah, didn't realized that `return` is actually getting out of the function... :)

PR #526, opened 2022-06-01
fix mutation that might generate invalid value

ZhangMan1126:


core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
```diff
@@ -64,9 +64,10 @@ object GeneUtils {
         val maxIndex = apc.getExploratoryValue(start, end)
 
         var n = 0
-        for (i in 0 until maxIndex) {
+        for (i in 0 until (maxIndex -1)) {
             n = i + 1
-            if (intpow2[i] > range) {
+            // check with Andrea regarding n instead of i
+            if (intpow2[n] > range) {
```
arcuri82, line 70: what was the reason for this change?

ZhangMan1126: it would generate `delta` which is out of the specified range.

if set `n` with `intpow2[i] > range`, actually the `n-1` is already more than the range. then it would have a chance to select it, i.e., n-1.
this could be reproduced by the tests provided by JP with iteration 2.

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/GeneUtils.kt
ZhangMan1126: unit tests are added now :)

PR #529, opened 2022-06-03
handling ?# in WB as well when calling OpenAPI in the generated tests

arcuri82:


core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -793,27 +793,23 @@ object RestActionBuilderV3 {
...
+            if (keyValue.size == 2 && keyValue[0].isNotBlank())
                 QueryParam(keyValue[0], StringGene(keyValue[0], keyValue[1]))
             else {
                 /*
                     key-value pair is not restricted for query
                     eg, /v2/api-docs?foo is considered as valid
                     see https://datatracker.ietf.org/doc/html/rfc3986#section-3.4
                  */
-                log.warn("Do not support the get a RestAction with the url $url")
+                log.warn("Currently noy supporting a GET RestAction with the url '$url' ," +
```
ZhangMan1126, line 811: noy -> not ?

ZhangMan1126 (commented):
thanks! I just have one minor comment. 
btw, there is a PR branched from this, then you might merge it to this first.

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
arcuri82: thx. fixed

PR #540, opened 2022-06-13
Fix min max

ZhangMan1126:


ZhangMan1126:
for tests in `GeneTest` , except `testRandomized`  with seed `1762`, all other tests pass now.

Regarding `testRandomized`  with seed `1762`, a `flatview` of `SqlMultidimensionalArrayGene` would result in 3346383 elements that leads to  `java.lang.OutOfMemoryError: GC overhead limit exceeded`

PR #544, opened 2022-06-17
e2e test cases for TRACE, OPTIONS and HEAD http verbs

jgaleotti:


e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/endpoints/EndpointsEMTest.java
```diff
@@ -0,0 +1,35 @@
...
+        runTestHandlingFlakyAndCompilation(
+                "EndpointsEM",
+                "org.foo.EndpointsEM",
+                500,
+                (args) -> {
+
+                    Solution<RestIndividual> solution = initAndRun(args);
+
+                    assertTrue(solution.getIndividuals().size() >= 1);
+
```
arcuri82, line 32: here we should  make sure to assert each endpoint has been called.

as RestAssured has no `trace`, shouldn't the test fails when compiled???

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/endpoints/EndpointsEMTest.java
jgaleotti: I added assertHasAtLeastOne() to check for TRACE, OPTIONS and HEAD.
The test case is currently failing since the TRACE command is not generated.

arcuri82: thx! i ll make a fix (but unlikely today)

PR #545, opened 2022-06-18
upgrade to H2 version 2.1.212

jgaleotti:


pom.xml
```diff
@@ -582,12 +582,12 @@
             <dependency>
                 <groupId>com.h2database</groupId>
                 <artifactId>h2</artifactId>
-                <version>1.4.200</version>
```
arcuri82, line 585: upgrading default to 2.1.212 is good ;)

however, as `1.x` is used in EMB, we should have a new module in which we have at least 1 test checking for `1.x`

arcuri82 (changes requested):

jgaleotti:
I added a new module (spring-rest-h2-v1) with e2e test cases for H2 v1.4.200

pom.xml
jgaleotti: I added a new module (spring-rest-h2-v1) with e2e test cases for H2 v1.4.200

PR #548, opened 2022-06-23
Fix refactor v2

ZhangMan1126:
with this version, it could generate tests for 12 JVM case studies of EMB with 10k search budget.

fix several problems
add tests to check sampled/mutated individuals regarding 
- binding 
- sequence of different types of children (db action, rest resource call) 
- consistence of impa
- impact update


arcuri82:
@ZhangMan1126 thx. i ll merge once CI is completed

PR #549, opened 2022-06-24
Rpc fix

man-zhang:
fix some problems in a sequence of building object gene
fix seeded test dto with `null`


client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/schema/params/ByteBufferParam.java
```diff
@@ -62,6 +62,7 @@ protected void setValueBasedOnValidInstance(Object instance) {
 
     @Override
     public void setValueBasedOnInstanceOrJson(Object json) throws JsonProcessingException {
+        if (json == null)  return;
         assert json instanceof String;
```
arcuri82, line 66: `assert` would be used for `private` methods, to assert something that should never happen, otherwise it is a bug.
but, for a `public`, it should rather throw a `IllegalArgumentException`. Recall [this discussion](https://github.com/EMResearch/EvoMaster/blob/master/docs/for_developers.md#pre-and-post-conditions).

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/schema/params/ByteBufferParam.java
ZhangMan1126: thanks! ll update them and request a review once tests pass

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/rpc/ParamDto.java
```diff
@@ -10,6 +10,8 @@
  */
 public class ParamDto {
 
+    public final static String NOT_NULL_MARK_OBJ_DATE = "{}";
```
arcuri82, line 13: add a bit more info/comment about this `public` parameter

arcuri82 (changes requested):

arcuri82 (changes requested):

PR #550, opened 2022-06-26
H2 column types and e2e test cases

jgaleotti:


core/src/main/kotlin/org/evomaster/core/database/schema/h2/H2GeometryTypeParser.kt
```diff
@@ -0,0 +1,71 @@
...
+     * Requires the string to be a valid H2 Geometry type definition
+     * [isParsable()] returns true.
+     * It returns a H2GeometryType instance with the geometricObject,
+     * the dimension (if any) and the spatialReferenceSystemIdentifier (if any)
+     */
+    fun parse(typeAsString: String): H2GeometryType {
+        if (!isParsable(typeAsString)) {
+            throw IllegalArgumentException("Cannot translate an invalid H2 geometry type definition: $typeAsString")
+        }
+        val pattern = Pattern.compile(H2_GEOMETRY_REGEX)
```
arcuri82, line 62: i guess the pattern can be computed just once, and saved in the companion object?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/database/schema/h2/H2GeometryTypeParser.kt
jgaleotti: good point. I pushed a fix to save the pattern in the companion object.

PR #556, opened 2022-07-05
fix disabled tests ie, `ProcessMonitorTest` and `MioAlgorithmOnTrackO…

ZhangMan1126:
…neMaxTest`


ZhangMan1126:
@arcuri82 thanks! the problem is solved

PR #559, opened 2022-07-10
pass on local then check on CI

ZhangMan1126:
the disabled test seems pass now.


jgaleotti (commented):
Sorry, I am a bit list. I do not see any changes to other files except GeneTest.kt.

jgaleotti:
Great, I will merge it then.

PR #563, opened 2022-08-04
Child grouping

arcuri82:


arcuri82:
@ZhangMan1126  hi. do you have any comments on these new data structures? before starting to modify samplers/mutators (eg starting from GraphQL) wanted to hear your opinion. also, this will impact the refactoring of DbAction in resource-handling

ZhangMan1126 (approved):
thanks!

PR #564, opened 2022-08-04
WireMock and DNS cache manipulation under generated test cases

seran:


seran:
Test cases for Java not completed yet. Will update once those are done.

seran:
Java also, completed. @ZhangMan1126

ZhangMan1126 (commented):
thanks, Seran! I have no comment on the code, but need a bit further check with you on tests on macOS.

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -443,6 +456,9 @@ class TestSuiteWriter {
             } else {
                 lines.add("private static String $baseUrlOfSut = \"${BlackBoxUtils.targetUrl(config, sampler)}\";")
             }
+            if (useWireMock()) {
+                lines.add("private static WireMockServer wireMockServer;")
```
arcuri82, line 460: is there going to be only one single server???
what if the SUT makes two calls to two different services using different ports?
wouldn't that need 2 WM instances running on 2 different IP loopback addresses?

arcuri82: could be fine to support only 1 single instance for now, but leave TODO comment, as ll need to be handled in next iterations (eg 1 of industrial SUTs uses at least 4 different services)

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -752,4 +854,8 @@ class TestSuiteWriter {
 
 
     private fun useRestAssured() = config.problemType != EMConfig.ProblemType.RPC
+
+    private fun useWireMock(): Boolean {
```
arcuri82, line 858: this check should be further extended. if no test case in the whole solution has any WM handling, there is no point to add wiremock in the generated JUnit file

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceAction.kt
```diff
@@ -89,28 +89,49 @@ class ExternalServiceAction(
...
-                        aResponse()
-                            .withStatus(viewStatus())
-                            .withBody(viewResponse())
-                    )
-                    .withMetadata(
-                        metadata()
-                            .attr("url", request.url)
-                    )
-            )
+            if (request.method.lowercase() == "get") {
```
arcuri82, line 106: what about all the other verbs like PATCH, PUT and DELETE?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/external/service/ExternalServiceAction.kt
seran: For now focused on GET and POST to make the basics work. Will add that too.

arcuri82: that's ok. but leave a TODO comment in the code

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
seran: Modified it to have more servers.

arcuri82:
hi @seran 
  was this PR ready to be reviewed after your changes?

seran:
@arcuri82 requested changes are done. But the generated test cases aren't accurate. Have to work on them once the actions refactoring is completed. It's not good to merge this branch now.

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -76,6 +82,10 @@ abstract class TestCaseWriter {
         lines.indented {
             val ind = test.test
             val insertionVars = mutableListOf<Pair<String, String>>()
+            val actions = getExternalServiceActions(ind)
```
ZhangMan1126, line 85: as we discussed, we handle external actions for each `RestCallAction` in terms of `EnterpriseGroup`, then reset the external between the group. 
here, you handle all external actions at the beginning of the test, right? 
it seems not correct.

ZhangMan1126 (changes requested):

ZhangMan1126 (commented):
thanks, Seran! you could request Andrea's review then.

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -210,6 +212,24 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
         lines: Lines,
         baseUrlOfSut: String
     ) {
+
+        val exActions = mutableListOf<HttpExternalServiceAction>()
+
+        // add all used external service actions for the action
+        if (TestWriterUtils.handleExternalService(config)) {
```
arcuri82, line 219: this `handleExternalService` should be moved directly inside `EMConfig` (see all similar methods there at the bottom).
Also, as it returns a boolean, should be renamed to represent it, eg, `isEnabledExternalServiceMocking`

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -210,6 +212,24 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
+            if (!format.isJavaOrKotlin()) {
+                log.warn("NOT support for other format ($format) except JavaOrKotlin")
```
arcuri82, line 221: this would print this  log each time we do not run on JVM, which is not correct. here, a further check that is needed is to see if there is an ExternalAction. if there is no ExternalAction, no warning should be printed

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -220,6 +240,21 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
         } else {
             addActionLines(call, lines, res, baseUrlOfSut)
         }
+
+        // reset all used external service action
+        if (exActions.isNotEmpty()) {
+            if (!format.isJavaOrKotlin()) {
+                log.warn("NOT support for other format ($format) except JavaOrKotlin")
```
arcuri82, line 247: this log without context would be confusing... add something like `In mocking of external services, we do NOT ...`

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -88,14 +93,66 @@ abstract class TestCaseWriter {
...
+    protected fun handleExternalServiceActions(
+        lines: Lines,
+        actions: List<HttpExternalServiceAction>
+    ) {
+        // TODO: Handle same request pattern exists multiple time.
+        //  Currently we don't handle the same request pattern exists multiple
+        //  time. For that, need to create WireMock scenario to have multiple
+        //  responses for the same request pattern.
+        actions
+            .filter { action -> action.active }
```
arcuri82, line 105: recall the implicit `it`. this could be written as `.filter { it.active }`

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -88,14 +93,66 @@ abstract class TestCaseWriter {
...
+                        lines.indented {
+                            lines.add(".withStatus(${response.status.getValueAsRawString()})")
+                            // TODO: Response is empty for now
+                            lines.add(".withBody(\"${response.response.getValueAsRawString()}\")")
+                        }
+                        lines.add(")")
+                    }
+                }
+                lines.add(")")
+                if (format.isJava()) {
```
arcuri82, line 136: why the check on `format.isJava()`?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -550,17 +592,54 @@ class TestSuiteWriter {
...
+                                }
+
+                                if (format.isKotlin()) {
+                                    addStatement(".extensions(ResponseTemplateTransformer(false)))", lines)
+                                }
+                            }
+                            addStatement("${name}.start()", lines)
+                        }
+                } else {
+                    log.warn("NOT support for other format ($format) except JavaOrKotlin")
```
arcuri82, line 630: recall previous comment about this

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -752,4 +844,21 @@ class TestSuiteWriter {
 
 
     private fun useRestAssured() = config.problemType != EMConfig.ProblemType.RPC
+
+    /**
+     * Returns a distinct List of [HttpExternalServiceAction] from the given solution
+     */
+    private fun getWireMockServers(solution: Solution<*>): List<HttpExternalServiceAction> {
```
arcuri82, line 851: the name is confusing... i mean, we could return several actions for the same `Server`. so, maybe `getWireMockServerActions`?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -752,4 +844,21 @@ class TestSuiteWriter {
...
+        val actions = mutableListOf<HttpExternalServiceAction>()
+        solution.individuals.filter { i -> i.individual is RestIndividual }
+            .forEach {
```
arcuri82, line 854: could simplify this code with a `flatMap`

core/src/main/kotlin/org/evomaster/core/output/service/TestWriterUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Helper functions for [TestCaseWriter] and [TestSuiteWriter]
+ */
+class TestWriterUtils {
+    companion object {
+        /**
+         * Check the current [EMConfig] and return boolean as the result
+         */
+        fun handleExternalService(config: EMConfig): Boolean {
+            if (config.externalServiceIPSelectionStrategy != EMConfig.ExternalServiceIPSelectionStrategy.NONE) {
```
arcuri82, line 15: can simplify with simple: `return config.externalServiceIPSelectionStrategy != EMConfig.ExternalServiceIPSelectionStrategy.NONE` 
also, see previous comment about rename and move to `EMConfig`

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/HttpExternalServiceAction.kt
```diff
@@ -122,6 +124,39 @@ class HttpExternalServiceAction(
...
         )
+
+    }
+
+    /**
+     * Will return a [MappingBuilder] based on the HTTP method
+     * TODO: Moved it to a ResponseBuilder, later
+     */
+    private fun getRequestMethod(request: HttpExternalServiceRequest): MappingBuilder {
+        val response = when (request.method) {
```
arcuri82, line 135: maybe force here a `.toUpperCase()`?

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/HttpExternalServiceAction.kt
```diff
@@ -122,6 +124,39 @@ class HttpExternalServiceAction(
...
+            "GET" -> get(getUrlPattern(request.url))
+            "POST" -> post(getUrlPattern(request.url))
+            "PUT" -> put(getUrlPattern(request.url))
+            "PATCH" -> patch(getUrlPattern(request.url))
+            "DELETE" -> delete(getUrlPattern(request.url))
+            "HEAD" -> head(getUrlPattern(request.url))
+            "TRACE" -> trace(getUrlPattern(request.url))
+            "OPTIONS" -> options(getUrlPattern(request.url))
+            else -> any(getUrlPattern(request.url))
```
arcuri82, line 144: i guess should issue a warning as well here

core/src/test/kotlin/org/evomaster/core/output/TestCaseWriterTest.kt
```diff
@@ -15,15 +15,17 @@ import org.evomaster.core.output.service.RestTestCaseWriter
 import org.evomaster.core.problem.rest.*
```
arcuri82, line 15: can u revert the formatting here? with all the formatting changes, it is hard to see which are the new modifications...

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/wiremock/service/ExternalServiceMockingFlakyEMTest.java
```diff
@@ -26,21 +25,21 @@ public static void initClass() throws Exception {
...
     public void externalServiceMockingTest() throws Throwable {
-        runTestHandlingFlaky(
-                "ExternalServiceMockingEMTest",
-                "org.bar.ExternalServiceMockingEMTest",
-                500,
-                false,
+        runTestHandlingFlakyAndCompilation(
+                "ExternalServiceMockingEMGeneratedTest",
+                "org.bar.ExternalServiceMockingEMGeneratedTest",
+                1000,
```
arcuri82, line 36: make sure u enable generation of tests for all the E2E tests for WM

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
seran: From the previous implementation. Forget to remove it.

PR #580, opened 2022-09-06
WireMock actions refactor with EnterpriseActionGroup

seran:


core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                if (parent !is EnterpriseActionGroup) {
+                    //TODO this should not really happen
+                    val msg = "Action is not inside an EnterpriseActionGroup"
+                    log.error(msg)
+                    throw RuntimeException(msg)
+                }
+
+                // If the action already exists mark it as used
+                esr.flatMap { it.value }.forEach { url ->
+                    val existingActions = parent.getViewOfChildren()
```
ZhangMan1126, line 74: you could filter `esr[index]` first based on existing `external services` in this group.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                        .filterIsInstance<HttpExternalServiceAction>()
+                        .filter { it.request.absoluteURL == url }
+
+                    val actions = sampler.getExternalService()
+                        .getExternalServiceActions()
+                        .filter { it.request.absoluteURL == url }
+
+                    if (existingActions.isEmpty()) {
```
ZhangMan1126, line 82: if filter existing action first, might not need this check.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                        if (actions.isNotEmpty()) {
```
ZhangMan1126, line 83: if the external service is visited, but it does not have an instance for it. it might be a bug. maybe have `Lazy.assert` here

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                        }
+                    } else {
+                        /*
+                            nothing was called so anything there from previous setups can be safely removed.
+                            however, "removing" could have some side-effects (TODO explains), so we just mark them
+                            as disabled
+
+                            Code never reaches this part
+                        */
+                        parent.getViewOfChildren()
```
ZhangMan1126, line 103: As we discussed, we do not `confirmNotUsed` here, we did it after fitness evaluation.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                    if (existingActions.isEmpty()) {
+                        if (actions.isNotEmpty()) {
+                            parent.addChildrenToGroup(
+                                actions,
+                                GroupsOfChildren.EXTERNAL_SERVICES
+                            )
+
+                            // update impact based on added genes
+                            // TODO: Man to review this..
+                            if (mutatedGenes != null && actions.isNotEmpty() && config.isEnabledArchiveGeneSelection()) {
```
ZhangMan1126, line 91: maybe better collect all newly added external services actions, then update impacts at end of this method, ie, after the block of `if (log.isTraceEnabled)`

core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
```diff
@@ -63,4 +63,11 @@ abstract class ApiExternalServiceAction(
     fun confirmNotUsed() {
         this.used = false
     }
+
+    /**
+     * make the action active explicitly based on [active]
+     */
+    fun activate() {
```
ZhangMan1126, line 70: why do we need this? active should be always based on `used`

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
```diff
@@ -109,38 +107,60 @@ class ExternalServiceHandler {
...
-                            counter++
-                        )
+                    val action = HttpExternalServiceAction(
+                        it,
+                        "",
+                        u,
+                        counter++
                     )
+                    action.doInitialize(randomness)
+                    action.confirmUsed() // TODO comment
```
ZhangMan1126, line 143: do not update the property here. always update it after the fitness evaluation.

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
```diff
@@ -200,22 +220,19 @@ class ExternalServiceHandler {
...
+        // WireMock throws an exception when there is no stub for the request
+        // to avoid the exception it handled manually
         wm.stubFor(
             any(anyUrl())
                 .atPriority(10)
                 .willReturn(
                     aResponse()
-                        .withStatus(500)
-                        .withBody("Internal Server Error")
+                        .withStatus(404)
```
ZhangMan1126, line 230: are the `404` and `Not Found` default settings of wiremock instance? then it might better to define it as private constant variable.

ZhangMan1126 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,134 @@ import kotlin.math.min
...
+                    throw RuntimeException(msg)
+                }
+
+                // If the action already exists mark it as used
+                esr.flatMap { it.value }.forEach { url ->
+                    val existingActions = parent.getViewOfChildren()
+                        .filterIsInstance<HttpExternalServiceAction>()
+                        .filter { it.request.absoluteURL == url }
+
+                    val actions = sampler.getExternalService()
```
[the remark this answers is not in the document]
seran: Going to introduce a new method in later tasks. Will change during that. Now leaving a comment.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
seran: There are chance, whether there is an action exists for the absoluteURL. If this is wrong need to investigate.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
```diff
@@ -58,43 +71,97 @@ class RestResourceFitness : AbstractRestFitness<RestIndividual>() {
...
+                    because new code is reached (eg due mutation of some genes).
+                    We do not want such new call to re-use a mocking from a previous action which were left on.
+                    So 2 options: (1) always reset before calling SUT, or (2) always reset after call in which external
+                    setups were used.
+                 */
+                externalServiceHandler.resetServedRequests()
+
+                val externalServiceActions = a.getExternalServiceActions()
+
+                externalServiceActions.filterIsInstance<HttpExternalServiceAction>().forEach {
```
ZhangMan1126, line 104: 1) before building the response, might `resetActive` first, i.e., set `active` based on `used`
2) when building the response, filter the actions whose `active` is true

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
```diff
@@ -58,43 +71,97 @@ class RestResourceFitness : AbstractRestFitness<RestIndividual>() {
...
+                //  made based on the absolute URL. Otherwise, mark it not used
+                requestedExternalServiceUrls.forEach { url ->
+                    externalServiceActions.filterIsInstance<HttpExternalServiceAction>().forEach { action ->
+                        if (action.request.absoluteURL == url) {
+                            action.confirmUsed()
+                        } else {
+                            // Code never reaches this part
+                            action.confirmNotUsed()
+                        }
+                        action.resetActive()
```
ZhangMan1126, line 146: might not `resetActive` here, otherwise the actual `active` info would be missing

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
```diff
@@ -109,38 +112,79 @@ class ExternalServiceHandler {
...
         return actions
     }
 
+    /**
+     * Return a list of [HttpExternalServiceAction]s for the given absoluteURL.
+     *
+     * There is a possibility to have multiple requests made for the same URL several,
+     * so it returns a list of actions.
+     */
+    fun getExternalServiceActionsForURL(url: String): List<HttpExternalServiceAction> {
```
ZhangMan1126, line 155: this method is used to create external service actions for a particular rest call.
there could exist multiple requests made for the same URL, but are they all related to the rest call?

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,118 @@ import kotlin.math.min
...
+                if (esr.containsKey(index)) {
+                    val urls = esr[index]
+
+                    if (urls!!.isNotEmpty()) {
+                        val actions = urls.map { url ->
+                            sampler.getExternalService()
+                                .getExternalServiceActionsForURL(url)
+                        }.flatten()
+
+                        if (actions.isNotEmpty()) {
```
ZhangMan1126, line 82: might `confirmeUsed` for all newly created actions

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +26,118 @@ import kotlin.math.min
...
+                val parent = action.parent
+                if (parent !is EnterpriseActionGroup) {
+                    //TODO this should not really happen
+                    val msg = "Action is not inside an EnterpriseActionGroup"
+                    log.error(msg)
+                    throw RuntimeException(msg)
+                }
+
+                if (esr.containsKey(index)) {
+                    val urls = esr[index]
```
ZhangMan1126, line 74: we might need to filter the urls which do not exist in the existing external service actions

ZhangMan1126 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
seran: 2. is already there, added `resetActive`

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
seran: I'm not sure. If the core is multithreaded there is a possibility. Otherwise, all will be related to the same REST call.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
seran: You mean, filter them out or add new action only for them?

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
ZhangMan1126: a bit unsure.  if there exist different rest actions which access the same URL, one made two requests to the URL, and the other made one. Here, the input argument is the URL. how do you know how many external actions we should return? In addition, if it is multithreaded, do we create two wiremock instances for the same URL at the same time?

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
ZhangMan1126: if the URL is added in the `fv`, it indicates that it has been used, then we confirm it here.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
ZhangMan1126: `ers[index].filterNot{existsURL.contains(it)}`, otherwise the duplicated external service actions might be added

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
ZhangMan1126: 1 should before this `forEach`, otherwise it is too late.
```
externalServiceActions.forEach{it.resetActive()}
externalServiceActions.filter{it.active}.filterIsInstance<HttpExternalServiceAction>().forEach{...}
```

ZhangMan1126 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
seran: What about when we create the action?

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
seran: We don't create two WireMock, but I was thinking like what if same REST call invoked twice.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
seran: Previous version had a validation for this, recent update I removed it. I can add something similar to prevent this.

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/ExternalServiceHandler.kt
ZhangMan1126: If there do not exist multiple WireMock instances for the same URL, in case that you create two external service actions, then the first one will never work since the 2nd will override the response. right?

`same REST call invoked twice`, does it indicate the REST action? 
if yes, we will have two REST actions in the individual, each of them is under `EnterpiseGroup`, then we could create the corresponding external actions for each.
If here it indicates the call to external service, based on my understanding, we do not support a sequence of responses for the same URL yet, e.g., give different responses based on the sequence  (could support it in the other branch as we discussed). 
in case there exist two external service actions referring to the same URL, how will you handle it now?

seran: I'm not so sure about the scenario as I mentioned in my previous comment. Right now there will only one action for the call. Need to check what will happen if there is a query parameter in the request.

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +27,126 @@ import kotlin.math.min
...
+            val parent = action.parent
+            if (parent !is EnterpriseActionGroup) {
+                //TODO this should not really happen
+                val msg = "Action is not inside an EnterpriseActionGroup"
+                log.error(msg)
+                throw RuntimeException(msg)
+            }
 
-            ind.addInitializingDbActions(0, actions)
+            if (esr.containsKey(index)) {
```
ZhangMan1126, line 72: as we discussed, add a doc about here we only handling of adding new external services here. regarding the external service actions which are not used, we handled them after fitness evaluation with `used` property

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
```diff
@@ -58,43 +70,100 @@ class RestResourceFitness : AbstractRestFitness<RestIndividual>() {
...
+                    )
+                }
+
+                // Mark the existing external service actions as used if there is call
+                //  made based on the absolute URL. Otherwise, mark it not used
+                requestedExternalServiceRequests.forEach { request ->
+                    externalServiceActions.filterIsInstance<HttpExternalServiceAction>().forEach { action ->
+                        if (action.request.absoluteURL == request.absoluteURL) {
+                            action.confirmUsed()
+                        } else {
```
ZhangMan1126, line 145: the if-else logic here is incorrect. you need to check the existing action regarding all of `requestedExternalServiceRequests` not one of them, eg, the existing action does not appear in any `requestedExternalServiceRequests`, we confirmed it `notUsed`

In addition, times need to be checked. Imagine that there exist two external service actions, but only one is accessed. then need to mark the first one `confirmUsed` and the second one `confirmNotUsed`

ZhangMan1126 (changes requested):

ZhangMan1126 (commented):
thanks, Seran! I do not have further comments. You could request a review from Andrea

seran:
Thank you @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -25,56 +27,132 @@ import kotlin.math.min
...
+            }
+        }
+
+        // all actions should have local ids
+        Lazy.assert {
+            ind.seeAllActions().all { it.hasLocalId() }
+        }
+
+        if (log.isTraceEnabled)
+            log.trace("{} existingExternalServiceData are added")
```
arcuri82, line 115: why the `{}` here? you are not interpolating anything in the log. did u miss a `, x.size())`  as second parameter of the `trace`?

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/HttpExternalServiceRequest.kt
```diff
@@ -8,11 +8,20 @@ import java.util.UUID
...
+    /**
+     * Most likely the UUID created by the respective WireMock
+     * while capturing the request.
+     */
     val id: UUID,
     val method: String,
     val url: String,
     val absoluteURL: String,
     val wasMatched: Boolean,
+    val wireMockSignature: String
```
arcuri82, line 21: explain in comment what `wireMockSignature` is

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/param/HttpWsResponseParam.kt
```diff
@@ -9,13 +9,15 @@ import org.evomaster.core.search.gene.optional.OptionalGene
...
+     * Avoid having 404 as status, since WireMock will return 404 if there is
+     * no stub for the specific request.
      */
-    val status: EnumGene<Int> = EnumGene("status", listOf(200, 400, 401, 403, 404, 500)),
+    val status: EnumGene<Int> = EnumGene("status", listOf(200, 400, 401, 403, 500)),
     /**
      * Response content type, for now supports only JSON
      */
-    responseType: EnumGene<String> = EnumGene("responseType", listOf("JSON")),
+    responseType: EnumGene<String> = EnumGene("responseType", listOf("JSON", "XML")),
```
arcuri82, line 20: remove `XML`, as we have no support for it right now. you can put it as a comment, eg `//TODO might want to support XML here in the future`

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -142,6 +148,7 @@ abstract class StructuralElement (
 
     fun addChildToGroup(position: Int, child: StructuralElement, groupId: String){
         verifyChildrenToInsert(child)
+        preChildrenSetup(listOf(child))
```
arcuri82, line 151: shouldn't this be done after checking validity of inputs?

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -169,6 +176,7 @@ abstract class StructuralElement (
      */
     open fun addChild(child: StructuralElement){
         verifyChildrenToInsert(child)
+        preChildrenSetup(listOf(child))
```
arcuri82, line 179: shouldn't this be done after checking validity of inputs?

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -178,6 +186,7 @@ abstract class StructuralElement (
 
     open fun addChild(position: Int, child: StructuralElement){
         verifyChildrenToInsert(child)
+        preChildrenSetup(listOf(child))
```
arcuri82, line 189: shouldn't this be done after checking validity of inputs?

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -192,6 +201,7 @@ abstract class StructuralElement (
 
     open fun addChildren(position: Int, list : List<StructuralElement>){
         verifyChildrenToInsert(list)
+        preChildrenSetup(list)
```
arcuri82, line 204: shouldn't this be done after checking validity of inputs?

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -132,6 +137,7 @@ abstract class StructuralElement (
 
     fun addChildToGroup(child: StructuralElement, groupId: String){
         verifyChildrenToInsert(child)
+        preChildrenSetup(listOf(child))
```
arcuri82, line 140: shouldn't this be done after checking validity of inputs?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
seran: This was changed by @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
seran: This was changed by @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
seran: This was changed by @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
seran: This was changed by @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
seran: This was changed by @ZhangMan1126

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
seran: Yes, size of the newly added actions. Fixed now.

core/src/main/kotlin/org/evomaster/core/problem/external/service/httpws/param/HttpWsResponseParam.kt
seran: When we have single value, had a assertion failure in `Gene.standardMutation()`. `responseType` gene is mutable.

seran: Can remove this, since we don't use this anywhere else.

arcuri82: should remove. if an assertion fails, then it means there is a bug somewhere. in which test did it happen? i can look at it

seran: https://github.com/EMResearch/EvoMaster/blob/3a78fe11e573f1ea2436cad7b8c040abef346bc9/core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt#L515

This assertion.

arcuri82: yes, but which test case led to that assertion to fail? was one of the new E2E for wiremock?

seran: Yes.

`src/test/java/org/evomaster/e2etests/spring/examples/wiremock/service/ExternalServiceMockingFlakyEMTest.java`

arcuri82: I just pushed a fix for that bug

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
arcuri82: I pushed a change, in which I moved it after the input validation (as well for the other cases)

arcuri82:
@seran hi. build failed. can u have a llook at it? thx

seran:
@arcuri82 sure

PR #581, opened 2022-09-06
Api external service

ZhangMan1126:
add super class `ApiExternalServiceAction`

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
@@ -4,6 +4,15 @@ import org.evomaster.core.problem.external.service.param.ResponseParam
 import org.evomaster.core.search.Action
 
 abstract class ApiExternalServiceAction(
+    /**
+     * response to return if the external service is accessed
+     *
+     * To Andrea,
+     * a type of external service might contain a certain type of response
+     * eg, HTTP response
+     * Do we need to make the response as a generic type T where T extends ResponseParam
+     * and bind the ApiExternalServiceAction with the T?
+     */
     val response: ResponseParam,
     active : Boolean = false,
     used : Boolean = false,
--- a/core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
@@ -1,0 +1,62 @@
+package org.evomaster.core.problem.external.service
+
+import org.evomaster.core.problem.external.service.param.ResponseParam
+import org.evomaster.core.search.Action
+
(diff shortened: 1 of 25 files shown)


core/src/main/kotlin/org/evomaster/core/problem/external/service/ApiExternalServiceAction.kt
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * response to return if the external service is accessed
+     *
+     * To Andrea,
+     * a type of external service might contain a certain type of response
+     * eg, HTTP response
+     * Do we need to make the response as a generic type T where T extends ResponseParam
+     * and bind the ApiExternalServiceAction with the T?
+     */
+    val response: ResponseParam,
```
arcuri82, line 16: likely yes, but we can refactor it when we ll need it

core/src/main/kotlin/org/evomaster/core/problem/external/service/param/ResponseParam.kt
```diff
@@ -2,23 +2,19 @@ package org.evomaster.core.problem.external.service.param
...
+ * @property responseType the data type of the response
+ * @property response represents the response to return
+ * @property extraProperties represent the extra properties to describe the responses, such as status code for HTTP response,
+ * note that all [extraProperties] are considered as part of children of this param.
+ */
+abstract class ResponseParam(
+    name: String,
+    val responseType: EnumGene<String>,
+    val response: OptionalGene,
+    val extraProperties: List<Gene>
```
arcuri82, line 19: hmmm unsure about this... shouldn't they be explicitely marked here, eg a gene for status code. otherwise, how are u going to infer their phenotype when computing fitness and generating JUnit tests?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/external/service/param/ResponseParam.kt
ZhangMan1126: this is the general response for any API. the HTTP response is identified with `HttpWsResponseParam.kt`

core/src/main/kotlin/org/evomaster/core/problem/external/service/param/ResponseParam.kt
```diff
@@ -2,23 +2,19 @@ package org.evomaster.core.problem.external.service.param
...
+) : Param(name, extraProperties.plus(response).toMutableList())
```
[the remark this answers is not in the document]
ZhangMan1126: thanks. added it with the latest commit.

core/src/main/kotlin/org/evomaster/core/problem/external/service/param/ResponseParam.kt
arcuri82: i m still unsure. let's discuss it next time we meet

PR #584, opened 2022-09-07
Gql scalar fields

asmab89:


arcuri82:
@asmab89 is this branch ready for review?

PR #593, opened 2022-09-13
gql_compose

asmab89:


e2e-tests/spring-graphql/src/main/kotlin/com/foo/graphql/allLimitReached/DataRepository.kt
```diff
@@ -0,0 +1,52 @@
+package com.foo.graphql.allLimitReached
```
arcuri82, line 1: should avoid upper-case letter in the name of packages. see:
https://github.com/EMResearch/EvoMaster/blob/master/docs/for_developers.md#naming-convention

arcuri82 (changes requested):
recall to pull from `master`

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/GraphQLActionBuilder.kt
```diff
@@ -130,9 +130,23 @@ object GraphQLActionBuilder {
...
+        if (params.any { p -> p is GQReturnParam }) {
+
+            if (!params.find { p -> p is GQReturnParam }?.let { isAllLimitInObjectFields(it) }!!  ) {
+                val action = GraphQLAction(actionId, element.fieldName, type, params)
+                actionCluster[action.getName()] = action
+            }
+
+        }else {
+            val action = GraphQLAction(actionId, element.fieldName, type, params)
+            actionCluster[action.getName()] = action
```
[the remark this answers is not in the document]
asmab89: Done

e2e-tests/spring-graphql/src/main/kotlin/com/foo/graphql/allLimitReached/DataRepository.kt
asmab89: Done

arcuri82:
hi @asmab89  did u see the comment `recall to pull from master` ? (should fix the issue with E2E failing in JDK 8)

PR #596, opened 2022-09-19
save SQL group when writing extra heuristics statistics 

jgaleotti:


core/src/main/kotlin/org/evomaster/core/database/ReplaceValuesDeParser.kt
```diff
@@ -0,0 +1,46 @@
+package org.evomaster.core.database
+
+import net.sf.jsqlparser.expression.*
+import net.sf.jsqlparser.util.deparser.ExpressionDeParser
+
+
+class ReplaceValuesDeParser() :
```
arcuri82, line 7: add some comment about what this class is for

core/src/main/kotlin/org/evomaster/core/search/service/ExtraHeuristicsLogger.kt
```diff
@@ -21,45 +28,61 @@ class ExtraHeuristicsLogger {
...
+        Files.createFile(filePath)
+    }
 
+    private fun getExtraHeuristicsFilePath() = Paths.get(config.extraHeuristicsFile).toAbsolutePath()
 
     fun writeHeuristics(list: List<HeuristicEntryDto>, actionId: Int) {
-        if (!config.writeExtraHeuristicsFile) {
-            return
-        }
+        if (config.writeExtraHeuristicsFile) {
```
arcuri82, line 50: revert. usually it is better to return immediately on a negated condition, as it avoid a long nested block

core/src/main/kotlin/org/evomaster/core/search/service/ExtraHeuristicsLogger.kt
```diff
@@ -21,45 +28,61 @@ class ExtraHeuristicsLogger {
...
 
-        val counter = time.evaluatedIndividuals
+            val counter = time.evaluatedIndividuals
+            val lines = list.map {
+                val group = if (it.type == HeuristicEntryDto.Type.SQL) {
+                    try {
+                        "\"${removeAllConstants(it.id)}\""
```
arcuri82, line 56: as this is going to be put in a CSV file, we need to make sure that we properly handle any `,` character

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/database/ReplaceValuesDeParser.kt
jgaleotti: ok

core/src/main/kotlin/org/evomaster/core/search/service/ExtraHeuristicsLogger.kt
jgaleotti: ok.

core/src/main/kotlin/org/evomaster/core/search/service/ExtraHeuristicsLogger.kt
jgaleotti: One question, what about "org.evomaster.client.java.controller.api.dto.HeuristicEntryDto#id" 
How the "," character is handled here? 
It looks as the same problem concerns both it.id and  removeAllConstants(it.id), isn't it?

arcuri82: hi. yes, technically it is the same issue. maybe we haven't encountered it before. anyway, if issue happens, we ll see problem as soon as we try to load the generated CSV file, due to column mismatches.
should check how `,` commas are escaped in CSV

PR #598, opened 2022-09-20
local id for structural element

ZhangMan1126:
local id for the structural element has been added.
after the wiremock PR is merged, I will add a check of local id for genes.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/Individual.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/Individual.kt
@@ -51,7 +51,7 @@ abstract class Individual(override var trackOperator: TrackOperator? = null,
 
 
     override fun preChildrenSetup(c : Collection<StructuralElement>) {
-        if (isLocalIdsNotAssigned())
+        if (isLocalIdsNotInitialized())
             setLocalIdsForChildren((c as List<ActionComponent>).flatMap { it.flatView() })
     }
 
@@ -440,7 +440,10 @@ abstract class Individual(override var trackOperator: TrackOperator? = null,
                 && flatView().run { this.map { it.getLocalId() }.toSet().size == this.size }
     }
 
-    private fun isLocalIdsNotAssigned() : Boolean{
+    /**
+     * @return if local ids are not initialized
+     */
+    private fun isLocalIdsNotInitialized() : Boolean{
         return flatView().all { !it.hasLocalId ()
                 // FIXME Man need to check local id for gene for the moment
                 //&& (it !is Action || it.seeTopGenes().all { g-> g.flatView().all { i-> !i.hasLocalId() }})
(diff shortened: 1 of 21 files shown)


core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -441,7 +441,10 @@ abstract class Individual(override var trackOperator: TrackOperator? = null,
     }
 
     private fun isLocalIdsNotAssigned() : Boolean{
-        return flatView().all { it.getLocalId() == ActionComponent.NONE_ACTION_COMPONENT_ID}
+        return flatView().all { !it.hasLocalId ()
```
arcuri82, line 444: instead of `all`, shouldn't it be `any`?

arcuri82: also, isn't this just `! areAllLocalIdsAssigned()` ?

core/src/main/kotlin/org/evomaster/core/search/StructuralElement.kt
```diff
@@ -34,8 +35,44 @@ abstract class StructuralElement (
...
+    fun resetLocalId() {
+        localId = NONE_LOCAL_ID
+    }
+
+    fun getLocalId() = localId
+
+    /**
+     * a unique id is used to identify this structural element in the context of an individual
+     */
+    private var localId : String = NONE_LOCAL_ID
```
arcuri82, line 74: internal fields should not be in the middle of the file... move it up before any methods, but after companion object

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
ZhangMan1126: it should be `all`.
the local id will be assigned only if none of its children has local id.

in this case, it would not be `! areAllLocalIdsAssigned()`

arcuri82: so, if 1 out 10 has `localID`, then the function returns `true`, which could be interpreted as the localIds are initialized.
if the logic must be with `all`, maybe a better name is `areAllLocalIdsNotInitialized()`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
ZhangMan1126: I will change the name.

if the structural element has 10 children, only one of them has local id. it should be a bug. All of its children should be initialized at the same time.
currently, if the one out of 10 children has local id, then the rest of the children will not be assigned with the local id, thus the problem would be identified when checking `validLocalId`.

PR #599, opened 2022-09-21
Gql contentful bb

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/StateBuilder.kt
```diff
@@ -18,6 +19,11 @@ object StateBuilder {
         "__Field", "__InputValue", "__Type", "__TypeKind"
     )
 
+    /**
+     * Used to create unique IDs
+     */
+    private val idGenerator = AtomicInteger()
```
arcuri82, line 25: this must be avoided, ie mutable static state. 
if really need it as static state, rather use `StaticCounter`

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/TempState.kt
```diff
@@ -25,5 +25,7 @@ data class TempState(
         /*
          * An intermediate data structure used for extracting Union types
          */
-        var tempUnionTables: MutableList<Table> = mutableListOf()
+        var tempUnionTables: MutableList<Table> = mutableListOf(),
+
+        var inputTypeName: MutableMap<String,String> = mutableMapOf()
```
arcuri82, line 30: what are the `key` and `value` here in this map? add documentation

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/StateBuilder.kt
asmab89: Done.

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/TempState.kt
asmab89: Done.

PR #613, opened 2022-10-04
Gene local id

ZhangMan1126:
handle local id for gene
update collection of gene impact based on the local id

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/Individual.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/Individual.kt
@@ -116,7 +116,8 @@ abstract class Individual(override var trackOperator: TrackOperator? = null,
     }
 
     fun doGlobalInitialize(searchGlobalState : SearchGlobalState){
-        doInitializeLocalId()
+        if (!areAllLocalIdsAssigned())
+            doInitializeLocalId()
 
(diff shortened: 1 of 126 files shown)


core/src/main/kotlin/org/evomaster/core/search/Individual.kt
```diff
@@ -162,9 +175,11 @@ abstract class Individual(override var trackOperator: TrackOperator? = null,
 
 
     open fun doInitialize(randomness: Randomness? = null){
-
-        seeAllActions()
-            .forEach { it.doInitialize(randomness) }
+        if (!areAllLocalIdsAssigned())
+            doInitializeLocalId()
+        if (!areAllGeneInitialized())
```
arcuri82, line 180: not sure about this one... the point was to fail if we try to initialize something that is already initialized. when building an individual, shouldn't have actions/genes that are already initialized

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/Individual.kt
ZhangMan1126: For `gene` and `action`, they can be randomized without an individual, 
but their local id can only be initialized once they are added to the individual.

PR #621, opened 2022-10-11
Socket connect

ZhangMan1126:
handle `socket.connect`


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/ReplacementList.java
```diff
@@ -57,7 +57,8 @@ public static List<MethodReplacementClass> getList() {
                     new WebRequestClassReplacement(),
                     new URIClassReplacement(),
                     new URLClassReplacement(),
-                    new UUIDClassReplacement()
+                    new UUIDClassReplacement(),
+                    new SocketClassReplacement()
```
arcuri82, line 61: should be kept in alphabetic order

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/InetAddressClassReplacement.java
```diff
@@ -0,0 +1,66 @@
...
+import org.evomaster.client.java.instrumentation.coverage.methodreplacement.MethodReplacementClass;
+import org.evomaster.client.java.instrumentation.coverage.methodreplacement.Replacement;
+import org.evomaster.client.java.instrumentation.coverage.methodreplacement.UsageFilter;
+import org.evomaster.client.java.instrumentation.shared.ReplacementCategory;
+import org.evomaster.client.java.instrumentation.shared.ReplacementType;
+import org.evomaster.client.java.instrumentation.staticstate.ExecutionTracer;
+
+import java.net.*;
+
+public class InetAddressClassReplacement implements MethodReplacementClass {
```
arcuri82, line 13: add comment that this class is not in used, eg not in `ReplacementList`, but kept if in future we want to mock it

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/SocketClassReplacement.java
```diff
@@ -0,0 +1,48 @@
...
+public class SocketClassReplacement implements MethodReplacementClass {
+    @Override
+    public Class<?> getTargetClass() {
+        return Socket.class;
+    }
+
+    @Replacement(
+            type = ReplacementType.TRACKER,
+            category = ReplacementCategory.NET,
+            id = "Socket_connect_Replacement",
```
arcuri82, line 23: `id` only needed for third-party libraries

arcuri82 (changes requested):

PR #622, opened 2022-10-12
Rpc ex action

ZhangMan1126:
rpc seeded external service action support


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/UnitsInfoDto.java
```diff
@@ -61,4 +61,12 @@ public class UnitsInfoDto {
...
      */
     public Map<String,String> parsedDtos;
+
+
+    /*
+       This is to save dto schema for specified dto
+       Key -> DTO full name
+       Value -> OpenAPI object schema
+    */
+    public Map<String, String> extractedSpecifiedDtos;
```
arcuri82, line 71: what is a `specified` dto?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
```diff
@@ -72,17 +79,26 @@ public static void registerSchemaIfNeeded(Class<?> valueType) {
...
+                    embedded.forEach(ClassToSchema::registerSchemaIfNeeded);
+                }
 
+            }
+        }catch (Exception e){
+            SimpleLogger.warn("Fail to get schema for Class:"+valueType.getName(), e);
+            /*
+                fail with tests
+             */
+            assert(false);
```
arcuri82, line 99: `assert(false)` is fine in core, but should avoided in driver, as user might start driver with `-ea` flag

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/UnitsInfoDto.java
ZhangMan1126: add doc to clarify this

arcuri82 (approved):

PR #624, opened 2022-10-16
Okhttp handle

ZhangMan1126:


e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/wiremock/base/WireMockManualTest.java
```diff
@@ -22,6 +23,8 @@ public class WireMockManualTest extends SpringTestBase {
 
     @BeforeAll
     public static void initClass() throws Exception {
+        CIUtils.skipIfOnGA();
```
arcuri82, line 26: hi. in all these skips, add a TODO to put it back once library is updated, with link to https://github.com/alibaba/java-dns-cache-manipulator/issues/115

PR #626, opened 2022-10-19
Fix class toschema

ZhangMan1126:
fix StackOverflowError with cycle object

### DIFF
--- a/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
+++ b/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
@@ -49,6 +49,8 @@ public class ClassToSchema {
      */
     private static final Map<Type, String> cache = new ConcurrentHashMap<>();
 
+    private static final Map<Type, String> refCache = new ConcurrentHashMap<>();
+
     private static final String fieldRefPrefix = "{\"$ref\":\"";
 
     private static final String fieldRefPostfix = "\"}";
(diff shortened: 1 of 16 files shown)


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
```diff
@@ -64,38 +66,89 @@ public static void registerSchemaIfNeeded(Class<?> valueType) {
...
+            }
 
+        }
+    }
+    /**
+     *
+     * @return a schema representation of the class along with schemas of its ref classes in the form
+     *      "name: { "type name": "schema", "type name": "schema", .... }"
+     */
+    public static String getOrDeriveSchemaWithItsRef(Class<?> klass){
```
arcuri82, line 85: isn't the name confusing? you are always re-deriving a string for the same input `klass`. maybe should create a cache with `klass` as key?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
```diff
@@ -64,38 +66,89 @@ public static void registerSchemaIfNeeded(Class<?> valueType) {
...
+     */
+    public static String getOrDeriveNonNestedSchema(Class<?> klass) {
+        return getOrDeriveSchema(klass.getName(), klass, false, Collections.emptyList());
+    }
 
-        if (cache.containsKey(type)) {
+    private static String getOrDeriveSchema(String name, Type type, Boolean useRefObject, List<Class<?>> nested) {
+
+        if (cache.containsKey(type) && !useRefObject) {
             return named(name, cache.get(type));
```
arcuri82, line 119: on `core` it is no problem, but on driver side need to be very carefull of performance, especially for string concatenations. might want to add cache to avoid re-computing `named()` each time

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
ZhangMan1126: add more doc to explain this method

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
ZhangMan1126: here, `cache` only saves schema with its properties.

if the `useRefObject` is true, it means that a schema for the object  will be a reference schema (such as. "$ref":"#/components/schemas/org.evomaster.client.java.instrumentation.object.dtos.CycleDtoA").

arcuri82: hi. maybe easier to discuss what I meant in a call. remind me tomorrow morning

ZhangMan1126: hi, thanks. update it a bit.

PR #630, opened 2022-10-24
Gql nullability

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/GraphQLActionBuilder.kt
```diff
@@ -376,31 +379,31 @@ object GraphQLActionBuilder {
 
             "int" ->
                 if (element.isKindOfFieldTypeOptional)
-                    OptionalGene(element.fieldName, IntegerGene(element.fieldName))
+                    NullableGene(element.fieldName,OptionalGene(element.fieldName, IntegerGene(element.fieldName)))
```
arcuri82, line 382: this looks weird... an optional gene inside a nullable one? i would have expected the other way round: ie. something can be optional, and, if there, it could be null

arcuri82: note this applies to all cases here

core/src/main/kotlin/org/evomaster/core/search/gene/collection/TupleGene.kt
```diff
@@ -170,7 +170,7 @@ class TupleGene(
         val i = it.getValueAsRawString()
         "${it.name} : $i"
     } else {
-        if (it is ObjectGene || (it is OptionalGene && it.gene is ObjectGene)) {
+        if (it is ObjectGene || (it.getWrappedGene(OptionalGene::class.java)?.gene  is ObjectGene)) {
```
arcuri82, line 173: this seems redundant. could directly check for `it.getWrappedGene(ObjectGene::class.java) != null`

e2e-tests/spring-graphql/src/test/kotlin/org/evomaster/e2etests/spring/graphql/nullableInput/GQLNullableInputEMTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        ) { args: MutableList<String> ->
+
+            args.add("--problemType")
+            args.add(EMConfig.ProblemType.GRAPHQL.toString())
+
+            val solution = initAndRun(args)
+
+            Assertions.assertTrue(solution.individuals.size >= 1)
+            assertHasAtLeastOneResponseWithData(solution)
+            assertNoneWithErrors(solution)
```
arcuri82, line 36: should add assertions, to make sure all queries are called at least once

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/GraphQLActionBuilder.kt
asmab89: Done

PR #643, opened 2022-11-04
Gql split

bogdanmarculescu:
Splitting into different files (for success, faults, other) for GraphQL test cases.

In this case it is just "Success" - test case runs and "data" is present in response, or "Fault" - test case response is empty or there is a field "errors" in the response.


arcuri82:
@bogdanmarculescu  hi. looks like CI is failing. please look into it

bogdanmarculescu:
Looking into it now. I think this was a previously hidden bug related to the executive summary.

PR #645, opened 2022-11-07
set gene based on seeded value

ZhangMan1126:
for task: handle cycleObject if it exceeds the depth 2


arcuri82:
@man-zhang  hi. what about this old PR? is it still relevant?

PR #654, opened 2022-11-24
Additional properties

ZhangMan1126:
for task

Fix ObjectGene and MapGene, especially in regards of supporting additionalProperties


ZhangMan1126:
Github seems to count `Files changed` based on the master when you created PR. To avoid showing the changes made with `all-api-guru`,  I close the PR then re-create it again.

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -596,7 +596,23 @@ class Main {
                     }
                 }
 
-            }else {
+            }else if (config.problemType == EMConfig.ProblemType.GRAPHQL) {
```
arcuri82, line 599: why this? is it some merge issue? have u pulled latest from master?

core/src/main/kotlin/org/evomaster/core/search/gene/FlexibleObjectGene.kt
```diff
@@ -0,0 +1,171 @@
...
+import org.evomaster.core.search.gene.utils.GeneUtils.isInactiveOptionalGene
+import org.evomaster.core.search.service.AdaptiveParameterControl
+import org.evomaster.core.search.service.Randomness
+import org.evomaster.core.search.service.mutator.MutationWeightControl
+import org.evomaster.core.search.service.mutator.genemutation.AdditionalGeneMutationInfo
+import org.evomaster.core.search.service.mutator.genemutation.SubsetGeneMutationSelectionStrategy
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class FlexibleObjectGene<T>(
```
arcuri82, line 19: add some documentation for this gene.

also, i guess there is quite a bit of code that would be similar to `ObjectGene`.

Wondering if having a new gene that internally contains an `ObjectGene` (for the fixed fields) and a `MapGene` (for the additional properties) would be cleaner. Or simply adding a `MapGene` directly inside `ObjectGene` as optional.

Wondering about all places in the code in which we have checks like `is ObjectGene` which will not work for this new gene

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/Main.kt
ZhangMan1126: have pulled the latest one now.

core/src/main/kotlin/org/evomaster/core/search/gene/FlexibleObjectGene.kt
ZhangMan1126: the reason why I did not extend it from `ObjectGene` is that `ObjectGene` is _fixed_ composite gene.
We might need to extract `ObjectGene` as `interface`, then further distinguish `FixedObjextGene` and `FlexiableObjectGene`. 
However, for `FixedObjextGene` and `ObjectGene`, most of methods (eg, methods related to copy, and mutation) are different.

`FlexiableObjectGene` was not defined with `ObjectGene` and `MapGene` because of its structure of children. currently, the `children` is a list of fields  (fixed fields plus additional ones). the structure of children might relate to mutation. 

another solution (which might not be clean) is that we extend `FixedObjextGene` from `ObjectGene`, and make it have fixed two fields, i.e., _fixedFields:`ObjectGene`_ and _additionalFields:`MapGene`_.  thus, we could identify `FixedObjextGene` as `FixedCompositegene`. but in this case, we make the structure of the gene a bit complex.

ZhangMan1126:
@arcuri82 there exists Java heap size problem for `base` (see commit 7a22986 and log https://github.com/EMResearch/EvoMaster/actions/runs/3542270995/jobs/5947565225)
I tried to increase xmx to 8192, then it failed in both `base` and `latest`. 8G  might exceed the memory we have in Github Action.

arcuri82:
@ZhangMan1126  when running on my machine, memory for the test (`GeneTest`) goes up to 2.5G, which is lower than 4G we have for surefire tests. re-running the build now to see if it was just a flaky failure

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -20,36 +26,183 @@ import org.slf4j.LoggerFactory
...
+            }
+        }
+
+        val remove = randomness.choose(additionalFields!!)
+        remove.removeThisFromItsBindingGenes()
+        killChild(remove)
+        return true
+    }
+
+    private fun addElement(randomness: Randomness) : PairGene<StringGene, FlexibleGene>?{
```
arcuri82, line 161: why this is method is called `add`? isn't just returning a copy, and not adding it?

core/src/main/kotlin/org/evomaster/core/search/gene/root/CompositeConditionalFixedGene.kt
```diff
@@ -0,0 +1,65 @@
+package org.evomaster.core.search.gene.root
+
+import org.evomaster.core.search.StructuralElement
+import org.evomaster.core.search.gene.Gene
+
+abstract class CompositeConditionalFixedGene(
```
arcuri82, line 6: add some javadoc for this class

arcuri82 (changes requested):

PR #656, opened 2022-11-29
Handling optional tuple.

asmab89:


arcuri82 (changes requested):
@asmab89  should have some test cases (either unit or E2E) that cover these cases

PR #657, opened 2022-11-30
option to skip classes to instrument

arcuri82:


arcuri82:
@ZhangMan1126  I didn't test it... can you check it out if it works for what you need? i m not going to do further modifications to this branch, so u can fix things directly if needed

PR #690, opened 2023-01-17
Gql names issues

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/GraphQLActionBuilder.kt
```diff
@@ -153,10 +153,18 @@ object GraphQLActionBuilder {
...
+
     ) {
-        val action = GraphQLAction(actionId, element.fieldName, type, params)
-        actionCluster[action.getName()] = action
+        if (state.inputTypeName[element.fieldName]?.isNotEmpty() == true) {
+            val action = GraphQLAction(actionId, state.inputTypeName[element.fieldName].toString(), type, params)
+            actionCluster[action.getName()] = action
+        } else {
+            val action = GraphQLAction(actionId, element.fieldName, type, params)
+            actionCluster[action.getName()] = action
```
arcuri82, line 165: as `actionCluster[action.getName()] = action` is the same in both branches, shouldn't it be moved after the `if`?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/builder/GraphQLActionBuilder.kt
asmab89: Done.

PR #712, opened 2023-02-14
Thread pool for harvesting external requests

seran:


core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/HarvestActualHttpWsResponseHandler.kt
```diff
@@ -60,22 +61,20 @@ class HarvestActualHttpWsResponseHandler {
...
 
 
     /**
-     * TODO
-     * to further improve the efficiency of collecting actual responses,
-     * might have a pool of threads to send requests
+     * TODO: Add EMConfig option to set value as config
      */
-    private lateinit var threadToHandleRequest: Thread
+    private var workerPool = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors())
```
ZhangMan1126, line 70: it might be too many if we run tasks in parallel.
how about `min(3, Runtime.getRuntime().avaiableProcessors())` in case a less powerful laptop?

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/HarvestActualHttpWsResponseHandler.kt
```diff
@@ -146,49 +145,45 @@ class HarvestActualHttpWsResponseHandler {
...
-    fun shutdown(){
+    fun shutdown() {
         Lazy.assert { config.doHarvestActualResponse() }
-        synchronized(lock){
-            if (threadToHandleRequest.isAlive)
-                threadToHandleRequest.interrupt()
+        synchronized(lock) {
             httpWsClient.close()
         }
+        workerPool.shutdown()
```
ZhangMan1126, line 163: shall we destroy the thread pool before closing the client?

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/HarvestActualHttpWsResponseHandler.kt
```diff
@@ -220,51 +218,54 @@ class HarvestActualHttpWsResponseHandler {
...
+        synchronized(lock) {
             val newRequests = notInCollected.filterNot { queue.contains(it) }
             if (newRequests.isEmpty())
                 return
             updateExtractedObjectDto()
             lock.notify()
             queue.addAll(newRequests)
+
+            val task = Runnable { sendRequestToRealExternalService() }
+            workerPool.execute(task)
```
ZhangMan1126, line 264: it seems directly sending requests based on a sequence of the requests to be sent by SUT (ie, send the request after each addition).
then the queue seems less useful that might be problematic if we want to optimize them.

we might need to discuss it tomorrow.

ZhangMan1126 (changes requested):
Thanks, Seran. there exist some parts which I am not quite sure. we might need to discuss them tomorrow.

ZhangMan1126:
please resolve the conflicts as well :)

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/HarvestActualHttpWsResponseHandler.kt
seran: `Runtime.getRuntime().avaiableProcessors()` will return number of processor available for the JVM. However, I changed it to as suggested.

ZhangMan1126 (changes requested):
create `service` package under `org.evomaster.core.problem.externalservice.httpws`
then move
`HttpWsExternalServiceHandler` and `HarvestActualHttpWsResponseHandler` under  `org.evomaster.core.problem.externalservice.httpws.service`

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/HarvestActualHttpWsResponseHandler.kt
```diff
@@ -113,11 +106,6 @@ class HarvestActualHttpWsResponseHandler {
      */
     private val lock = Object()
```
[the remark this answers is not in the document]
seran: forget to remove it

arcuri82 (changes requested):

PR #715, opened 2023-02-21
handle_comparison_with_dates

jgaleotti:
SQL do handle: 'MinorThanEquals' and other comparisons with dates.
It still does not support subqueries.


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/HeuristicsCalculator.java
```diff
@@ -490,11 +499,35 @@ private Object getValue(Expression exp, DataRow data) {
...
+            str = removeQuotes(str);
+            return str;
         } else {
             cannotHandle(exp);
             return null;
         }
     }
+
+    /**
+     * Given a string "Hello World" returns "ello Worl".
```
arcuri82, line 517: this description is confusing... so it is not `removingQuotes` but always removing first and last char? if so, should change name of function (eg, `removeFirstAndLast`), or change semantics (ie, throw an execption if first and last are not quotes). also should it deal only with `"` or also `'`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/HeuristicsCalculator.java
```diff
@@ -490,11 +499,35 @@ private Object getValue(Expression exp, DataRow data) {
...
+     * @return
+     */
+    private String removeQuotes(String str) {
+        if (str.length() < 2) {
+            throw new IllegalArgumentException("Cannot remove quotes from " + str);
+        }
+        return str.substring(1, str.length() - 1);
+    }
+
+    private boolean startsAndEndsWithQuotes(String str) {
```
arcuri82, line 530: just to clarify, does SQL treat `"` and `'` the same or differently?

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/HeuristicsCalculator.java
jgaleotti: Changed to removeFirstAndLastCharacter()

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/HeuristicsCalculator.java
jgaleotti: yes. I added a test case to reflect that.

PR #719, opened 2023-03-07
External request harvester EMConfig for number of threads

seran:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1812,6 +1812,10 @@ class EMConfig {
     @Probability(activating = true)
     var probOfMutatingResponsesBasedOnActualResponse = 0.0
 
+    @Cfg("Number of threads for external request harvester")
+    @Experimental
```
arcuri82, line 1816: should add something like: "No more threads than numbers of processors will be used."

also, add constraints, ie, `@Min(1.0)`

arcuri82 (changes requested):

PR #722, opened 2023-03-10
HTTP POST allowed in HarvestActualHttpWsResponseHandler

seran:


man-zhang (commented):
I only have a comment for adding a bit of clarification.
after you fix this, you could directly ask for Andrea's review.

PR #723, opened 2023-03-13
Support MongoDB

hghianni:
- Monitoring of Find executions.
- Calculation of heuristic distance for queries.
- E2E tests.


arcuri82:
@hghianni hi. thanks! 
unfortunately this a super-busy week :( I ll look at this PR hopefully by next week

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1026,6 +1026,9 @@ class EMConfig {
     @Cfg("Tracking of SQL commands to improve test generation")
     var heuristicsForSQL = true
 
+    @Cfg("Tracking of Mongo commands to improve test generation")
+    var heuristicsForMongo = true
```
arcuri82, line 1030: it should be marked as `@Experimental`, with `false` as default value. only once stable we can put it on by default

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1151,6 +1154,11 @@ class EMConfig {
             " on the JVM.")
     var instrumentMR_EXT_0 = true
 
+    @Cfg("Execute instrumentation for method replace with category MONGO." +
+            " Note: this applies only for languages in which instrumentation is applied at runtime, like Java/Kotlin" +
+            " on the JVM.")
+    var instrumentMR_MONGO = true
```
arcuri82, line 1160: it should be marked as `@Experimental`, with `false` as default value. only once stable we can put it on by default

core/src/main/kotlin/org/evomaster/core/problem/enterprise/service/EnterpriseFitness.kt
```diff
@@ -211,5 +211,28 @@ abstract class EnterpriseFitness<T> : FitnessFunction<T>() where T : Individual
...
+    private fun handleMongoHeuristics(dto: TestResultsDto, fv: FitnessValue) {
+        if (configuration.heuristicsForMongo) {
+
+            for (i in 0 until dto.extraHeuristics.size) {
+
+                val extra = dto.extraHeuristics[i]
+
+                extraHeuristicsLogger.writeHeuristics(extra.heuristics, i)
+
+                val toMinimize = extra.mongoHeuristics
```
arcuri82, line 227: likely this will need update after removing `mongoHeuristics`

e2e-tests/spring-rest-mongo/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-mongodb</artifactId>
+            <version>2.5.4</version>
```
arcuri82, line 48: all these repeated version numbers should be moved to root pom file, to avoid duplication, and make update to new version easier (as needs to be modified only in 1 place)

e2e-tests/spring-rest-mongo/src/test/java/org/evomaster/e2etests/spring/rest/mongo/foo/MongoEMFitnessTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void testRunEM() throws Throwable {
+
+        runTestHandlingFlakyAndCompilation(
+                "MongoEMFitness",
+                "org.foo.spring.rest.mongo.MongoEMFitness",
+                10000,
+                (args) -> {
+                    args.add("--enableWeightBasedMutationRateSelectionForGene");
+                    args.add("false");
```
arcuri82, line 32: will need to add the experimental parameters that will need to be put to `false` by default in `EMConfig`

arcuri82 (changes requested):

arcuri82:
@hghianni @jgaleotti  many thanks for the PR! ;) see my comments

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/ExtraHeuristicsDto.java
```diff
@@ -19,4 +19,6 @@ public class ExtraHeuristicsDto {
     public List<HeuristicEntryDto> heuristics = new ArrayList<>();
 
     public ExecutionDto databaseExecutionDto;
+
+    public List<HeuristicEntryDto> mongoHeuristics = new ArrayList<>();
```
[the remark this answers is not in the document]
hghianni: I think i'm missing something. I understand it's cleaner to have all heuristics in the same place but I don't get why would it be necessary to update all search algorithms.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/MongoHandler.java
```diff
@@ -0,0 +1,96 @@
...
+     * The heuristics based on the Mongo execution
+     */
+
+    private final List<MongoOperationDistance> distances;
+
+    private volatile boolean calculateHeuristics;
+
+    public MongoHandler() {
+        distances = new ArrayList<>();
+        operations = new CopyOnWriteArrayList<>();
```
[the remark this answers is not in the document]
hghianni: No particular reason. I'll change it.

hghianni:
Hi @arcuri82, thanks for your comments. I left some of my own related to some doubts I have. The ones marked as solved are the ones that I fully agree with and have already corrected in the code. Thank you!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/ExtraHeuristicsDto.java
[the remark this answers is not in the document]
arcuri82: right-click on `heuristics`, and select `Find Usages`. you ll see all the places it is used, like `EnterpriseFitness`. if you introduce a new variable, then you need to update/extend all those places as well to have the same behavior

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.kt
```diff
@@ -0,0 +1,327 @@
...
+import org.bson.codecs.DecoderContext
+import org.bson.codecs.DocumentCodec
+import org.bson.conversions.Bson
+import org.evomaster.client.java.controller.mongo.operations.*
+import org.evomaster.client.java.controller.mongo.operations.synthetic.*
+import org.evomaster.client.java.instrumentation.coverage.methodreplacement.DistanceHelper
+import org.evomaster.core.mongo.QueryParser
+import kotlin.math.abs
+
+class MongoHeuristicsCalculator {
```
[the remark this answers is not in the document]
arcuri82: hi, sorry, no. that's a major architectural decision, with potentially a lot of side effects. 
should be in Java under client, with all code related to MongoDB accessed via reflection.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/selectors/AndSelector.kt
```diff
@@ -0,0 +1,19 @@
...
+class AndSelector : MultiConditionQuerySelector() {
+    override fun operator(): String = "\$and"
+
+    override fun parseConditions(value: Any?): QueryOperation? {
+        return when (value) {
+            is List<*> -> {
+                val filters = value.map { filter -> QueryParser().parse(filter as Document)}
+                AndOperation(filters)
+            }
+            else -> null
```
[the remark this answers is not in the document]
arcuri82: the function takes as input `value: Any?`, which is nullable. so, `null` is an accepted value, isn't it? what i mean is that `Any?` practically provides no info on what expected to be passed as input here. eg, why not `List<*>?`?

arcuri82:
hi. thx @hghianni . sorry for late reply, but Easter... anyway see my comments

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.kt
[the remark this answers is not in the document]
hghianni: All heuristics code converted to Java. Also using reflection to access code related to MongoDB

hghianni:
Hi @arcuri82 !, I think I fixed all the issues. Changes are in the last two commits. Thank you!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/ExtraHeuristicsDto.java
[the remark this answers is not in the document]
hghianni: Removed mongoHeuristics

client-java/controller/pom.xml
```diff
@@ -185,6 +185,15 @@
             <artifactId>libthrift</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mongodb</groupId>
+            <artifactId>bson</artifactId>
+            <version>4.2.3</version>
```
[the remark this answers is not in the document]
hghianni: Left the dependency but change scope to test

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
```diff
@@ -0,0 +1,136 @@
...
+            return decode.invoke(documentCodec, asBsonReader.invoke(bsonDocument), build.invoke(builderInstance));
+        } catch (NoSuchMethodException | InstantiationException | IllegalAccessException |
+                 InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static String getType(Object bsonType) {
+        try {
+            Class<?> bsonTypeClassMapClass = Class.forName("org.bson.codecs.BsonTypeClassMap");
```
arcuri82, line 83: as this is in the controller, might work fine. but, to be on safe side, rather use classloader from `bsonType`

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
```diff
@@ -0,0 +1,136 @@
...
+        } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException |
+                 InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static Object getTypeFromNumber(Integer number) {
+        Class<?> bsonTypeClass;
+        try {
+            bsonTypeClass = Class.forName("org.bson.BsonType");
```
arcuri82, line 98: hmm... so here we would have a reference to an object from MongoDB... so `Class.forName` should be fine (this applies for previous comments as well). however, something like this will not work under `instrumentation` module (again, long story, might be easier to explain in a meeting). amyway, you will see it immediately when you run with both embedded and external drivers

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
```diff
@@ -0,0 +1,136 @@
...
+            return valueOf.invoke(null, alias);
+        } catch (ClassNotFoundException | IllegalAccessException | InvocationTargetException |
+                 NoSuchMethodException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static Class<?> documentClass() {return getClass("org.bson.Document");}
+
+    private static Class<?> bsonDocumentClass() {return getClass("org.bson.BsonDocument");}
```
arcuri82, line 121: for performance reasons, these could all be stored directly in static variables, instead of being searched at each use. however, would need to make sure to do not crash the controller for SUTs that do not use MongoDB

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/MongoClassReplacement.java
```diff
@@ -0,0 +1,76 @@
...
+    private static Object handleFind(String id, Object mongoCollection, List<Object> args, Object query) {
+        long start = System.currentTimeMillis();
+        try {
+            Method findMethod = retrieveFindMethod(id, mongoCollection);
+            Object result = findMethod.invoke(mongoCollection, args.toArray());
+            long end = System.currentTimeMillis();
+            handleMongo(mongoCollection, query, true, end - start);
+            return result;
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            throw new RuntimeException(e);
```
arcuri82, line 64: this is changing the SUT by throwing a different exception. see other replacement methods. eg, can use somthing like:

```
try {
      ...
        } catch (IllegalAccessException e) {
            throw new RuntimeException(e);
        } catch (InvocationTargetException e) {
            throw e.getCause();
        }
```

arcuri82: otherwise you might break the behavior of the SUT if it has `catch` clauses for particular exceptions

e2e-tests/spring-rest-mongo/src/test/java/org/evomaster/e2etests/spring/rest/mongo/foo/MongoEMFitnessTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class
+
+
+MongoEMFitnessTest extends RestTestBase {
+
+    @BeforeAll
+    public static void initClass() throws Exception {
+        RestTestBase.initClass(new MongoPersonsAppController());
+    }
```
arcuri82, line 21: class loading can be very nasty... might need to do something like `ExternalServiceMockingFlakyEMTest`, eg:

```
 @BeforeAll
        ServiceController serviceController = new ServiceController();
        EMConfig config = new EMConfig();
        config.setInstrumentMR_NET(true);
        SpringTestBase.initClass(serviceController, config);
    }
```

but for `instrumentMR_MONGO`. this depends on when the SUT classes are loaded

arcuri82 (changes requested):

arcuri82:
@hghianni thx! see my comments
might also want to have a meeting together with @jgaleotti , so i can explain in more details some of the issues with class loading. also, i can show you how to easily add taint analysis here, but can be done after this PR is merged

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/MongoClassReplacement.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    private static Method retrieveFindMethod(String id, Object mongoCollection) {
+        return getOriginal(new MongoClassReplacement(), id, mongoCollection);
```
arcuri82, line 69: this `new MongoClassReplacement()` should had been a `singleton` object. otherwise it re-compute everything at each invocation

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/execution/MongoExecutionDto.java
```diff
@@ -0,0 +1,7 @@
+package org.evomaster.client.java.controller.api.dto.database.execution;
+
+
+public class MongoExecutionDto {
```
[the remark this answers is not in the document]
hghianni: Not now but it will be used when supporting insertions. I removed it.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -0,0 +1,411 @@
...
+                    .mapToDouble(value -> distanceToClosestElem((ArrayList<?>) actualValue, value))
+                    .min()
+                    .getAsDouble();
+        } else {
+            return distanceToClosestElem(expectedValues, actualValue);
+        }
+    }
+
+    private Double calculateDistanceForNotIn(NotInOperation<?> operation, Object doc) {
+        ArrayList<?> unexpectedValues = operation.getValues();
```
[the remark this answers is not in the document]
hghianni: Fixed this and similiar issues

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -0,0 +1,411 @@
...
+            return (double) ((Integer) val1 - (Integer) val2);
+        }
+
+        if (val1 instanceof String && val2 instanceof String) {
+            return (double) DistanceHelper.getLeftAlignmentDistance((String) val1, (String) val2);
+        }
+
+        if (val1 instanceof ArrayList<?> && val2 instanceof ArrayList<?>) {
+            // Modify
+            return 1.0;
```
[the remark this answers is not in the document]
hghianni: Done

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/MongoHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+        return min;
+    }
+
+    private static Iterable<?> getDocuments(Object collection) {
+        try {
+            Class<?> collectionClass = Class.forName("com.mongodb.client.MongoCollection");
+            return (Iterable<?>) collectionClass.getMethod("find").invoke(collection);
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException | ClassNotFoundException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
hghianni: Message added

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/MongoClassReplacement.java
hghianni: Fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/selectors/AndSelector.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * { $and: [ { <expression1> }, { <expression2> } , ... , { <expressionN> } ] }
+ */
+public class AndSelector extends MultiConditionQuerySelector {
+
+    @Override
+    protected QueryOperation parseConditions(ArrayList<?> value) {
+        ArrayList<QueryOperation> conditions = new ArrayList<>();
+        value.forEach(condition -> conditions.add(new QueryParser().parse(condition)));
```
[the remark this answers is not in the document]
hghianni: Fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
hghianni: Using  classloader now

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
hghianni: Done

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/classes/MongoClassReplacement.java
hghianni: Fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -0,0 +1,411 @@
...
+            InvertedTypeOperation op = (InvertedTypeOperation) operation;
+            return new TypeOperation(op.getFieldName(), op.getType());
+        }
+        return operation;
+    }
+
+    private <T1, T2> Double compareValues(T1 val1, T2 val2) {
+
+        if (val1 instanceof Double && val2 instanceof Double) {
+            return (Double) val1 - (Double) val2;
```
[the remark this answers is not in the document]
hghianni: Yes but it's expected. I use abs(compareValues(v1, v2)) when required.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/MongoHandler.java
```diff
@@ -0,0 +1,108 @@
...
+            double dist = calculator.computeExpression(info.getQuery(), doc);
+            if (dist == 0) return 0;
+            if (dist < min) min = dist;
+        }
+        return min;
+    }
+
+    private static Iterable<?> getDocuments(Object collection) {
+        try {
+            Class<?> collectionClass = Class.forName("com.mongodb.client.MongoCollection");
```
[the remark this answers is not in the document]
hghianni: I was able to generate tests using external controllers for both Mongo and non-Mongo systems. I believe that the reason it works when the dependencies are not loaded is because instrumentation does not occur, and therefore, the entire Mongo heuristic calculation is not used. Although we previously discussed adding a check to see if Mongo dependencies were loaded, I'm now not convinced that it is necessary.

e2e-tests/spring-rest-mongo/src/test/java/org/evomaster/e2etests/spring/rest/mongo/foo/MongoEMFitnessTest.java
hghianni: Both MongoFitness tests are working fine now so I don't understand if it is really necessary to change them.

arcuri82:
@hghianni  hi. did you do a `force-push` on a branch? why? force-push should usually be avoided (as it messes up the history, and can lose commits), unless special cases in which there is no alternative. can u explain what happened?  thx

PR #726, opened 2023-03-14
Tt extend

man-zhang:
improve the algorithm for SQL insertions.


man-zhang:
hello, @arcuri82 ,

With this PR, we modify the `copyValueFrom`, ie,
if the value violates the local validation, the `copyValueFrom` will not be actually performed as below. 

the test `TableConstraintEvaluatorTest` failed now due to such an update.
we might need JP's suggestion to fix the failed tests.

```java
override fun copyValueFrom(other: Gene) {
        if (other !is StringGene) {
            throw IllegalArgumentException("Invalid gene type ${other.javaClass}")
        }
        val current = this.value
        this.value = other.value

        // newly added check
        if (!isLocallyValid()){
            this.value = current
            return
        }
        this.selectedSpecialization = other.selectedSpecialization

        this.specializations.clear()
        this.specializations.addAll(other.specializations)

        killAllChildren()
        addChildren(other.specializationGenes.map { it.copy() })

        this.tainted = other.tainted

        this.bindingIds.clear()
        this.bindingIds.addAll(other.bindingIds)
    }override fun copyValueFrom(other: Gene) {
        if (other !is StringGene) {
            throw IllegalArgumentException("Invalid gene type ${other.javaClass}")
        }
        val current = this.value
        this.value = other.value

        if (!isLocallyValid()){
            this.value = current
            return
        }
        this.selectedSpecialization = other.selectedSpecialization

        this.specializations.clear()
        this.specializations.addAll(other.specializations)

        killAllChildren()
        addChildren(other.specializationGenes.map { it.copy() })

        this.tainted = other.tainted

        this.bindingIds.clear()
        this.bindingIds.addAll(other.bindingIds)
    }
```

arcuri82:
hi.
@man-zhang thx
@jgaleotti can you please have a look at these failing tests?

arcuri82:
@man-zhang also, we need to modify the signature of `copyValueFrom` to return a `Boolean`, specifying in the comments that it might return false if the copied value would lead to break local constraints in the current gene

jgaleotti:
sure

jgaleotti:
I pushed some changes to fix the test cases. But there are still some failings tests (org.evomaster.e2etests.spring.examples.db.crossfks.CrossFkEMTest.testDisableTaintSampleEM). Should I take a look into that one?

man-zhang:
> I pushed some changes to fix the test cases. But there are still some failings tests (org.evomaster.e2etests.spring.examples.db.crossfks.CrossFkEMTest.testDisableTaintSampleEM). Should I take a look into that one?

man-zhang:
thanks @jgaleotti I will handle the failing test `org.evomaster.e2etests.spring.examples.db.crossfks.CrossFkEMTest.testDisableTaintSampleEM`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -513,7 +513,7 @@ class EMConfig {
         }
     }
 
-    fun shouldGenerateSqlData() = isMIO() && (generateSqlDataWithDSE || generateSqlDataWithSearch)
+    fun shouldGenerateSqlData() = generateSqlDataWithDSE || generateSqlDataWithSearch
```
arcuri82, line 516: why removing the check `isMIO`?

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
```diff
@@ -148,6 +149,30 @@ class RestPath(path: String) {
...
+     * /root/{rootName}/foo/{fooName}/bar/{barName}
+     * /root/{rootName}/foo/{fooName}/bar are sibling
+     *
+     * note that in some cases,
+     *
+     * GET /root/{rootName}/foo/{fooName}/bar
+     * might need its sibling (eg, /root/{rootName}/foo/{fooName}/bar/{barName})
+     * to prepare resources
+     */
+    fun isSibling(other: RestPath): Boolean {
```
arcuri82, line 165: i m not sure `sibling` is the right word here. sibling would mean something like "at the same level", whereas here it looks more a parent-child relations

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
```diff
@@ -183,6 +208,26 @@ class RestPath(path: String) {
...
+
+    /**
+     * @return whether this is possible ancestor of [other]
+     *
+     * this is to handle the case eg.
+     * /root/{rootName}/bar/{barName} might be considered as
+     * the ancestor of /root/{rootName}/foo/{fooName}/bar/{barName}
+     *
+     */
+    fun isPossibleAncestorOf(other: RestPath): Boolean {
```
arcuri82, line 220: why the term `ancestor`? i find it confusing here

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
man-zhang: here, we consider that those two paths can prepare resources for each other, then use sibling 

POST /root/{rootName}/foo/{fooName}/bar/{barName}
GET /root/{rootName}/foo/{fooName}/bar

POST /root/{rootName}/foo/{fooName}/bar
GET /root/{rootName}/foo/{fooName}/bar/{barName}

the name of the method is modified and add more docs for clarifying this.

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
man-zhang: the doc and the name of the method are modified

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
man-zhang: thanks. this is a mistake due to the merge

PR #729, opened 2023-03-17
Gql enum string

asmab89:


core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
```diff
@@ -175,79 +174,62 @@ object GraphQLUtils {
...
 
-                    val i = gene.getValueAsRawString()
-                    printableInputGene.add("${gene.name} : $i")
-
-                }
+        }
+        return printableInputGene
+    }
 
+    fun inputsPrinting(
```
arcuri82, line 190: add docs stating what this function is supposed to do (or provide better function name)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/graphql/GraphQLUtils.kt
asmab89: Done

PR #730, opened 2023-03-20
Jackson ObjectMapper.readValue() method replacements

seran:
Skipped all the methods which uses `Jackson` classes as method arguments.


man-zhang (changes requested):
hi, Seran, thanks. I did not find tests that use the method replacement you introduced for Jackson. please add them (see an example `JacksonEMTest`).

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/JacksonObjectMapperClassReplacement.java
```diff
@@ -35,65 +33,118 @@ protected String getNameOfThirdPartyTargetClass() {
...
         String content = new BufferedReader(
                 new InputStreamReader(src, Charset.defaultCharset()))
                 .lines()
                 .collect(Collectors.joining(System.lineSeparator()));
 
-        JsonTaint.handlePossibleJsonTaint(content,valueType);
+        ClassToSchema.registerSchemaIfNeeded(typeClass);
+        JsonTaint.handlePossibleJsonTaint(content, typeClass);
 
         src = new ByteArrayInputStream(content.getBytes());
```
arcuri82, line 84: there is quite a bit of code here that is the same as previous function. should be possibly refactored in a private method to avoid code duplication

scripts/loopback-aliases-macos.sh
```diff
@@ -32,7 +32,7 @@ fi
 
 # Function to set aliases
 set_alias() {
-	for i in {2..254}
+	for i in {2..20}
```
arcuri82, line 35: why this change?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/JacksonObjectMapperClassReplacement.java
arcuri82: but, to save time, i ll do this refactoring.

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/JacksonObjectMapperClassReplacement.java
seran: Thank you.

scripts/loopback-aliases-macos.sh
seran: If I set till 254, internet stop working. I don't know why, so changed it for now.

arcuri82: in that case, add such comment

PR #738, opened 2023-03-31
Fix quantifier rx gene

man-zhang:


core/src/main/kotlin/org/evomaster/core/parser/GenePostgresLikeVisitor.kt
```diff
@@ -26,7 +26,10 @@ class GenePostgresLikeVisitor : PostgresLikeBaseVisitor<VisitResult>() {
         }
 
         val disjunction = DisjunctionRxGene("disj", terms, true, true)
-        val gene = RegexGene("LIKE", DisjunctionListRxGene(listOf(disjunction)))
+
+        val text = RegexUtils.getRegexExpByParserRuleContext(ctx)
+
+        val gene = RegexGene("LIKE$text", DisjunctionListRxGene(listOf(disjunction)))
```
arcuri82, line 32: for readability, i guess should put a space? ie, `LIKE $text`. this applies to all other cases here

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/parser/RegexUtils.kt
```diff
@@ -35,4 +36,12 @@ object RegexUtils {
                 .replace("\\E\\Q", "")
     }
 
+    fun getRegexExpByParserRuleContext(ctx : ParserRuleContext) : String{
+        return try {
+            ctx.text.run { ":$this" }
```
arcuri82, line 41: `:` is not part of the regex. so should not be returned here

arcuri82 (changes requested):

PR #739, opened 2023-03-31
Copy value from

man-zhang:


core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -187,25 +187,36 @@ class ObjectGene(
...
         if (other.isFixed != isFixed)
             throw IllegalArgumentException("cannot copy value for ObjectGene if their isFixed is different")
 
+        val current = copy()
+
+        var ok = true
+
         for (i in fixedFields.indices) {
-            this.fixedFields[i].copyValueFrom(other.fixedFields[i])
+            ok = ok && this.fixedFields[i].copyValueFrom(other.fixedFields[i])
```
arcuri82, line 203: can't this lead to a partial update? eg, the first N fields get copied, and then one is wrong, which leads to return `false` with a partial update?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -187,25 +187,36 @@ class ObjectGene(
...
+        }
+
+        if (!ok || !isLocallyValid()){
+            Lazy.assert { copyValueFrom(current) }
```
arcuri82, line 207: an modifying state operation inside an assertion? that would not apply at runtime outside the running of tests (which have `-ea` in maven and by default in IntelliJ).
or did u rather meant something like `val x = copyValueFrom(current);  assert(x);`?

core/src/main/kotlin/org/evomaster/core/search/gene/UUIDGene.kt
```diff
@@ -79,12 +80,19 @@ class UUIDGene(
...
         if (other !is UUIDGene) {
             throw IllegalArgumentException("Invalid gene type ${other.javaClass}")
         }
-        this.mostSigBits.copyValueFrom(other.mostSigBits)
-        this.leastSigBits.copyValueFrom(other.leastSigBits)
+        val current = copy()
+        val ok = this.mostSigBits.copyValueFrom(other.mostSigBits) && this.leastSigBits.copyValueFrom(other.leastSigBits)
+
+        if (!ok || !isLocallyValid()){
+            Lazy.assert { copyValueFrom(current) }
```
arcuri82, line 91: see previous comment about this. apply also for all other occurrences of this code.
however, as it seems there is a bit of duplicated code, should it be put in a common function?

core/src/main/kotlin/org/evomaster/core/search/gene/collection/PairGene.kt
```diff
@@ -60,12 +61,23 @@ class PairGene<F,S>(
...
-        second.copyValueFrom(other.second)
+        val firstCurrent = first.copy()
+        val ok1 = first.copyValueFrom(other.first)
+
+        if (!ok1){
+            return false
+        }
+        val ok2 = second.copyValueFrom(other.second)
+        if (!ok2){
+            Lazy.assert {first.copyValueFrom(firstCurrent) }
```
arcuri82, line 76: refactor all these `Lazy.assert`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
man-zhang: it might lead to the partial update, but it will be fixed with `copyValueFrom(current)` in the following code.

core/src/main/kotlin/org/evomaster/core/search/gene/UUIDGene.kt
man-zhang: thanks! modified now

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -939,6 +946,25 @@ abstract class Gene(
...
+     *
+     * @return if the value is updated with [updateValue]
+     */
+    fun updateValueOnlyIfValid(updateValue: () -> Boolean, undoIfUpdateFails: Boolean) : Boolean{
+        val current = copy()
+        val ok = updateValue()
+        if (!ok && !undoIfUpdateFails) return false
 
+        if (!ok || !isLocallyValid()){
+            assert(copyValueFrom(current))
```
arcuri82, line 963: rewrite with:
```
val copied = copyValueFrom(current)
assert(copied)
```
don't use state-change operations directly inside assertions

arcuri82 (changes requested):

PR #740, opened 2023-03-31
Ocvn fix

man-zhang:
- collect impacts based on specified types of targets
- fix the performance issue on ocvn


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -39,6 +40,11 @@ class EMConfig {
 
         private const val headerRegex = "(.+:.+)|(^$)"
 
+        private const val targetSeparator = ";"
+        private const val targetNone = "\\b(None|NONE|none)\\b"
+        private const val targetPrefix = "\\b(Class|CLASS|class|Line|LINE|line|Branch|BRANCH|branch|MethodReplacement|METHODREPLACEMENT|method[r|R]eplacement|Success_Call|SUCCESS_CALL|success_[c|C]all|Local|LOCAL|local|PotentialFault|POTENTIALFAULT|potential[f|F]ault)\\b"
```
arcuri82, line 45: hmmm, not sure about this one. if modify the names in `ObjectiveNaming`, this will break without any warning, isn't it?
also, to ignore case in a Java regex, can easily use `(?i)`, see [https://www.baeldung.com/java-case-insensitive-string-matching]()

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1829,10 +1843,15 @@ class EMConfig {
...
 
-    @Cfg("a probability of enabling single insertion strategy to insert rows into database.")
+
+    @Cfg("Specify prefixes of targets (e.g., MethodReplacement, Success_Call, Local) which will exclude in impact collection. " +
+            "Multiple exclusions should be separated with semicolon (i.e., ;).")
+    @Regex(targetExclusionRegex)
     @Experimental
-    @Probability(activating = true)
-    var probOfEnablingSingleInsertionForTable = 0.0
+    var excludeTargetsForImpactCollection = "Local;MethodReplacement"
```
arcuri82, line 1851: should be based on `ObjectiveNaming` constants

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
man-zhang: does eg, `cLAss` satisfy regex?

the test `testValidTargetExclusions` is used to check if there is any update existing in the `ObjectiveNaming`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
man-zhang: this now is updated

PR #748, opened 2023-04-18
Client thread local

man-zhang:
shared client map in order to make a client instance for each thread


man-zhang:
@arcuri82  I just found an uncommitted code which is only for doc update. should not have an impact on the GA results.

PR #757, opened 2023-04-26
Minimize

arcuri82:


arcuri82:
@man-zhang  you can see here some major changes (it is just easier to add you as reviewer, so you can see the differences in code)

PR #761, opened 2023-05-10
Check seeded tests

man-zhang:
exclude seeded tests that are not valid and provide warning info for the invalid ones.


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/RPCProblemDto.java
```diff
@@ -32,6 +33,6 @@ public class RPCProblemDto extends ProblemInfoDto{
     /**
      * a list of seeded RPC tests
      */
-    public List<List<RPCActionDto>> seededTestDtos;
+    public Map<String, List<RPCActionDto>> seededTestDtos;
```
arcuri82, line 36: what is the key here? add comment

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -590,7 +590,7 @@ public final void extractRPCSchema(){
      * parse seeded tests for RPC
      * @return a list of tests, and each test is a list of RCPActionDto
```
arcuri82, line 591: update docs... now we do not return a list, but a map (also clarify what are the string keys here)

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/RPCProblemDto.java
man-zhang: test case name which is used to provide warning info

PR #763, opened 2023-05-11
Tesis mongo

arcuri82:


arcuri82:
@hghianni @jgaleotti 
there were some issues with GitHub (eg, option approve workflows execution from external did not show up), so i tmp merge into this branch. i ll check the PR here. if need of changes, i ll either grant permission access on this branch, or can make PR to this PR's branch

arcuri82:
@jgaleotti  hi. looks like 1 test in ` GeneTest.testRandomized` fails for this branch. do u have time to look into it?

arcuri82:
@hghianni i granted you `write` rights. so u should be able to work directly on this branch, instead of using a fork

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/MongoInfo.java
```diff
@@ -0,0 +1,29 @@
+package org.evomaster.client.java.instrumentation;
+
+import java.io.Serializable;
+
+/**
+ * Info related to MONGO command execution.
+ *
+ * This class is IMMUTABLE
```
arcuri82, line 8: @hghianni this is not necessarily the case. you have references to `Object` type (eg, `collection`) that can be mutated from outside, making this class not immutable.

Either remove the comment about immutability, or enforce immutability (which might be tricky when dealing with `Object` refs, as would need to do deep-clones on input)

jgaleotti:
Hi @arcuri82 ,  
It looks as `GeneTest.testRandomized` is passing now. I guess this is completely unrelated to the work done by @hghianni, could we accept the PR and repair the flaky test case when it surfaces again?

arcuri82:
@hghianni @jgaleotti  merged ;) thx

PR #767, opened 2023-05-22
support customized mock db

man-zhang:
support customized method to set up mock objects for databases
calculate targets covered by seeded tests
add unit tests for RPC test writers
add e2e test for customized methods


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/rpc/MockDatabaseDto.java
```diff
@@ -0,0 +1,54 @@
+package org.evomaster.client.java.controller.api.dto.problem.rpc;
+
+/**
+ * mock object for database
+ */
+public class MockDatabaseDto {
```
arcuri82, line 6: add bit more info here on what Mocking a Database is here.
also, is it something specific only to RPC? or could it applied to REST and GraphQL as well? if so, shouldn't be under `.../rpc`. if not, add bit mention on why

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/rpc/MockDatabaseDto.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * a key refers to the Web API
+     */
+    public String appKey;
+
+    /**
+     * it refers to a type of database, eg, mybatis
```
arcuri82, line 14: ??? mybatis is a library to deal with DB migrations... eg, like flyway.
why is it considered here as database type?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestSampler.kt
```diff
@@ -112,7 +112,7 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
         }
 
         initAdHocInitialIndividuals()
-
+        initSeededTests()
```
arcuri82, line 115: why for GraphQL it is under `if (config.seedTestCases)` but not here?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestSampler.kt
```diff
@@ -169,17 +169,18 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
      */
     fun initAdHocInitialIndividuals(){
         customizeAdHocInitialIndividuals()
+    }
 
+    override fun initSeededTests(infoDto: SutInfoDto?) {
         // if test case seeding is enabled, add those test cases too
         if (config.seedTestCases) {
```
arcuri82, line 176: ok, i see. but we should keep things consistent

arcuri82 (changes requested):

arcuri82 (approved):

PR #768, opened 2023-05-25
more JPA constraints are supported

jgaleotti:


e2e-tests/spring-rest-openapi-v2/pom.xml
```diff
@@ -165,6 +165,13 @@
             <artifactId>okhttp</artifactId>
         </dependency>
 
+        <!-- for Geometry -->
+        <dependency>
+            <groupId>org.locationtech.jts</groupId>
+            <artifactId>jts-core</artifactId>
+            <version>1.19.0</version>
```
arcuri82, line 172: any specific reason why this version number is specified here and not in the main root pom file?

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v2/pom.xml
jgaleotti: My bad. I will change it.

arcuri82 (approved):

PR #777, opened 2023-07-03
Basic grpc

man-zhang:
grpc support

### DIFF
--- a/client-java/controller/pom.xml
+++ b/client-java/controller/pom.xml
@@ -196,10 +196,54 @@
             <scope>test</scope>
         </dependency>
 
+        <!--gRPC RPC test-->
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty-shaded</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-protobuf</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-stub</artifactId>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
+
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>${kr.motd.maven.version}</version>
+            </extension>
+        </extensions>
+
         <plugins>
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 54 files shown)


man-zhang:
@arcuri82 the latest commit is only for doc, then request the review.

client-java/controller/pom.xml
```diff
@@ -196,10 +196,54 @@
...
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
+
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
```
arcuri82, line 222: add a comment specifying what this extension is doing

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -835,8 +835,14 @@ public final void executeAction(RPCActionDto dto, ActionResponseDto responseDto)
                     responseDto.rpcResponse = resSchema.getDto();
                     if (dto.doGenerateAssertions && dto.responseVariable != null)
                         responseDto.assertionScript = resSchema.newAssertionWithJava(dto.responseVariable, dto.maxAssertionForDataInCollection);
-                    else
-                        responseDto.jsonResponse = objectMapper.writeValueAsString(response);
+                    else{
+                        try {
+                            responseDto.jsonResponse = objectMapper.writeValueAsString(response);
+                        }catch (JsonProcessingException e){
+                            // cannot convert to json
```
arcuri82, line 842: so if it happens we silently ignore it?

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/schema/types/JavaDtoSpec.java
```diff
@@ -0,0 +1,16 @@
...
+public enum JavaDtoSpec {
+    /**
+     * pure Java
+     */
+    PURE,
+
+    /**
+     * using proto3
+     */
+    PROTO3
```
arcuri82, line 15: add bit more info on why it is needed... eg, some examples

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/rpc/schema/types/JavaDtoSpec.java
```diff
@@ -0,0 +1,16 @@
+package org.evomaster.client.java.controller.problem.rpc.schema.types;
+
+/**
+ * supported specification of Java Dto
+ */
+public enum JavaDtoSpec {
+    /**
+     * pure Java
+     */
+    PURE,
```
arcuri82, line 10: `PURE` might not be best word here... "pure" functions have a very specific meaning, not sure if confusing then dealing with "pure" dtos.
what about `DEFAULT` ?

arcuri82 (changes requested):

man-zhang:
@arcuri82 there might be a bug in Github. the new commit somehow does not display in this PR (https://github.com/EMResearch/EvoMaster/commits/basic-grpc). If the CI passes, I will close this one, and create another PR in order to include all comments in this branch.

PR #785, opened 2023-08-08
reviewing exp configs

arcuri82:


arcuri82:
@man-zhang hi. few times `org.evomaster.e2etests.spring.rpc.examples.hypermutation.HypermutationEMTest` does timeout. we changed some default settings, might be related to that. can u look into it when you have time? thx

man-zhang:
yes.  I will check this.

man-zhang:
hi @arcuri82  the test seems fine now. I checked it locally, and the problem can not be reproduced.

PR #802, opened 2023-09-08
Rpc test split

man-zhang:
second study for industrial application


core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
```diff
@@ -90,6 +122,23 @@ object TestSuiteSplitter {
         return splitResult
     }
 
+    /**
+     * @return split test suite based on specified maximum number [limit]
+     */
+    fun splitSolutionByLimitSize(solution: Solution<ApiWsIndividual>, limit: Int) : List<Solution<ApiWsIndividual>>{
```
arcuri82, line 128: slightly confused... is this applied directly on the main Solution object? 

I mean, I would expect that, first there is a clustering of test cases in different groups (eg based on exceptions and HTTP status code). then, if any of such group is larger than limit N, that is split in test suite of size up to N.
ie, max test case number should be the last action done on the splitting process.

is that what happening here?

man-zhang: It can be applied at the first step or last step as the input parameter is `solution` which can be the original one or grouped ones. 
Currently, I set it as the last step. if you see its usage, all are based on `splitResult.splitOutcome`.
For instance, if the grouped test suite is named as `X` whose length is more than `N`,  now the test will be split into multiple ones and named as eg, `X_0`, `X_1` and so on.

Issue #701, opened 2023-02-04
EvoMaster process terminated abruptly when using BlackBox and a local file for schema


ElenaLobo:
Hi,
I am trying to run the EvoMaster using Black Box technique. I do not have access to the source code of the API but I have the schema downloaded as a local file and the API up and running in the remote server. I am also using authentication.
This is how my command looks like:
java -jar evomaster.jar  --blackBox true --bbSwaggerUrl file://microservice-swagger.yaml --bbTargetUrl  http://remote-host/api/v1/microservice --header0 "Authorization: Bearer <token>" --outputFormat JAVA_JUNIT_5 --outputFolder \GeneratedTCs --maxTime 30s --ratePerMinute 60

I got the following result, saying it is most likely a bug in the EvoMaster and asked to log this issue.

* EvoMaster version: 1.6.0
* WARNING: You are doing Black-Box testing, but you did not specify the 'problemType'. The system will default to RESTful API testing.
* Initializing...
* [ERROR] EvoMaster process terminated abruptly. This is likely a bug in EvoMaster. Please copy&paste the following stacktrace, and create a new issue on https://github.com/EMResearch/EvoMaster/issues
java.lang.reflect.InvocationTargetException: invokedynamic: method=public void org.evomaster.core.problem.rest.service.AbstractRestSampler.initialize(), target=org.evomaster.core.problem.rest.service.RestSampler@556d0826
        at com.netflix.governator.lifecycle.LifecycleMethods.methodInvoke(LifecycleMethods.java:316)
        at com.netflix.governator.lifecycle.LifecycleMethods.methodInvoke(LifecycleMethods.java:298)
        at com.netflix.governator.lifecycle.LifecycleManager.startInstance(LifecycleManager.java:266)
        at com.netflix.governator.lifecycle.LifecycleManager.add(LifecycleManager.java:175)
        at com.netflix.governator.guice.InternalLifecycleModule.onProvision(InternalLifecycleModule.java:81)
        at com.google.inject.internal.ProvisionListenerStackCallback$Provision.provision(ProvisionListenerStackCallback.java:117)
        at com.google.inject.internal.ProvisionListenerStackCallback.provision(ProvisionListenerStackCallback.java:66)
        at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:93)
        at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:296)
[...]

PR #818, opened 2023-10-04
Localfileschema bug

onurd86:


.gitignore
```diff
@@ -126,3 +126,4 @@ Migrations/
 /test-old-libraries/target/
 /e2e-tests/spring-web/target/
 /e2e-tests/spring-rest-mongo/target/
+/core/src/main/resources/openapi_pet.json
```
arcuri82, line 129: why this?

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
```diff
@@ -62,17 +63,40 @@ object OpenApiAccess {
...
-            Paths.get(URI.create(openApiUrl))
-        } else {
-            Paths.get(openApiUrl)
+
+        // path variable
+        var path: Path? = null;
+
+        // checking the given URL
+        try {
+            path = when (openApiUrl.startsWith(fileScheme, true)) {
```
arcuri82, line 75: this looks bit weird... why replacing an `if` statement with a `when` on booleans?

if it was due to the try/catch, that can be written as:

```
val path = try { if... 
                 catch(...)
```

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
```diff
@@ -62,17 +63,40 @@ object OpenApiAccess {
...
+
+        // file scheme
         val fileScheme = "file:"
-        val path = if (openApiUrl.startsWith(fileScheme, true)) {
-            Paths.get(URI.create(openApiUrl))
-        } else {
-            Paths.get(openApiUrl)
+
+        // path variable
+        var path: Path? = null;
```
arcuri82, line 71: why declaring it as nullable? was the compiler complaining?

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
```diff
@@ -62,17 +63,40 @@ object OpenApiAccess {
...
+                    " is not a valid path");
         }
 
-        return path.toFile().readText()
+        // Path should not be null, either
+        // it is set or an exception is thrown
+        if (path == null) {
+            throw SutProblemException("Could not set up the path: $openApiUrl");
+        }
+        else {
```
arcuri82, line 92: unnecessary `else` declaration, as the `then` throws an exception (ie, avoid identation here)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestSampler.kt
```diff
@@ -253,6 +253,11 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
         if (swagger.paths == null) {
             throw SutProblemException("There is no endpoint definition in the retrieved Swagger file")
         }
+        // Onur: to give the error message for invalid swagger
+        else if (swagger.paths.size == 0){
+            throw SutProblemException("The swagger file ${configuration.bbSwaggerUrl.toString()} " +
```
arcuri82, line 258: no need for the `.toString()`, as it is already a string... didn't IntelliJ flag it out?

core/src/test/kotlin/org/evomaster/core/problem/rest/OpenAPILocalURLIssueTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+/*
+Testing the local URL issue with OpenAPI, 4 test cases:
+- a local file which exists and a valid URL
+- a local file which does not exist and a valid URL
+- a local file which exists and an invalid URL
+- a local file which does not exist and an invalid URL
+- a local file which is not a valid swagger but a valid JSON
+- a local file which is not a valid JSON
+- an invalid url
```
arcuri82, line 18: the original issue in #701 was `IllegalArgumentException: URI path component is empty`. does any of these tests reproduce it?

core/src/test/resources/invalid_json.json
```diff
@@ -0,0 +1 @@
+AAAAAAAAAAA
```
arcuri82, line 1: under `src/test/resources` there is already a folder called `swagger` to contain swagger schemas, with several sub-folders. move those 3 files there, where appropriate, eg can create new folder `src/test/resources/swagger/urlissue`

arcuri82 (changes requested):

.gitignore
onurd86: Hi Andrea, 

Thank you for your comment. I used that file in the same directory as the main executable just to regenerate bug #701 before fixing and running test cases. I have removed that line since it is not necessary.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: Hi Andrea,

Thank you so much for your feedback. I have removed the else clause from there.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: Hi Andrea,

Thank you so much for your comment. I have rewritten the variable declaration according to the way you mentioned.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: Hi Andrea,

Thank you so much for your comment. I changed the code in a way that it throws an exception if the path is null in the variable declaration as:

val path = {
} ?: throw SutProblemException("Could not set up the path: $openApiUrl")

So, if the path is null, it throws an exception mentioning that the path could not be set. I put null before because of the compiler complaining. The way I changed is suggested by IntelliJ.

core/src/test/kotlin/org/evomaster/core/problem/rest/OpenAPILocalURLIssueTest.kt
onurd86: Hi Andrea,

Thanks for your feedback. I executed Evomaster with the following parameters:

--blackBox true --bbSwaggerUrl file://openapi_pet.json --bbTargetUrl "" http://remote-host/api/v1/microservice --header0 "Authorization: Bearer " --outputFormat JAVA_JUNIT_5 --outputFolder sample --maxTime 30s --ratePerMinute 60

I made sure that file://openapi_pet.json is also in the same directory as the executable. It did not provide the error message: "URI path component is empty", but it provided the error message "URI has an authority component". I am investigating the difference.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: I have removed "?: throw SutProblemException("Could not set up the path: $openApiUrl")" and just used try-catch for exception handling.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
```diff
@@ -62,16 +62,32 @@ object OpenApiAccess {
...
+                Paths.get(URI.create(openApiUrl))
+            }
+            else {
+                Paths.get(openApiUrl)
+            }
+        }
+        // Exception is thrown if the path is not valid
+        catch (e: Exception) {
+            // state the exception with the error message
+            throw SutProblemException("The file path provided for the OpenAPI Schema $openApiUrl," +
```
arcuri82, line 81: remove the `,`

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
```diff
@@ -62,16 +62,32 @@ object OpenApiAccess {
...
+                        " ended up with the following error: " + e.message)
         }
+
+        // If the path is valid but the file does not exist, an exception is thrown
         if (!Files.exists(path)) {
-            throw SutProblemException("Cannot find OpenAPI schema at file location: $openApiUrl")
+            throw SutProblemException("The provided swagger file does not exist: $openApiUrl")
```
arcuri82, line 87: use `OpenAPI` term instead of `swagger` (this latter is old name, linked to a specific company)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestSampler.kt
```diff
@@ -253,6 +253,11 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
         if (swagger.paths == null) {
             throw SutProblemException("There is no endpoint definition in the retrieved Swagger file")
         }
+        // Onur: to give the error message for invalid swagger
+        if (swagger.paths.size == 0){
+            throw SutProblemException("The swagger file ${configuration.bbSwaggerUrl} " +
```
arcuri82, line 258: use `OpenAPI` in output errors and logs. although we might still refer to `swagger` in old source code (which we should had refactored, but it is not a big deal)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: Hi Andrea,

Thanks for your comment. I have removed "," from the error message as requested.

core/src/main/kotlin/org/evomaster/core/problem/rest/OpenApiAccess.kt
onurd86: Hi Andrea,

Thanks for your comment, I have replaced swagger with OpenAPI in error messages.

arcuri82 (approved):

PR #821, opened 2023-10-05
support for javax.validation.constraints.Email

jgaleotti:


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/heuristic/ValidatorHeuristics.java
```diff
@@ -32,6 +31,7 @@ always get higher than this value (this can be achieved by using it as a base).
     private static final String PREFIX_HIBERNATE = "org.hibernate.validator.constraints.";
     private static final String PREFIX_JIRUKTA = "cz.jirutka.validator.collection.constraints.";
 
+    private static final String EMAIL_REGEX_PATTERN = "^(.+)@(.+)$";
```
arcuri82, line 34: shouldn't we use something bit more sophisticated? eg: https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s01.html

although we do not fully support all kinds of regex (eg no handling yet for `?:`)

arcuri82 (changes requested):

arcuri82 (approved):

PR #825, opened 2023-10-18
introducing DnsAction as an init group

arcuri82:


arcuri82:
@seran looks at the changed files. we can discuss it in meeting

PR #827, opened 2023-10-23
Continue with Mongo support

hghianni:


client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -281,6 +283,50 @@ private double calculateDistanceForInvertedType(InvertedTypeOperation operation,
...
+        String field = operation.getFieldName();
+        Object actualPoint = getValue(doc, field);
+
+        double x1 = Math.toRadians(operation.getLongitude());
+        double y1 = Math.toRadians(operation.getLatitude());
+        double x2;
+        double y2;
+
+        // GeoJsonPoint in document
+        if (isDocument(actualPoint) && getValue(actualPoint, "type").equals("Point") && getValue(actualPoint, "coordinates") instanceof List<?>) {
```
arcuri82, line 296: is this case sensitive? ie, always be `Point` and never `point`?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/CustomTypeToOasConverter.java
```diff
@@ -0,0 +1,6 @@
+package org.evomaster.client.java.instrumentation.object;
+
+public abstract class CustomTypeToOasConverter {
```
arcuri82, line 3: add some JavaDoc for this class. for example, what is `Oas`?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/CustomTypeToOasConverter.java
```diff
@@ -0,0 +1,6 @@
...
+    public abstract String convert();
+    public abstract boolean isInstanceOf(Class<?> klass);
```
arcuri82, line 5: add some JavaDocs for these methods

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -538,7 +538,7 @@ object RestActionBuilderV3 {
                         if (it !is String)
                             LoggingUtil.uniqueWarn(log, "an item of enum is not string (ie, ${it::class.java.simpleName}) for a property whose `type` is string and `name` is $name")
                         it.toString()
-                    } as MutableList<String>).apply { add("EVOMASTER") })
+                    } as MutableList<String>).apply {})
```
arcuri82, line 541: why removing this?

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
```diff
@@ -583,7 +583,7 @@ object RestActionBuilderV3 {
             "password" -> return createNonObjectGeneWithSchemaConstraints(schema, name, StringGene::class.java, enableConstraintHandling)//StringGene(name) //nothing special to do, it is just a hint
             "binary" -> return createNonObjectGeneWithSchemaConstraints(schema, name, StringGene::class.java, enableConstraintHandling)//StringGene(name) //does it need to be treated specially?
             "byte" -> return createNonObjectGeneWithSchemaConstraints(schema, name, Base64StringGene::class.java, enableConstraintHandling)//Base64StringGene(name)
-            "date" -> return DateGene(name)
+            "date" -> return DateGene(name, onlyValidDates = true)
```
arcuri82, line 586: why this?

core/src/main/kotlin/org/evomaster/core/search/gene/mongo/ObjectIdGene.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.evomaster.core.search.service.mutator.genemutation.AdditionalGeneMutationInfo
+import org.evomaster.core.search.service.mutator.genemutation.SubsetGeneMutationSelectionStrategy
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+/**
+ * In MongoDB, documents stored in a collection require a unique _id field that acts as a primary key.
+ * ObjectIds are used as the default value for the _id field
+ */
+class ObjectIdGene(
```
arcuri82, line 19: i do not understand why there is a need for such a new Gene, instead of creating it directly as Regex gene, eg, using `RegexHandler.createGeneForJVM(regex)`

arcuri82: also, shouldn't there be `ObjectIdClassReplacement` for its constructor, to enable taint analysis? 
@hghianni if you are unsure what I am talking about, have a chat with @jgaleotti . or we can setup a video meeting

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
hghianni: Hi @arcuri82, I'm uncertain about adding an extra value to the enum, as it could potentially generate invalid data and lead to casting issues. The same concern applies to handling invalid dates. Is there a purpose that I might be overlooking?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/CustomTypeToOasConverter.java
hghianni: OAS refers to Open API Specification.

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
arcuri82: ah, I see. sending invalid data in REST makes sense (robustness testing), but not when injecting data into a database. Maybe should add boolean input variable to check if invalid data should be added or not

core/src/main/kotlin/org/evomaster/core/search/gene/mongo/ObjectIdGene.kt
arcuri82: btw, regarding this latter point, it could be done in a further PR / branch

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
hghianni: Agreed, I will add the boolean.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
hghianni: It's always Point as it's explained [here](https://datatracker.ietf.org/doc/html/rfc7946#section-1.4).

core/src/main/kotlin/org/evomaster/core/search/gene/mongo/ObjectIdGene.kt
hghianni: I believe the decision to create a specific gene is better. While it's true that ObjectIdGene wraps a RegexGene, I consider it more than just a regex; it models a specific type of Mongo and also has it's particular output in EJSON. About enabling taint analysis, I will talk to Juan Pablo about it.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
arcuri82: ok. thx. add a link to that as comment

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
arcuri82: hi @hghianni  i need to refactor that class for some other reasons... so i ll do this. i ll let you know once branch is then merged into master

core/src/main/kotlin/org/evomaster/core/search/gene/mongo/ObjectIdGene.kt
arcuri82: @hghianni  if it needs special output, then yes, could make sense to model as its own gene... but creating a new genes is not trivial, unfortunately :( we still miss a lot of documentation (+ checks and tests) to make sure important parts are not missing (eg impact analsyis). but i guess could be fine for now

core/src/main/kotlin/org/evomaster/core/problem/rest/RestActionBuilderV3.kt
arcuri82: @hghianni you can pull a fix from master (adding of invalid values is controlled by a parameter now)

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
hghianni: Link added

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/CustomTypeToOasConverter.java
hghianni: Documentation added

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/CustomTypeToOasConverter.java
hghianni: Documentation added

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -281,6 +283,50 @@ private double calculateDistanceForInvertedType(InvertedTypeOperation operation,
...
+
```
arcuri82, line 297: hi @hghianni . you wrote "link added", but I do not see it here... did you push it?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/CursorPreparerClassReplacement.java
```diff
@@ -0,0 +1,54 @@
...
+        Object result = find.apply(mongoCollection);
+
+        long endTime = System.currentTimeMillis();
+
+        handleMongo(mongoCollection, query, true, endTime - startTime);
+
+        return result;
+    }
+
+    private static Object getField(Object object, String fieldName) {
```
arcuri82, line 45: move up this method into `ThirdPartyMethodReplacementClass`, as it is a utility function that could be re-used elsewhere

core/src/main/kotlin/org/evomaster/core/search/gene/mongo/ObjectIdGene.kt
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * The name of this gene
+     */
+    name: String,
+
+    private val id: RegexGene = RegexGene(
```
arcuri82, line 25: instead of creating it by hand, should use `RegexHandler.createGeneForJVM(regex)`. this will avoid issues in future, in case we change regex genes. also, it is more readable

arcuri82 (changes requested):
hi. sorry for late reply. busy period...

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
hghianni: Hi @arcuri82, I added a link about GeoJsonPoint to the GsonJsonPointToOasConverter but I will add another link here about the case sensitivity.

arcuri82 (approved):

PR #838, opened 2023-11-09
adding empty draft of security module, integrated after the search

arcuri82:


onurd86:
solution = securityRest.applySecurityPhase()

I executed Evomaster with the parameter --security true. I checked the generated unit test file. That file is empty since the method call "solution = securityRest.applySecurityPhase()" overrides the set of solutions with an empty set.

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -201,10 +186,41 @@ class Main {
...
+            if(config.security){
+                //apply security testing phase
+                LoggingUtil.getInfoLogger().info("Starting to apply security testing")
+
+                //TODO might need to reset stc, and print some updated info again
+
+                when(config.problemType){
+                    EMConfig.ProblemType.REST -> {
+                        val securityRest = injector.getInstance(SecurityRest::class.java)
+                        solution = securityRest.applySecurityPhase()
```
onurd86, line 202: This line:

solution = securityRest.applySecurityPhase()

overrides the variable solution, which contains the set of tests other than security tests.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.evomaster.core.search.Solution
+
+
+/**
+ * Service class used to do security testing after the search phase
+ */
+class SecurityRest {
+
+    fun applySecurityPhase() : Solution<RestIndividual>{
+
```
onurd86, line 14: I assume generators for Security tests are going to be written as classes and they will be called in this method. Is that accurate? As example, if I want to test one of the generated endpoints for SQL injection vulnerabilities, where should I implement this test?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestModule.kt
```diff
@@ -80,5 +80,9 @@ class RestModule(private val bindRemote : Boolean = true) : AbstractModule(){
 
         bind(HarvestActualHttpWsResponseHandler::class.java)
             .asEagerSingleton()
+
+        bind(SecurityRest::class.java)
+            .asEagerSingleton()
+
```
onurd86, line 86: bind(SecurityRest::class.java)
            .asEagerSingleton()

These two lines appear 3 times in different classes. I guess code duplication is inevitable in this case.

onurd86 (commented):
I did not see any issues other than the solution which contains all other tests being overridden by an empty set. 

To test writing a simple security test, we can provide 'OR 1=1" as an input for a simple API and see the response. The response should not contain any database results.

PR #839, opened 2023-11-14
Iteration 1/5 of advanced distance

Walrus2028:
Iteration 1/5 of advanced distance.

Features: number and boolean types, arithmetic expressions, AND and OR operators and comparison operators (11 integration tests + unit tests)

The rest of the iterations and the complete plan are in the email I sent minutes ago.

Some comments about this PR:

-I updated the JSQL version to the last one because I used that version in the advanced implementation. That change implied to also update the standard distance implementation because some classes and methods were deprecated.

-Using EM with the advanced distance flag, I tested a custom application I created that performs queries using columns of type integer. I obtained the same results as when using the standard distance. This test allowed me to find out that the JSQL dependency version had to be changed in the java client module and not in the sql distance module (in the last case, it won’t work).

-I added java docs to the classes. These java docs will evolve as classes get more complex in the next iterations. In general, I followed all the code conventions in the document you provided.

-I added unit tests for all helpers (in addition to previous integration and unit tests).

### DIFF
--- a/client-java/pom.xml
+++ b/client-java/pom.xml
@@ -25,6 +25,10 @@
         <module>distance-heuristics</module>
     </modules>
 
+    <properties>
+        <jsqlparser.version>4.7</jsqlparser.version>
+    </properties>
+
     <dependencyManagement>
         <dependencies>
             <dependency>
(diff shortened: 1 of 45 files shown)


Walrus2028:
Despite of the fact that I compiled EM with these changes (without tests because some tests were failing in my computer, not particularly in this branch, but in master) and I tested an application with it, I saw some errors due to JSQL library version. A solution could be removing the dependency from the top and allow each module to use their own version, since sometimes we want to modify only the sql distance implementation in the client and not the sql generation implementation in the core. Andrea, what do you think about it?

arcuri82:
@Walrus2028 regarding your comment, not sure I m following... what error did u get exactly? i don't think we use JSQL library in core, do we?

arcuri82:
also, did you do a mvn `install`, with the demo API using the snapshot dependency of client?

Walrus2028:
Andrea, JP, I already solved the issue using the same JSQL parser version for all the project and updating the deprecated classes/methods in the db constraints module, that was using and older version.

PR #843, opened 2023-11-17
Refactor JpaConstraint

agusaldasoro:
This PR includes only refactors and no change in functionality.
- Refactor in `ClassAnalyzer` which includes:
  - Moving annotation names into another file
  - Modified `getAnnotationName` method to remove validation of namespace not null outside
- Created a builder for `JpaConstraint`: `JpaConstraintBuilder` to avoid having a constructor with so many parameters
- Refactor `ClassAnalyzerTest` to have a `@BeforeEach` setup


arcuri82:
@agusaldasoro  thx! I don't have comments (seems good ;)), but cannot merge, due to PR coming from a fork (as CI is not run). I gave you write permission. create a branch in EM repository, merge this into that branch, and then a create a new PR from that branch (so we can check if CI is OK)

PR #844, opened 2023-11-17
handling of Collection fields

jgaleotti:


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
```diff
@@ -304,7 +304,7 @@ private static String getSchema(Type type, Boolean useRefObject, List<Class<?>>
 
         if ((klass != null && (klass.isArray() || List.class.isAssignableFrom(klass) || Set.class.isAssignableFrom(klass)))
                 ||
-                (pType != null && (List.class.isAssignableFrom((Class) pType.getRawType()) || Set.class.isAssignableFrom((Class) pType.getRawType())))) {
+                (pType != null && (Collection.class.isAssignableFrom((Class) pType.getRawType()) ))) {
```
arcuri82, line 307: what are examples of collection classes that should be mapped to an array? 

 also, if that is the case, the other equivalent part of the `||` should be updated accordingly, isn't it?

aslo, would be good to have a test to check this (eg with something that is no List nor Set)

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/object/ClassToSchema.java
jgaleotti: In ocvn, the class org.devgateway.ocds.persistence.mongo.flags.ReleaseFlags has a field "flaggedStats" of type Collection<FlagTypeCount>. I added a commit with a test case plus the changes to the other places where the logic was tested similarly.

arcuri82 (approved):

PR #847, opened 2023-11-23
taint for ObjectId(String)

jgaleotti:


arcuri82 (changes requested):
hi @jgaleotti  would be good to have 1 E2E test for this

arcuri82 (approved):

PR #851, opened 2023-11-23
docs: Add Evomaster Controller API Spec on Open API

agusaldasoro:
Add in docs the Open API Spec for Evomaster Controller

### DIFF
--- a/docs/spec/controller.yaml
+++ b/docs/spec/controller.yaml
@@ -1,0 +1,246 @@
+openapi: 3.0.3
+info:
+  title: EvoMaster - OpenAPI 3.0
+  description: |-
+    Evomaster Client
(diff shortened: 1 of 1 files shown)


arcuri82:
@agusaldasoro thanks, but this is not necessary. Having the schema could be useful, but:
- either the schema is automatically derived from the source code (there are plugins for this)
- or the schema is used to automatically generate the interfaces and DTOs for the API (there are tools for this)
In other words, code and schema must depend on each other somehow (ie a change in one should be reflected/detected in the other).
manually maintaining the schema and code in sync would be too much work :(

jgaleotti:
Sorry, I understood that tools were not working to extract the API schema. I told Agustina to do this in order to get a better understanding on how the controller works. My bad :(

agusaldasoro:
No problem! Will close the PR

PR #852, opened 2023-11-23
assertions if failed to generate test

arcuri82:


arcuri82:
@jgaleotti hi CI is still running, but it will fail. there is some issues when generating Mongo insertions, in which writer produces invalid code. by default, that does not crash EM. but now added assertions to crash during testing. 
please either you @jgaleotti or @hghianni create branch of this branch, try to fix the issue, and then make a PR toward this branch.
also note the refactoring of names/packages for E2E, to make it consistent with other E2E modules (somehow I missed it when I reviewed the PRs)

thx!

arcuri82:
the failing test is `MongoPersonsWithoutPostEMTest`

PR #854, opened 2023-11-25
Fix issue with test case writer when Mongo insertions fail

hghianni:


hghianni:
Hi @arcuri82, the changes we made to the construction of RegexGene within the ObjectIdGene had unintended consequences, leading to an increase in failed insertions. This issue arises from the two variables, extraPrefix and extraPostfix, within DisjunctionRxGene, which generate invalid ObjectIds. Consequently, some insertions in the MongoScriptRunner were successful while others were not. When an insertion fails, an exception is thrown, leaving ActionResults inconsistent. Without clarity on the success or failure of these actions, the MongoWriter should assume they failed an ignore them (this was not working as expected) to avoid issues in the generated test. But despite failure if some insertion was successful it may have produced a change in the response of a GET for example and we should be able to distinguish it and include it in the test.

To address this, the following changes were implemented:

- An additional check was introduced in MongoWriter to produce output only if at least one insertion was successful.
- In case of an insertion failure in MongoScriptRunner, the index is now stored, enabling the differentiation of successful insertions.

I still haven't made any changes related to ObjectIdGene. How do you think we should solve it?

arcuri82:
hi @hghianni  regarding `ObjectIdGene`, the regex is wrong:

`private val id: RegexGene = RegexHandler.createGeneForJVM("[0-9a-f]{24}")`

it should had been: `^[0-9a-f]{24}$`

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/operations/MongoInsertionResultsDto.java
```diff
@@ -1,12 +1,39 @@
...
 public class MongoInsertionResultsDto {
     /**
      * whether the insertion at the index of a sequence of Mongo insertions (i.e., {@link MongoDatabaseCommandDto#insertions})
      * executed successfully
      */
     public List<Boolean> executionResults = new ArrayList<>();
+    public Integer failedInsertionIndex = -1;
+
+    public Boolean insertionFailed() {
+        return failedInsertionIndex != -1;
```
arcuri82, line 18: better to write `>= 0`

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/operations/MongoInsertionResultsDto.java
```diff
@@ -1,12 +1,39 @@
...
+            for (int i = 0; i < failedInsertionIndex; i++) {
+                results.set(i, true);
+            }
+
+            executionResults = results;
+        }
+    }
+
+    private static int findFailedInsertion(Exception e) {
+        Matcher matcher = Pattern.compile("index (\\d+)").matcher(e.getMessage());
```
arcuri82, line 36: explain this regex in a comment.
also, `Pattern.compile("index (\\d+)")` could be saved in a static variable, to avoid recompute it each time. but check if `matcher` is thread-safe

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/database/operations/MongoInsertionResultsDto.java
hghianni: Hi @arcuri82, in this [link](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) says Instances of this class (referring to Pattern) are immutable and are safe for use by multiple concurrent threads. Instances of the Matcher class are not safe for such use. Does this mean we can store the Pattern in a static variable, but a new Matcher has to be created each time?

PR #864, opened 2023-12-04
support for int8 and int16 in createGeneForDto

jgaleotti:


arcuri82 (changes requested):
@jgaleotti sorry, but should avoid re-formatting files when doing PRs. it makes it hard to see the actual new changes. also, I guess it was done with an IDE? I do not like several of the formatting, eg splitting input argument on a single line  into 1 line per argument, etc. 
can you redo the PR without the auto-formatting? thx!

jgaleotti:
Sure, no problem.

PR #868, opened 2023-12-10
fix for email java regex

jgaleotti:


core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -88,6 +88,8 @@ atom
  : quote
  | patternCharacter+
  | DOT
+ | ESCAPED_DOT
+ | ESCAPED_PLUS
```
arcuri82, line 92: what about all other characters that could be escaped? 
are we going to add ad-hoc special cases for all of them?

I guess this could be fine for now... as handling of regex needs to improvements anyway...
but add a descriptive comment stating this is a temporary solution

arcuri82 (changes requested):

PR #869, opened 2023-12-14
Create Solver Module

agusaldasoro:
This PR creates the Solver module, as a simple first implementation.

The idea of this PR is to test the CI running of the test.

This code just:
- Starts a docker image of z3
- Loads from a file constraints written in `smt2`
- Runs z3 through the running Docker
- Stops the Docker


solver/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <artifactId>evomaster</artifactId>
+        <version>2.0.1-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.example</groupId>
+    <artifactId>solver</artifactId>
+
+    <properties>
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
```
arcuri82, line 17: we should target JDK 8, but still be able to run on latest LTS. note that we still haven't checked 21, but I guess shouldn't be much difference (if any) from 17

solver/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>1.7.24</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>RELEASE</version>
+            <scope>test</scope>
```
arcuri82, line 40: those version numbers should be in parent root pom. not here, unless there is a reason for it (in such a case, still would need a comment to clarify)

solver/src/main/java/ConstraintSolver.java
```diff
@@ -0,0 +1,47 @@
...
+
+    // The default entrypoint runs directly z3, we need to keep it running
+    public static final String ENTRYPOINT = "while :; do sleep 1000 ; done";
+
+    private final String containerPath = "/smt2-resources/";
+    private final GenericContainer<?>  z3Prover;
+
+    public ConstraintSolver() {
+
+        String resourcesFolder = System.getProperty("user.dir") + "/src/test/resources/";
```
arcuri82, line 20: hmmm... this should rather be a temp folder

solver/src/main/java/ConstraintSolver.java
```diff
@@ -0,0 +1,47 @@
...
+
+        ImageFromDockerfile image = new ImageFromDockerfile()
+                .withDockerfileFromBuilder(builder -> builder
+                        .from(Z3_DOCKER_IMAGE)
+                        .entryPoint(ENTRYPOINT)
+                        .build());
+
+        z3Prover = new GenericContainer<>(image)
+                        .withFileSystemBind(resourcesFolder, containerPath, BindMode.READ_WRITE);
```
arcuri82, line 29: why binding to actual FS? does solver write to disk, and we must read from disk to get the solver's results? if so, clarify

arcuri82 (changes requested):

solver/src/main/java/ConstraintSolver.java
agusaldasoro: The current implementation of the Z3 solver I'm using reads content either from STDIN or from a file. Since I couldn't find a way to send data through STDIN every time I want to execute a command, I've opted for the second option. Additionally, I need to dynamically generate various inputs, which is why I've linked the file system folder to the Docker volume. I then confirmed the functionality by testing the creation of a new file, and it worked as expected.

In summary, I'm reading input from a file, and the result is returned in STDOUT. Does that make sense? If you agree, I can also document this in the code for added clarity.

agusaldasoro:
@arcuri82, appreciate the review!

- I've fixed the `pom` file and added two dependencies. Does that look good to you?
- I've tweaked the solver implementation to make the resource folder configurable. When running the code, it should now use a `tmp` folder. Currently, with only tests in place, it reads the resource folder as intended.

pom.xml
```diff
@@ -406,6 +407,11 @@
                 <artifactId>slf4j-api</artifactId>
                 <version>1.7.24</version>
             </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-simple</artifactId>
+                <version>1.7.24</version>
```
arcuri82, line 413: this is the same version as `slf4j-api` on line 408, isn'it? if so, such version number should be declared as a variable under `<properties>`

solver/src/main/java/ConstraintSolver.java
arcuri82: hi. thanks. yes, but add this kind of info here in the code as a comment. 

for this prototype, we do not care too much about performance. but, in the long run, an approach that requires to communicate with an external process via files is not efficient.

arcuri82: maybe use `ConstraintSolver` as an abstract class or interface, and actual implementation (eg `ConstraintSolverZ3InDocker`) as a subclass

solver/src/main/java/ConstraintSolver.java
```diff
@@ -0,0 +1,45 @@
...
+
+        z3Prover.start();
+    }
+
+    @Override
+    public void close() {
+        z3Prover.stop();
+    }
+
+    public String solve(String exampleName) throws IOException, InterruptedException {
```
arcuri82, line 37: add JavaDoc specifying what is the input String, and what is the output String

arcuri82 (changes requested):

solver/src/main/java/ConstraintSolver.java
agusaldasoro: Agree! I've added a comment explaining this and also created the ConstraintSolver interface. Anyway, for now, the solve method receives the input file name as a parameter, the idea is to refactor that to receive a set of constraints and will do that on another PR.

agusaldasoro:
@arcuri82 applied all the feedback! One question, do you think that it's okay to have a new module name `solver` for this? Or should it be inside the `core`? Thanks!

arcuri82 (approved):

PR #879, opened 2024-01-16
Test writer changes to accommodate HostnameResolutionAction

seran:


core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -44,6 +45,10 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
         if (initializingMongoResults.any { (it as? MongoDbActionResult) == null })
             throw IllegalStateException("the type of results are expected as MongoDbActionResults")
 
+        val initializingHostnameResolutionActions = ind.individual
+            .seeInitializingActions()
+            .filterIsInstance<HostnameResolutionAction>()
+            .distinct()
```
arcuri82, line 51: why `distinct`? they should be by construction distinct. if not, it is a bug in the handling of insertions into individual

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -62,6 +67,13 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
                 },
                 lines, insertionVars = insertionVars, skipFailure = config.skipFailureSQLInTestFile)
         }
+
+        var anyHostnameResolutionAction = false
+        if (initializingHostnameResolutionActions.isNotEmpty()) {
+            anyHostnameResolutionAction = handleHostnameResolutionActions(lines, initializingHostnameResolutionActions)
+        }
+
+        return anyHostnameResolutionAction
```
arcuri82, line 76: why returning this?

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -97,13 +98,20 @@ abstract class TestCaseWriter {
...
+
         lines.indented {
             val ind = test.test
             val insertionVars = mutableListOf<Pair<String, String>>()
-            handleFieldDeclarations(lines, baseUrlOfSut, ind, insertionVars)
+            anyHostnameResolutionActions = handleFieldDeclarations(lines, baseUrlOfSut, ind, insertionVars)
             handleActionCalls(lines, baseUrlOfSut, ind, insertionVars, testCaseName = test.name, testSuitePath)
         }
 
+        if (anyHostnameResolutionActions) {
```
arcuri82, line 110: if a generated test fails, then this last statement would not be exectued, living the cache dirty. i guess should rather be in a `AfterEach` method in `TestSuiteWriter` if anywhere in the test suite there is need to use such DNS manipulation

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -136,6 +144,21 @@ abstract class TestCaseWriter {
...
+    ): Boolean {
+        var any = false
+
+        actions.forEach { action ->
+            lines.add("DnsCacheManipulator.setDnsCache(\"${action.hostname}\", \"${action.localIPAddress}\")")
+            lines.appendSemicolon(format)
+            any = true
+        }
+
+        return any
```
arcuri82, line 159: why this boolean? can just check if input `actions` is not empty?

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -202,14 +202,16 @@ abstract class ApiWsStructureMutator : StructureMutator() {
...
+
         // update impact based on added genes
-        if (mutatedGenes != null && config.isEnabledArchiveGeneSelection()) {
-            individual.updateImpactGeneDueToAddedInitializationGenes(
-                mutatedGenes,
-                old,
-                listOf(addedInsertions)
-            )
-        }
+//        if (mutatedGenes != null && config.isEnabledArchiveGeneSelection()) {
```
arcuri82, line 208: why removing this?

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/service/HttpWsExternalServiceHandler.kt
```diff
@@ -196,7 +198,7 @@ class HttpWsExternalServiceHandler {
             val action = HostnameResolutionAction(it.remoteHostName, hostnameLocalAddressMapping[it.remoteHostName]!!)
             output.add(action)
         }
-        return output
+        return output.distinct()
```
arcuri82, line 201: this likely would not work, as `HostnameResolutionAction` has no overridden `equals` based on `hostname`. but, anyway, already in `hostnameResolutionInfos` they should be kept distinct

core/src/main/kotlin/org/evomaster/core/problem/rest/resource/RestResourceCalls.kt
```diff
@@ -184,8 +184,8 @@ class RestResourceCalls(
      */
     fun seeActions(filter: ActionFilter): List<out Action> {
         return when (filter) {
-            ActionFilter.ALL -> sqlActions.plus(externalServiceActions).plus(mainActions)
-            ActionFilter.INIT -> sqlActions.plus(mongoDbActions)
+            ActionFilter.ALL -> sqlActions.plus(externalServiceActions).plus(mainActions) // FIXME: Is this correct?
```
arcuri82, line 187: once the refactoring of `EnterpriseIndividual` is completed, i ll refactor all those action filters. so, no need to worry too much about this now

core/src/main/kotlin/org/evomaster/core/problem/rest/service/RestResourceFitness.kt
```diff
@@ -206,6 +206,7 @@ class RestResourceFitness : AbstractRestFitness<RestIndividual>() {
         if (config.extractSqlExecutionInfo && config.probOfEnablingResourceDependencyHeuristics > 0.0)
             dm.updateResourceTables(individual, dto)
 
+        // FIXME: Inclusion of HostnameResolutionAction will have any impact?
```
arcuri82, line 209: this will be refactored... so for now it is fine to just leave a `FIXME` comment

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
seran: this is used to write, `clearDNSCache` at the end of the test case.

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
seran: fixed it at the insertion, forget to remove this.

core/src/main/kotlin/org/evomaster/core/problem/externalservice/httpws/service/HttpWsExternalServiceHandler.kt
seran: this is unnecessary now, since the insertion logic is fixed now.

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
seran: we are clearing at `BeforeEach`. this is unnecessary now. Will clean it,

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
seran: This was causing error at impact calculation (`org/evomaster/core/search/EvaluatedIndividual.kt:817`). Also `HostnameResolutionAction` doesn't have any `Gene`s. Is this necessary for impact calculation?

arcuri82: @seran no, not necessary. will actually be removed in a different branch. for now, just add comment stating commented out due to not having genes

arcuri82 (approved):

PR #880, opened 2024-01-19
support for LocalDateTime

jgaleotti:


e2e-tests/spring-rest-mongo/src/test/java/org/evomaster/e2etests/spring/rest/mongo/reservations/ReservationsEMTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Test
+    public void testRunEM() throws Throwable {
+
+        runTestHandlingFlakyAndCompilation(
+                "MongoReservationsEM",
+                "org.foo.spring.rest.mongo.MongoEMFitness",
+                10000,
+                (args) -> {
+                    args.add("--enableWeightBasedMutationRateSelectionForGene");
```
arcuri82, line 33: why this parameter?

e2e-tests/spring-rest-mongo/src/main/java/com/mongo/reservations/ReservationsController.java
```diff
@@ -0,0 +1,21 @@
...
+@RestController
+@RequestMapping(path = "/reservations")
+public class ReservationsController {
+
+    @Autowired
+    private ReservationRepository reservations;
+
+    @GetMapping("findAll")
+    public ResponseEntity<Void> findAll() {
+        int status = (reservations.findAll().isEmpty()) ? 400 : 200 ;
```
arcuri82, line 16: shouldn't we explicitely check that at least one `checkInDateTime` is not null?

arcuri82 (changes requested):

e2e-tests/spring-rest-mongo/src/test/java/org/evomaster/e2etests/spring/rest/mongo/reservations/ReservationsEMTest.java
jgaleotti: copy paste error

e2e-tests/spring-rest-mongo/src/main/java/com/mongo/reservations/ReservationsController.java
jgaleotti: ok. I will add a @NotNull annotation plus some logic in the controller

arcuri82 (approved):

Issue #759, opened 2023-04-28
Cannot find `EvoMaster_Test.java` file in the output folder


swag2198:
Hi,
On running EvoMaster blackbox on my macOS (Apple M1Pro, 13.1) for [petstore](https://petstore.swagger.io/) online API, I am getting the following four files as output:
```
- EvoMaster_fault_representatives_Test.java
- EvoMaster_faults_Test.java
- EvoMaster_others_Test.java
- EvoMaster_successes_Test.java
```
However, when my colleague runs the same version of the jar file in their Windows machine, they get the following four files as output:
```
- EvoMaster_faults_Test.java
- EvoMaster_others_Test.java
- EvoMaster_successes_Test.java
- EvoMaster_Test.java
```
Could you kindly shed some light on why I am not getting the `EvoMaster_Test.java` file in the output directory? Could it be because of java version issue as I am using java 11.0.18 (mentioned below), while they were using an old version of java (1.8).

## Other important info:
- version of EvoMaster (EM) used: black-box mode, version 1.6.1
- how EM is run (eg, if from JAR or from one of its OS installers): downloaded the jar file
- version of applicable runtimes (eg, JVM, NodeJS and .Net). For Java, can paste the output of `java --version`
```
java 11.0.18 2023-01-17 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.18+9-LTS-195)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.18+9-LTS-195, mixed mode)
```
- command-line options used to run EM
```
java -jar evomaster.jar --blackBox true --bbSwaggerUrl file:///Users/swagatam/Desktop/misc/AAT/RAFT_v2/data/petstore/swagger.json --outputFormat JAVA_JUNIT_4 --writeStatistics true --maxTime 120s
```

arcuri82:
Hi.
  thx for reporting this bug. `EvoMaster_Test.java` shouldn't had been generated, as it does not have a classifying label between `EvoMaster` and `Test`. I ll look into it. thx

PR #886, opened 2024-01-23
Fixing the issue: #759

onurd86:
https://github.com/EMResearch/EvoMaster/issues/759

By fixing the split method inside TestSuiteSplitter.kt.

Solution is not taken as it is for the executive summary. It is taken as everything being copied except Termination condition. By default, the solution object has the termination condition NONE. In this case, I changed it to Termination.SUMMARY.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
@@ -90,9 +90,28 @@ object TestSuiteSplitter {
 
         // BMR: Splitting support for new problems
         val sol = if(config.problemType == EMConfig.ProblemType.GRAPHQL){
-            solution as Solution<GraphQLIndividual>
+            //solution as Solution<GraphQLIndividual>
+
+            Solution(
+                solution.individuals as MutableList<EvaluatedIndividual<GraphQLIndividual>>,
+                solution.testSuiteNamePrefix,
+                solution.testSuiteNameSuffix,
+                Termination.SUMMARY,
+                solution.individualsDuringSeeding as List<EvaluatedIndividual<GraphQLIndividual>>,
+                solution.targetsDuringSeeding
+            )
+
         } else {
-            solution as Solution<RestIndividual>
+            //solution as Solution<RestIndividual>
+
+            Solution(
+                solution.individuals as MutableList<EvaluatedIndividual<RestIndividual>>,
+                solution.testSuiteNamePrefix,
+                solution.testSuiteNameSuffix,
+                Termination.SUMMARY,
+                solution.individualsDuringSeeding as List<EvaluatedIndividual<RestIndividual>>,
+                solution.targetsDuringSeeding
+            )
         }
         val metrics = mutableListOf(DistanceMetricErrorText(config.errorTextEpsilon), DistanceMetricLastLine(config.lastLineEpsilon))
         val errs = sol.individuals.filter {ind ->
--- a/core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
@@ -90,28 +90,9 @@ object TestSuiteSplitter {
 
         // BMR: Splitting support for new problems
         val sol = if(config.problemType == EMConfig.ProblemType.GRAPHQL){
-            //solution as Solution<GraphQLIndividual>
-
-            Solution(
-                solution.individuals as MutableList<EvaluatedIndividual<GraphQLIndividual>>,
-                solution.testSuiteNamePrefix,
-                solution.testSuiteNameSuffix,
-                Termination.SUMMARY,
-                solution.individualsDuringSeeding as List<EvaluatedIndividual<GraphQLIndividual>>,
-                solution.targetsDuringSeeding
-            )
-
+            solution as Solution<GraphQLIndividual>
         } else {
-            //solution as Solution<RestIndividual>
-
-            Solution(
-                solution.individuals as MutableList<EvaluatedIndividual<RestIndividual>>,
-                solution.testSuiteNamePrefix,
-                solution.testSuiteNameSuffix,
-                Termination.SUMMARY,
-                solution.individualsDuringSeeding as List<EvaluatedIndividual<RestIndividual>>,
-                solution.targetsDuringSeeding
-            )
+            solution as Solution<RestIndividual>
         }
         val metrics = mutableListOf(DistanceMetricErrorText(config.errorTextEpsilon), DistanceMetricLastLine(config.lastLineEpsilon))
         val errs = sol.individuals.filter {ind ->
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
```diff
@@ -113,7 +113,7 @@ object TestSuiteSplitter {
                 if(errs.size <= 1){
                     splitResult.splitOutcome = splitByCode(sol, config)
                     // TODO: BMR - what is the executive summary behaviour for 1 or fewer errors?
-                    splitResult.executiveSummary = sol
+                    splitResult.executiveSummary = sol.convertSolutionToExecutiveSummary()
```
arcuri82, line 116: not sure about this. if there is at most 1 error, then I think there should be no executive summary. so the logic here would be to leave it empty, and then make sure to not crash when checking its presence

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
```diff
@@ -0,0 +1,174 @@
...
+            args.add("--bbSwaggerUrl");
+            args.add(baseUrlOfSut + "/v2/api-docs");
+            args.add("--outputFormat");
+            args.add("JAVA_JUNIT_4");
+            args.add("--outputFolder");
+            args.add(outFolderName);
+            args.add("--executiveSummary");
+            args.add("true");
+            args.add("--maxTime");
+            args.add("30s");
```
arcuri82, line 59: that's quite an high value... that ll make the test run for at least 30s... is it really necessary?

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
```diff
@@ -0,0 +1,174 @@
...
+                } else if (s.equals("EvoMaster_Test.java")) {
+                    emptyFound = true;
+                }
+            }
+
+            // summary should be found, empty should not be found
+            Assert.assertTrue(summaryFound);
+            Assert.assertFalse(emptyFound);
+        } catch(Exception e) {
+            e.printStackTrace();
```
arcuri82, line 91: if exceptiont, test should fail

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
```diff
@@ -0,0 +1,174 @@
...
+            args.add("--bbSwaggerUrl");
+            args.add(baseUrlOfSut + "/v2/api-docs");
+            args.add("--outputFormat");
+            args.add("JAVA_JUNIT_4");
+            args.add("--outputFolder");
+            args.add(outFolderName);
+            args.add("--executiveSummary");
+            args.add("false");
+            args.add("--maxTime");
+            args.add("30s");
```
arcuri82, line 133: same comment as previous test

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
```diff
@@ -0,0 +1,174 @@
...
+                if (s.equals("EvoMaster_fault_representatives_Test.java")) {
+                    summaryFound = true;
+                } else if (s.equals("EvoMaster_Test.java")) {
+                    emptyFound = true;
+                }
+            }
+
+            // summary should be found, empty should not be found
+            Assert.assertFalse(summaryFound);
+            Assert.assertFalse(emptyFound);
```
arcuri82, line 163: there is a lot of duplicate code in these 2 tests... should refactor commonality in a private method

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+            for (String s : files) {
+                if (s.equals("EvoMaster_fault_representatives_Test.java")) {
+                    summaryFound = true;
+                } else if (s.equals("EvoMaster_Test.java")) {
+                    emptyFound = true;
+                }
+            }
+
+            // summary should be found, empty should not be found
```
arcuri82, line 87: if want to test the new logic, should have a SUT for which no error is found (because there is none to find), and, then, even when `executiSummary` is on no `_fault_representatives_` test suite is created

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
onurd86: Hi Andrea, 

Thanks for your comment. I have fixed this. Now, there is just finally block.

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
onurd86: Hi Andrea,

Thanks for your comment. I have made it 5 seconds since our only goal is to see files created in this test case. 5 seconds should be enough for this in my opinion.

core/src/main/kotlin/org/evomaster/core/output/TestSuiteSplitter.kt
onurd86: Thanks, Andrea for your comment. I kept that one as it is. However, before the executive summary file is created, I added a condition that checks if all test cases belong to the success category. If so, the executive summary is not created.

e2e-tests/spring-rest-openapi-v2/src/test/java/org/evomaster/e2etests/spring/examples/filecreationissue/FileCreationIssueEMTest.java
onurd86: Thanks Andrea for your suggestion. I have made a class and a static method as a helper to avoid code duplication.

arcuri82 (approved):

PR #888, opened 2024-01-25
removing impact warning messages 

jgaleotti:
Adding logic to avoid "unexpected handling on Initialization Action after repair" when it is not necessary


core/src/main/kotlin/org/evomaster/core/search/EvaluatedIndividual.kt
```diff
@@ -783,6 +784,11 @@ class EvaluatedIndividual<T>(
         }
         mutatedGenes.addedInitializationGenes.addAll(diff.flatMap { it.seeTopGenes() })
 
+        if (addedInsertions!!.flatten().isEmpty()) {
```
arcuri82, line 787: @man-zhang what would be the case of having `! diff.isEmpty()` but then this `if` is true? ie, the `diff` is not empty, but `addedInsertions`  is? 
would that be a bug in EM?

man-zhang: the difference might be due to db action repair. 
here, we aim to keep consistency between genes and their impacts.

man-zhang: Hi, @arcuri82 
I cannot comment on the code which has been merged.
Please check the line 777,

``val diff = individual.seeInitializingActions()
            .filter { !old.contains(it) && ((it is SqlAction && !it.representExistingData) || it is MongoDbAction) }``

I am not sure whether the `MongoDbAction` should be always considered as `diff`. 
Could you please have a check?

man-zhang:
hi, @arcuri82  might need some discussion about impacts of MongoDb and DNS actions

jgaleotti:
sound good to me!

man-zhang:
sounds good to me as well. it works for me before 3pm Oslo time.

arcuri82:
thx @man-zhang . @jgaleotti has created a meeting invitation

arcuri82:
@man-zhang  can you take over this branch? and make sure MongoDB works fine with impact collection?

jgaleotti:
Any updates on this PR?

arcuri82:
@jgaleotti hi. waiting for some more fixes in other branches before resuming work on this one.

arcuri82 (approved):

jgaleotti:
thxs!

JP

PR #889, opened 2024-01-27
fix in regex of externalServiceIP

jgaleotti:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2017,7 +2017,7 @@ class EMConfig {
 
     @Cfg("User provided external service IP.")
     @Experimental
-    @Regex("^127\\.((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){2}([3-9]|[0-9][0-9]|[0-2][0-5][0-5]?)\$")
+    @Regex("^127(\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}\$")
```
arcuri82, line 2020: @seran was the original motivation to have it like that the fact that `127.0.0.3` is the minimum allowed? in this case, should be explicitly stated in the `@Cfg` documentation, explaining the reason: `127.0.0.1` is localhost, and `127.0.0.2` is treated specially, ie `ExternalServiceSharedUtils.RESERVED_RESOLVED_LOCAL_IP`

@jgaleotti why the modification? you said it was responsible for crashes in Mongo experiments... but I am wondering if it is rather a bug in `generate_ip` of `exp.py`. @seran can you check?

seran: @arcuri82 yeah, the idea is allow from `127.0.0.3`. I can update in the current branch I'm working to include the documentation.

seran: @jgaleotti the current `exp.py` script in the master has `last_generated_ip = "127.0.0.2"`. If you're using the latest version of `EvoMaster`, please change the address to `127.0.0.3`.

@arcuri82 I'll update script as well.

jgaleotti: Hi @arcuri82, I think the problem is a bit more complicated than allowing 127.0.0.1 and 127.0.0.2.
The following addresses were generated by generate_ip and were not considered valid by the current Regex annotation for externalServiceIP:

- 127.0.0.162
- 127.0.0.182
- 127.0.1.1
- 127.0.1.161
- 127.0.1.181
- 127.0.10.1
- 127.0.10.161
- 127.0.10.181
- 127.0.11.1
- 127.0.2.161
- 127.0.2.181
- 127.0.3.1
- 127.0.3.161
- 127.0.3.181
- 127.0.4.1
- 127.0.4.161
- 127.0.4.181
- 127.0.5.1
- 127.0.5.161
- 127.0.5.181
- 127.0.6.1
- 127.0.6.161
- 127.0.6.181
- 127.0.7.1
- 127.0.7.161
- 127.0.7.181
- 127.0.8.1
- 127.0.8.161
- 127.0.8.181
- 127.0.9.1
- 127.0.9.161
- 127.0.9.181

I think this is not a problem in generate_ip because all these addresses are higher than 127.0.0.2.

Cheers,
JP

jgaleotti: If externalServiceIP range is from 127.0.0.3 to 127.255.255.255, the following pattern seems to do the trick:
^127\.((0\.){2}([3-9]|[1-9]\d|[12]\d\d)|0\.([1-9]|[1-9]\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)|([1-9]|[1-9]\d|[12]\d\d)(\.([1-9]?\d|[12]\d\d)){2})$
(generated by https://www.analyticsmarket.com/freetools/ipregex/)

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2017,7 +2017,7 @@ class EMConfig {
 
     @Cfg("User provided external service IP.")
     @Experimental
-    @Regex("^127\\.((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){2}([3-9]|[0-9][0-9]|[0-2][0-5][0-5]?)\$")
+    @Regex("^127\\.((0\\.){2}([3-9]|[1-9]\\d|[12]\\d\\d)|0\\.([1-9]|[1-9]\\d|[12]\\d\\d)\\.([1-9]?\\d|[12]\\d\\d)|([1-9]|[1-9]\\d|[12]\\d\\d)(\\.([1-9]?\\d|[12]\\d\\d)){2})\$")
```
arcuri82, line 2020: i don't understand this regex :( not saying it is wrong, just that it is not easy to read. should split in different strings (eg `const val`) with some explanation on each step

arcuri82 (changes requested):

arcuri82 (approved):

PR #890, opened 2024-01-30
Refactor enterprise individual

man-zhang:
for rest problem


man-zhang:
Hi, @arcuri82 , do you have time to check the test `MinimizeEMTest`?

arcuri82:
hi @man-zhang  the issue seems in `RestResourceFitness`, line 82. Once an individual is flattened, we should still be able to compute its fitness, and that should not change (in theory). but fitness function now is not aware of such flattening

arcuri82 (approved):

PR #893, opened 2024-02-01
Solve to SqlAction in DbConstraintSolver using SqlActionBuilder

agusaldasoro:
Refactor `DbConstraintSolver` to return a list of `SqlAction` instead of a string


solver/pom.xml
```diff
@@ -35,6 +35,16 @@
             <groupId>org.evomaster</groupId>
             <artifactId>evomaster-client-java-sql</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.evomaster</groupId>
```
arcuri82, line 39: add comment that this creates a circular dependency. 
in long run, we don't want this module to depend on core, if want to use it outside of EM.
fine for now, but future might need an intermediate representation  X (eg convertion from X to SqlAction would be done inside core)

arcuri82: or we might move this module directly inside core... let's leave it here for now, we can think about its integration later on

arcuri82 (changes requested):

solver/pom.xml
agusaldasoro: Oh, okay! Will leave a comment with that info, thanks!

arcuri82 (approved):

Issue #873, opened 2023-12-26
Bug in GraphQL Blackbox mode


marcellomaugeri:
- Stacktrace
`java.lang.reflect.InvocationTargetException: invokedynamic: method=public void org.evomaster.core.problem.graphql.service.GraphQLSampler.initialize(), target=org.evomaster.core.problem.graphql.service.GraphQLSampler@4074023c
        at com.netflix.governator.lifecycle.LifecycleMethods.methodInvoke(LifecycleMethods.java:316)
        at com.netflix.governator.lifecycle.LifecycleMethods.methodInvoke(LifecycleMethods.java:298)
        at com.netflix.governator.lifecycle.LifecycleManager.startInstance(LifecycleManager.java:266)
        at com.netflix.governator.lifecycle.LifecycleManager.add(LifecycleManager.java:175)
        at com.netflix.governator.guice.InternalLifecycleModule.onProvision(InternalLifecycleModule.java:81)
        at com.google.inject.internal.ProvisionListenerStackCallback$Provision.provision(ProvisionListenerStackCallback.java:117)
        at com.google.inject.internal.ProvisionListenerStackCallback.provision(ProvisionListenerStackCallback.java:66)
        at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:93)
        at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:296)
        at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
        at com.google.inject.internal.SingletonScope$1.get(SingletonScope.java:169)
        at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:45)
        at com.google.inject.internal.FactoryProxy.get(FactoryProxy.java:60)
        at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40)
        at com.google.inject.internal.SingletonScope$1.get(SingletonScope.java:169)
        at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:45)
        at com.google.inject.internal.InternalInjectorCreator.loadEagerSingletons(InternalInjectorCreator.java:213)
        at com.google.inject.internal.InternalInjectorCreator.injectDynamically(InternalInjectorCreator.java:186)
[...]

PR #894, opened 2024-02-03
Issue #873: Provide a more detailed error if the introspection query's connection fails

marcellomaugeri:
Solve Issue #873


core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -61,35 +63,38 @@ class IntrospectiveQuery {
...
                     }
                 """.trimIndent(), MediaType.APPLICATION_JSON_TYPE)
-
-        //TODO check if TCP problems
-        val response = SearchTimeController.measureTimeMillis({ ms, res ->
+        val response: Response?
```
arcuri82, line 66: thx. in contrast to Java, in Kotlin most statements are expressions. so can write something directly like `val response = try {...`

arcuri82: also why nullable with `?` ? 
if want to check case for null, could make it non-null, with explicit throw exception after `try` with the `?:` Elvis operator

core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -61,35 +63,38 @@ class IntrospectiveQuery {
...
-                for (h in list) {
-                    request = request.header(h.first, h.second)
+                    for (h in list) {
+                        request = request.header(h.first, h.second)
+                    }
+                    request.buildPost(query)
+                            .invoke()
+                } catch (e: Exception) {
+                    log.error("Failed query to '$graphQlEndpoint' :  $query")
+                    throw e
```
arcuri82, line 82: the `throw SutProblemException(` should rather be done here, with `e.message` added to the message of the error, something like:
```
throw SutProblemException("Failed introspection query to '$graphQlEndpoint', please check connection and URL format. Error: ${e.message}" )
```

core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -61,35 +63,38 @@ class IntrospectiveQuery {
...
-                    .invoke()
-            } catch (e: Exception) {
-                log.error("Failed query to '$graphQlEndpoint' :  $query")
-                throw e
-            }
-        })
+            })
+        } catch (e: Exception) {
+            throw SutProblemException("Failed introspection query to '$graphQlEndpoint', please check connection and URL format")
+        }
```
arcuri82, line 87: not needed here of this outer try/catch. see previous comment

core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt
```diff
@@ -61,35 +63,38 @@ class IntrospectiveQuery {
...
 
 
 
 
         /*
            Extract the body from response as a string
         */
-        val body = response.readEntity(String::class.java)
+        val body = response?.readEntity(String::class.java)
```
arcuri82, line 95: not needed `?` if variable is made non-nullable, with explicit `?:` check on nullability (if needed)

core/src/main/kotlin/org/evomaster/core/search/service/SearchTimeController.kt
```diff
@@ -28,10 +28,13 @@ class SearchTimeController {
...
-            loggingFunction.invoke(System.currentTimeMillis() - startTime, result)
-
-            return result
+            try {
+                val result: T = function.invoke()
+                loggingFunction.invoke(System.currentTimeMillis() - startTime, result)
+                return result
+            } catch(e: Exception) {
+                throw e
+            }
```
arcuri82, line 37: what is the reason of this? why catching an exception and throwing it again without doing anything with it?

arcuri82 (changes requested):
@marcellomaugeri  thx for providing a PR! see my comments

marcellomaugeri:
@arcuri82 Sorry for the delay in troubleshooting. I think it's all sorted now, I was a bit confused as I'm fairly new to Kotlin, thanks for the explanations. However I'm still not sure about the double try-catch to get the answer, could you please check again? Thank you.

arcuri82:
hi @marcellomaugeri 
 sorry, but I don't understand why you are having a double-catch.
Can you please check the branch `related-to-pr-894`, to see if that solves the issue you have? 
thx

marcellomaugeri:
@arcuri82 yes, your branch solves the issue, thank you for your time and patience. I was overcomplicating things! Thank you for your suggestions, I am quite new to Kotlin and I got confused because the code was not compiling and then I added the double-catch, the rest you know

PR #896, opened 2024-02-07
re-enabled all WM tests to start debugging issues on CI

arcuri82:


core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestFitness.kt
```diff
@@ -808,7 +796,14 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
         )
 
-        handleExternalServiceInfo(individual, fv, dto.additionalInfoList)
+        val wmStarted = handleExternalServiceInfo(individual, fv, dto.additionalInfoList)
+        if(wmStarted){
+            /*
+                Quite tricky... if a WM instance was started as part of this test case evaluation,
+                then the results are invalid, as they are based on WM not being there.
+             */
+            return null
```
arcuri82, line 805: is this still valid? 
in last meeting, didn't we write `We have on purpose a IP address that is left empty (nothing running on it) to deal with such case.
So, when we have an host lookup in SUT (with no port), defaulting to resolve to that in Hostname Action
should solve such issue, isn't it?
So, no need to kill the test. Actually, make sure we can cover those cases (eg, we have a E2E in which explicitely
check it is handle by not covering 400 status in a branch)`
but maybe best to do it in a different branch. i ll create a Trello card for it

arcuri82: i guess maybe we could just comment out `null` here, as anyway this behavior is already disabled in the `Archive`

e2e-tests/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/wiremock/jsonarray/WmJsonArrayEMTest.kt
```diff
@@ -49,11 +45,9 @@ class WmJsonArrayEMTest : SpringTestBase() {
 
                 assertTrue(solution.individuals.size >= 1)
 
-                if (!CIUtils.isRunningGA()) {
-                    //FIXME same issue as other WM tests... pass locally. Maybe should try again on CircleCI
-                    assertHasAtLeastOne(solution, HttpVerb.GET, 200, "/api/wm/jsonarray", "OK X")
-                    assertHasAtLeastOne(solution, HttpVerb.GET, 200, "/api/wm/jsonarray", "OK X and Y")
-                }
+                //FIXME same issue as other WM tests... pass locally. Maybe should try again on CircleCI
```
arcuri82, line 48: is this comment still relevant?

core/src/main/kotlin/org/evomaster/core/search/service/Archive.kt
```diff
@@ -390,6 +390,11 @@ class Archive<T> where T : Individual {
      */
     fun addIfNeeded(ei: EvaluatedIndividual<T>): Boolean {
 
+        // TODO: Disabled, since it is not necessary
```
arcuri82, line 393: actually, this should be put back.
if there is a deth sentence, do not add individual.
this logic could still be used in other contexts.
however, we do not give anymore death sentence if a test starts a WM instance

arcuri82 (commented):
@seran  please see new comments I wrote

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AbstractRestFitness.kt
seran: If we don't disable in the `Archive`, we have to return `null` at this point to avoid tests without WM setup in the generated tests.

seran:
@arcuri82 PR is ready for review

PR #909, opened 2024-02-20
Add support in solver for more than 1 table and all numeric types

agusaldasoro:
- Add support for more than 1 table
- Add support for all numeric types
- Add support for enum genes
- Added a new file with tests `SqlBuilderFromSchemaTest` that contains all the parsing issues that I found while testing my code.


solver/src/main/java/DbConstraintSolverZ3InDocker.java
```diff
@@ -96,54 +104,119 @@ String solveFromFile(String fileName) {
...
+                    }
+
+                    String solvedValue = solvedConstraints.get(tableVariableKey);
+
+                    if (currentGene instanceof EnumGene) {
+                        // There is nothing to do, as the enum has the values already parsed
+                        return;
+                    }
+
+                    if (currentGene instanceof IntegerGene) {
```
arcuri82, line 152: maybe not for now, but could leave a comment... aren't we missing info on constraints of the values? i mean, these genes take as input min/max info as well

solver/src/main/java/Smt2Writer.java
```diff
@@ -28,7 +24,7 @@ public class Smt2Writer  {
     private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(Smt2Writer.class.getName());
 
     // The variables that solve the constraint
-    private final Set<String> variables = new HashSet<>();
+    private final Map<String, String> variables = new HashMap<>();
```
arcuri82, line 27: everytime you have maps as fields, in comments always specify what is the 'key' and 'value' entries, especially when usig primitives such as String

solver/src/main/java/Smt2Writer.java
```diff
@@ -125,30 +122,81 @@ private Pair<Set<String>, String> parseCheckExpression(SqlCondition condition) {
...
 
                 return new Pair<>(variables, toOr(orMembers));
             }
 
+            if (condition instanceof SqlInCondition) {
+                SqlInCondition inCondition = (SqlInCondition) condition;
+
+                String columnName = inCondition.getSqlColumn().getColumnName();
+                String variable = asTableVariableKey(table, columnName);
+                Map<String, String> variables = new HashMap<>();
```
arcuri82, line 138: see previous comment: specify key and value entries in a comment

solver/src/main/java/Smt2Writer.java
```diff
@@ -125,30 +122,81 @@ private Pair<Set<String>, String> parseCheckExpression(SqlCondition condition) {
...
+            case "INTEGER":
+            case "BIGINT":
+            case "LONG":
+                return "Int";
+            case "FLOAT":
+            case "DOUBLE":
+                return "Real";
+        }
+
+        throw new RuntimeException("The sql type is not supported by smt solver");
```
arcuri82, line 185: concatenate the `sqlType` to this error message

solver/src/test/java/SqlBuilderFromSchemaTest.java
```diff
@@ -0,0 +1,234 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+
+public class SqlBuilderFromSchemaTest {
+
+    private static Connection connection;
+
+    @BeforeAll
+    static void setup() throws Exception {
+         connection = DriverManager.getConnection("jdbc:h2:mem:constraint_test", "sa", "");
```
arcuri82, line 31: are these connections ever closed? or is it possible issue of test dependencies if using same DB url in different tests? (as they would work on same DB instance)

arcuri82 (changes requested):

solver/src/main/java/DbConstraintSolverZ3InDocker.java
agusaldasoro: I'm not sure if I understood correctly the comment. 
Do you mean that `currentGene` already contains the info for min and max values? If that's the case, then when copying the value from the new one (`currentGene.copyValueFrom(geneWithSolvedValue)`), those values persist.

Anyway, as in `SqlBuilderFromSchemaTest` there are some cases in the max and min are not correctly parsed from the constraints, so I will address that in another PR.

arcuri82: yes. based on constraints defined in schema, `currentGene` might already have info on min and max values.

solver/src/main/java/DbConstraintSolverZ3InDocker.java
```diff
@@ -96,54 +104,137 @@ String solveFromFile(String fileName) {
...
+                        Integer value =  Integer.parseInt(solvedValue);
+                        Gene geneWithSolvedValue = new IntegerGene(
```
arcuri82, line 172: @agusaldasoro also, thinking about it... but why creating a new gene here? can't you just modify `value` directly?

solver/src/main/java/DbConstraintSolverZ3InDocker.java
agusaldasoro: Yes, you are right. I will update only the value from the gene

solver/src/main/java/DbConstraintSolverZ3InDocker.java
agusaldasoro: Yes! done!

arcuri82 (approved):

PR #913, opened 2024-02-27
Removing default WM action for hostname

seran:


core/src/main/kotlin/org/evomaster/core/problem/enterprise/EnterpriseIndividual.kt
```diff
@@ -343,6 +343,13 @@ abstract class EnterpriseIndividual(
         killChildren { it is SqlAction && actions.contains(it)}
     }
 
+    /***
+     * remove specified list of [HostnameResolutionAction] from the initializing actions.
+     */
+    fun removeHostnameResolutionAction(actions: List<HostnameResolutionAction>) {
+        killChildren {  it is HostnameResolutionAction && actions.contains(it) }
```
arcuri82, line 350: as it is, it would not work properly. if checking for equality (ie in `contains`), need to make sure that `HostnameResolutionAction` overrides `equals` and `hasthTo` methods

arcuri82 (changes requested):
@seran  sorry, seems like I missed the review request for this PR

core/src/main/kotlin/org/evomaster/core/problem/enterprise/EnterpriseIndividual.kt
seran: added `hashCode` and `equals`

PR #929, opened 2024-03-08
re-enabled correct check for slash

arcuri82:


arcuri82:
@onurd86 you can look at these changes in this branch, related to the '/' issue

PR #936, opened 2024-03-19
working on new integration tests for REST

arcuri82:


arcuri82:
@onurd86 you can look at the modifications done in this branch (now merged). next week we can do some pair programming and use the scaffolding here to setup some integration tests

PR #940, opened 2024-03-26
refactoring parsing of dto into auth objet

arcuri82:


arcuri82:
@onurd86 refactoring is now merged

PR #944, opened 2024-04-04
fixed issue with non-determinism in tests

arcuri82:


arcuri82:
@seran branch is merged

PR #953, opened 2024-04-10
Changes to WireMock reset location

seran:


core/src/main/kotlin/org/evomaster/core/search/service/SearchAlgorithm.kt
```diff
@@ -90,6 +94,8 @@ abstract class SearchAlgorithm<T> where T : Individual {
 
         ssu.enabled = false
 
+        externalServiceHandler.resetWireMockServers()
```
arcuri82, line 97: hmmm... from an architectural point of view, no sure we should have reference to WM in `SearchAlgorithm`. would need to discuss where best to put it

arcuri82 (changes requested):

arcuri82:
@seran conflicts after merging other PRs into master

core/src/main/kotlin/org/evomaster/core/search/service/SearchAlgorithm.kt
seran: can we move the `Minimizer` from `SearchAlgorithm` to the `Main.initAndRun`?

arcuri82: hmmm... in theory could be possible, but i would prefer not.

why would be important to reset WM before minimization? doesn't each test case using WM resetting its instances anyway?

seran: I did this just to make sure WM is clean before `Minimizer`. Let me disable the change and see whether it works or breaks.

arcuri82 (approved):

PR #955, opened 2024-04-11
Apply new optimal set of settings after tuning

amidgol:
This change is based on the results of the tuning paper. We found a new set of configs to be optimal, detecting a higher number of faults while keeping a competitive degree of line and branch coverage. This setting is referred to as "best faults" in the paper.


arcuri82 (changes requested):
@amidgol  thanks. you should ask me to review the PR once the CI passes. it fails now. reason is that modifying `EMConfig` requires to run `ConfigToMarkdown` to update the documentation. run it, check CI passes, and then ask again for my review

arcuri82:
@amidgol some E2E tests fail after the change of parameters. This shouldn't really happen (but might be issue with the E2E).
Few seem related to `AdaptiveHypermutation`, so you might need to have a chat or help from @man-zhang for looking into it

arcuri82 (approved):

PR #972, opened 2024-04-30
fix bug extracting a schema from a database with duplicate table names

jgaleotti:


client-java/sql/src/main/java/org/evomaster/client/java/sql/SchemaExtractor.java
```diff
@@ -557,11 +596,11 @@ private static void handleTableEntry(Connection connection, DbSchemaDto schemaDt
 
         tableDto.primaryKeySequence.addAll(primaryKeySequence.values());
 
+        //ResultSet columns = md.getColumns(null, schemaDto.name, tableDto.name, null);
```
arcuri82, line 599: isn't this commented line identically to the following one? shouldn't it be removed?

arcuri82 (changes requested):

client-java/sql/src/main/java/org/evomaster/client/java/sql/SchemaExtractor.java
jgaleotti: yes. Copy paste error.

arcuri82 (approved):

PR #973, opened 2024-05-02
Changes to accommodate WM in generated tests regardless of HttpWsExternalService is available or not 

seran:


e2e-tests/spring-rest-openapi-v3/em.yaml
```diff
@@ -0,0 +1,89 @@
+### Template configuration file for EvoMaster.
```
arcuri82, line 1: why adding this to the repository? was it a mistake?

arcuri82 (changes requested):

e2e-tests/spring-rest-openapi-v3/em.yaml
seran: yes. somehow missed it.

PR #990, opened 2024-05-26
Adding Python as output option, creates empty python file

Pgarrett:
# What
Adding Python as an output format option only allowed for black-box testing. Currently, selecting this output creates a python file with no tests, but only the block comment for describing the test file.


arcuri82 (changes requested):
@Pgarrett  thx for your first PR ;)

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -273,33 +272,34 @@ class TestSuiteWriter {
...
-        lines.add(" * Used time: ${searchTimeController.getElapsedTime()}")
+        lines.addBlockCommentLine(" Used time: ${searchTimeController.getElapsedTime()}", format)
         classDescriptionEmptyLine(lines)
-        lines.add(" * Needed budget for current results: ${searchTimeController.neededBudget()}")
+        lines.addBlockCommentLine(" Needed budget for current results: ${searchTimeController.neededBudget()}", format)
         classDescriptionEmptyLine(lines)
-        lines.add(" * ${solution.termination.comment}")
-        lines.add(" */")
+        lines.addBlockCommentLine(" ${solution.termination.comment}", format)
+        lines.endCommentBlock(format)
```
arcuri82, line 302: weird... GitHub cancelled my comment... fortunately it was just one :)

anyway, maybe best to move `format` directly in constructor of `Lines` so that it does not have to given as input everywhere (and anyway it would not change for a test)

arcuri82 (changes requested):

arcuri82:
@Pgarrett  btw, it is best if you work in a branch of a clone of EM, and not a fork. right now, there is no good way to run CI for external forks :(  (it is a long story...)

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Good idea, I'll most likely close this PR and open one straight from the EM clone with the change included if it's ok for you.

I had originally tried to push a branch to EM but something happened, that's why I moved to a fork. I'll try again and let you know in case there's any issue.

arcuri82: @Pgarrett sounds good. btw, `master` branch is protected, cannot push directly to it

Pgarrett: Yes, obviously, but I tried pushing my own and was unable to do so. I'll try again @arcuri82 . Thanks!

Pgarrett:
Closing this PR as discussed since a new one has been opened with a branch straight from the master branch

PR #991, opened 2024-05-28
Adding Python as output option, creates empty python file

Pgarrett:
# What
Adding Python as an output format option only allowed for black-box testing. Currently, selecting this output creates a python file with no tests, but only the block comment for describing the test file.


Pgarrett:
@arcuri82 just as an FYI, since I changed the `Lines` constructor to take `OutputFormat`, I changed the methods that took format as an input parameter and removed it. If you feel that makes for much more files than expected in the PR, I can revert it and send them in a separate commit, it's the second commit of this PR.

arcuri82 (approved):

PR #996, opened 2024-05-30
Fix JS & Dotnet minimization

jcbasso:
While working in a driver and comparing it with different test SUTs I found that JS and Dotnet were having a big difference on line coverage when comparing it with Java solution. (from expected ~95% to ~55% of line coverage in NCS)

A bit more of debugging made me realize that the issue was when using the param minimization=true.

After a bit of more debugging I found that there was added in 2.0.0 (if I am not mistaken) a new param in the testResults endpoint called `allCovered` that was being ignored in both controllers.

I added this new param in both of them and from my testings in NCS looks like everything is working as intended now hitting the expected line coverage with both drivers.

### DIFF
--- a/client-dotnet/src/EvoMaster.Controller/EmbeddedSutController.cs
+++ b/client-dotnet/src/EvoMaster.Controller/EmbeddedSutController.cs
@@ -23,6 +23,10 @@ namespace EvoMaster.Controller {
             ExecutionTracer.SetAction(new Action(dto.Index, dto.InputVariables));
         }
 
+        public override IList<TargetInfo> GetAllCoveredTargetInfos() {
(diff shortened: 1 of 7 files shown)


arcuri82:
hi @jcbasso ,

 thanks for your interest in EM, but, as stated in the documentation: `Note: there was initial support for other languages as well, like for example JavaScript/TypeScript and C#, but they are not in a stable, feature-complete state. The support for those languages has been dropped, at least for the time being.`

we no longer support JS and C# for backend instrumentation (it is way too much work, unfortunately).

jcbasso:
Oh alright Andrea, thanks for the input!

Feel free to close the PR then if you believe you don't want to add new changes to them!

PR #999, opened 2024-06-05
Escape canonical names for WM

seran:


arcuri82:
@seran thx, see changes in other branch.

PR #1002, opened 2024-06-08
Python imports and init tests

Pgarrett:
# What

- Python test classes now contain the class name and imports.
- Test cases are created empty with the `pass` directive. This will be removed in the next PR when python tests are able to actually perform requests and assertions
- main footer added so as to be able to directly execute tests directly calling for example `python3 EvoMaster_successes_Test.py`

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
@@ -476,7 +476,6 @@ class TestSuiteWriter {
             lines.add("import json")
             lines.add("import unittest")
             lines.add("import requests")
-            createRequirementsTxt("requests==2.25.1")
         }
 
         when {
@@ -500,16 +499,6 @@ class TestSuiteWriter {
             defineClass(name, lines)
             lines.addEmpty()
         }
-    }
-
-    private fun createRequirementsTxt(testFileContent: String) {
-        val path = Paths.get(config.outputFolder, "requirements.txt")
-
-        Files.createDirectories(path.parent)
-        Files.deleteIfExists(path)
-        Files.createFile(path)
-
-        path.toFile().appendText(testFileContent)
     }
 
     private fun classFields(lines: Lines, format: OutputFormat) {
--- a/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
@@ -935,6 +935,12 @@ class TestSuiteWriter {
             lines.addEmpty(2)
             lines.add("}")
         }
+
+        if (config.outputFormat.isPython()) {
+            lines.add("if __name__ == '__main__':")
+            lines.indent()
+            lines.add("unittest.main()")
+        }
     }
 
     private fun defineClass(name: TestSuiteFileName, lines: Lines) {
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -472,7 +472,17 @@ class TestSuiteWriter {
             addUsing("EvoMaster.Controller", lines)
         }
 
-        lines.addEmpty(4)
+        if (format.isPython()) {
+            lines.add("import json")
+            lines.add("import unittest")
+            lines.add("import requests")
+            createRequirementsTxt("requests==2.25.1")
```
arcuri82, line 479: should not create an external txt file. as for Java and Koltin, it is up to user to set dependencies on (eg, see [https://github.com/EMResearch/EvoMaster/blob/master/docs/library_dependencies.md](https://github.com/EMResearch/EvoMaster/blob/master/docs/library_dependencies.md)). 
if really need to specify something here, you can add it as a comment

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -486,12 +496,22 @@ class TestSuiteWriter {
...
         }
 
-        if (format.isJavaOrKotlin() || format.isCsharp()) {
+        if (format.isJavaOrKotlin() || format.isCsharp() || format.isPython()) {
             defineClass(name, lines)
             lines.addEmpty()
         }
     }
 
+    private fun createRequirementsTxt(testFileContent: String) {
```
arcuri82, line 505: remove

core/src/test/kotlin/org/evomaster/core/output/service/TestSuiteWriterTest.kt
```diff
@@ -103,4 +98,49 @@ class TestSuiteWriterTest{
...
+        val writer = injector.getInstance(TestSuiteWriter::class.java)
+        //write the test suite
+        writer.writeTests(solution, FakeController::class.qualifiedName!!, null)
+
+        // the requirements file should exist
+        val requirementsFile = Paths.get("${config.outputFolder}/requirements.txt")
+        assertTrue(Files.exists(requirementsFile))
+
+        val testContent = String(Files.readAllBytes(requirementsFile))
+        assertTrue(testContent.contains("requests==2.25.1"))
```
arcuri82, line 132: avoid hard coding constants like `"requests==2.25.1"`. put it as constant variable in the code, and refer to the variable here instead of hard coding

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Oh, I see. Hadn't checked that library_dependencies article. Will remove this now. Regarding if need documentation for this case, requests is not included in the python standard library, but it's a widely used library, so anyone creating python tests will most likely already have the `requests` library in their python env

core/src/test/kotlin/org/evomaster/core/output/service/TestSuiteWriterTest.kt
Pgarrett: Done

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Done

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Done

arcuri82 (approved):

PR #1011, opened 2024-06-15
Most of Python functionality for tests

Pgarrett:
# What

- Python test cases are now able to perform requests and assertions
- Pending improvement of JSON assertion for future PR
- Pending set cookie for future PR
- Pending utils file with functions for future PR
- Added test checking python format
- Added tests checking JSON assertions that will need to change once assertions are improved


arcuri82 (changes requested):
@Pgarrett thx ;) see comments

Pgarrett:
@arcuri82 something odd appears to have happened when rebasing, since it brought changes from another commit... But it appears to be fine, I'm not undoing anything it seems

arcuri82 (approved):

PR #1023, opened 2024-06-28
Adding python utils file for resolve_location and check valid uri. Init expectation handling

Pgarrett:
- Adding python utils file with expectation handling, location header resolving and checking if a uri is valid or not


core/src/main/kotlin/org/evomaster/core/output/oracles/SupportedCodeOracle.kt
```diff
@@ -47,27 +48,35 @@ class SupportedCodeOracle : ImplementedOracle() {
...
-                lines.add(" Note: No supported codes appear to be defined. https://swagger.io/docs/specification/describing-responses/.")
-                lines.add(" This is somewhat unexpected, so the code below is likely to lead to a failed expectation")
-                lines.add("*/")
+                if (format.isJavaOrKotlin()) {
+                    addNoSupportedCodeComments(lines, "below")
+                }
                 when {
                     format.isJava() -> lines.add(".that($variableName, Arrays.asList().contains($name.extract().statusCode()))")
                     format.isKotlin() -> lines.add(".that($variableName, listOf<Int>().contains($name.extract().statusCode()))")
+                    format.isPython() -> lines.add(".that(self.$variableName, $name.status_code in [])")
```
Pgarrett, line 65: @arcuri82 I've got an issue here. Python does not support having a comment in between method calls, something that in java would look like:

anObject.doSomething()
// a comment
.doSomethingElse()

My first thought was not to use the fluent API and thus have every `that` call in a separate method, having for example:
`expectation_handler.expect(ems).that(swtich1, condition1)`
`expectation_handler.expect(ems).that(swtich2, condition2)`

The problem I bumped into is that the expectationsMasterSwitch is a private value in the [PartialOracles](https://github.com/WebFuzzing/EvoMaster/pull/1023/files#diff-332fc827e7d77b6da33b13013fff415eb36c504b36765a6976a36657782c791dR86) class. Do you have any suggestions?

arcuri82: @Pgarrett  hi. don't need to worry about this one... "oracles" and "expectations" are going to be removed due to a planned refactoring related to the reporting of detected faults (which will be done in the next future).
so, for time being, can just skip it (and leave a comment)

core/src/main/kotlin/org/evomaster/core/output/service/PythonUtilsFileWriter.kt
```diff
@@ -0,0 +1,122 @@
...
+            lines.add("self.master_switch = False")
+        }
+        lines.addEmpty()
+        lines.add("def expect(self, master_switch=False):")
+        lines.indented {
+            lines.add("self.master_switch = master_switch")
+            lines.add("return self")
+        }
+        lines.addEmpty()
+        lines.add("def that(self, active, condition):")
```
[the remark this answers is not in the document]
arcuri82: you can skip it

core/src/main/kotlin/org/evomaster/core/output/service/PythonUtilsFileWriter.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    private val utilsFilename = "em_test_utils.py"
+
+    fun writePythonUtilsFile(config: EMConfig) {
+        val lines = Lines(config.outputFormat)
+        addImports(lines)
+        writeResolveLocation(lines)
+        writeIsValidUriOrEmpty(lines)
+        writeExpectationHandler(lines)
+        saveToDisk(lines.toString(), config)
```
arcuri82, line 18: this feels rather cumbersome.
I think a better approach would be to have the actual implementation of `em_test_utils.py`  under `core/src/main/resources`, so that it is part of generated jar file for core process.
then, you can read its content with something like: `Main::class.java.getResource("/em_test_utils.py").readText()`

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -177,7 +177,7 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
     private fun shouldCheckExpectations() =
     //for now Expectations are only supported on the JVM
         //TODO C# (and maybe JS as well???)
-        config.expectationsActive && config.outputFormat.isJavaOrKotlin()
+        config.expectationsActive && (config.outputFormat.isJavaOrKotlin() || config.outputFormat.isPython())
```
arcuri82, line 180: skip it for python. it will be removed anyway in the near future

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/service/PythonUtilsFileWriter.kt
Pgarrett: Sure! I'll change it to that

arcuri82 (approved):

PR #1027, opened 2024-07-03
Adding support for graphQL APIs in python

Pgarrett:
This PR adds support for GraphQL APIs, adding the content type header and sending the body in the query


Pgarrett:
@arcuri82 I believe this is it for python test outputs. Or do you feel something is still missing?

core/src/main/kotlin/org/evomaster/core/output/service/GraphQLTestCaseWriter.kt
```diff
@@ -123,6 +125,19 @@ class GraphQLTestCaseWriter : HttpWsTestCaseWriter() {
             lines.append("${path?.let { GeneUtils.applyEscapes(it, mode = GeneUtils.EscapeMode.NONE, format = format) }}\"")
         }
 
+        if (format.isPython()) {
```
arcuri82, line 128: isn't this the same code as done in `RestTestCaseWriter`? if so, should be refactored in a protected utility method, moved up in the hierarchy, eg in `HttpWsTestCaseWriter`

arcuri82 (changes requested):

arcuri82:
hi @Pgarrett . possibly done, for now. we ll see once we add new set of E2E for black-box, if we get any issue there. thx

arcuri82 (approved):

PR #1041, opened 2024-07-26
Tests for utils tests in python

Pgarrett:
- Adding tests for the em_test_utils file used in test generation
- Adding CI for the em_test_utils python tests
- Parity between python and js utils tests


arcuri82:
@Pgarrett thx. I m currently on vacation abroad. I LL look at it later next week

arcuri82 (approved):

PR #1052, opened 2024-08-14
Add tests for Solver Module

agusaldasoro:
- Added test to improve code coverage over Solver Module
- Fix: Support for negative int numbers


solver/src/test/java/org/evomaster/solver/Z3DockerExecutorTest.java
```diff
@@ -7,6 +7,7 @@
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
arcuri82, line 10: shouldn't import shaded libraries from third-party dependencies

solver/src/test/java/org/evomaster/solver/Z3DockerExecutorTest.java
```diff
@@ -78,44 +78,63 @@ public void dynamicFile() throws IOException {
...
-        users2Expected.put("AGE", new IntValue(31));
-        users2Expected.put("POINTS", new IntValue(7));
-        assertEquals(new StructValue(users2Expected), response.get().get("users2"));
+    /**
+     * Test handling an empty file
+     */
+    @Test
+    public void emptyFile() {
+        try {
+            executor.solveFromFile("empty.smt");
```
arcuri82, line 135: if no exception, then test case wrongly passes? should rather use `assertThrows()`

solver/src/test/java/org/evomaster/solver/Z3DockerExecutorTest.java
```diff
@@ -78,44 +78,63 @@ public void dynamicFile() throws IOException {
...
+        } catch (Exception e) {
+            assertEquals("No result after solving file ", e.getMessage(), "Exception message should be 'error: file is empty'");
```
arcuri82, line 137: brittle check. if we change the error message, then this test fails. 
if really need to check this, the error message could be put in a static final variable, and not string duplicated here

solver/src/test/java/org/evomaster/solver/smtlib/SMTLibTest.java
```diff
@@ -0,0 +1,102 @@
+package org.evomaster.solver.smtlib;
+
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableList;
```
arcuri82, line 4: see previous comments

arcuri82 (approved):

PR #1082, opened 2024-09-21
Numbered naming strategy for test cases

Pgarrett:
- Adding `TestCaseNamingStrategy` with the `NumberedTestCaseNamingStrategy` as a default behaviour 
- Removed `customNaming` parameter in favour of the naming strategies


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -579,6 +582,10 @@ class EMConfig {
             throw ConfigProblemException("The use of 'prematureStop' is meaningful only if the stopping criterion" +
                     " 'stoppingCriterion' is based on time")
         }
+
+        if (namingStrategy == NamingStrategy.DEFAULT) {
```
arcuri82, line 586: no need for this `DEFAULT`. the default can be directly `NUMBERED`. default is needed when we have differences between configurations, eg when a default  fro black-box testing would be different then a default for white-box testing. but it is not the case here

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -981,6 +988,10 @@ class EMConfig {
             " If 0 or negative, the timeout is not applied.")
     var testTimeout = 60
 
+    @Important(2.2)
```
arcuri82, line 991: this shouldn't be among `Important`. once we decide a good strategy, that ll be default, and not something user would play with

arcuri82: as such, then this ption should be moved down in the file, after the block in which Important configs are defined

core/src/main/kotlin/org/evomaster/core/output/service/naming/NamingStrategy.kt
```diff
@@ -0,0 +1,10 @@
+package org.evomaster.core.output.service.naming
+
+enum class NamingStrategy {
+
+    DEFAULT,
```
arcuri82, line 5: can remove `DEFAULT`. see previous comments

core/src/main/kotlin/org/evomaster/core/output/service/naming/TestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,13 @@
+package org.evomaster.core.output.service.naming
+
+import org.evomaster.core.output.NamingHelper
+import org.evomaster.core.output.TestCase
+import org.evomaster.core.search.EvaluatedIndividual
+
+interface TestCaseNamingStrategy {
```
arcuri82, line 7: when declaring interfaces, then all methods need some JavaDocs. 
also, if these methods require solution to be pass in constructor, and not as input parameter, then maybe to consider to have this class rather as abstract, with solution saved in a protected field

core/src/main/kotlin/org/evomaster/core/output/service/naming/TestCaseNamingStrategyFactory.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class TestCaseNamingStrategyFactory(
+    private val namingStrategy: NamingStrategy
+) {
+
+    fun create(solution: Solution<*>): TestCaseNamingStrategy {
+        if (namingStrategy.isNumbered()) {
+            return NumberedTestCaseNamingStrategy(solution)
+        }
+        return NumberedTestCaseNamingStrategy(solution)
```
arcuri82, line 13: if it is not recognized, should rather throw an IllegalStateExcecption

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
Pgarrett: Ok! Thought we were going to allow users to choose if they wanted their tests to be named with the numbered or action strategy, beyond the thesis testing scope

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
Pgarrett: Won't we be using the choice of naming strategies for testing purposes?

core/src/main/kotlin/org/evomaster/core/output/service/naming/TestCaseNamingStrategy.kt
Pgarrett: Will change to an abstract class then, makes more sense

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
Pgarrett: (moved it to the other config, had misunderstood)

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
Pgarrett: Also misunderstood

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1420,9 +1423,6 @@ class EMConfig {
     @Min(-1.0)
     var maxLengthOfTraces = 10
 
-    @Cfg("Enable custom naming and sorting criteria")
```
arcuri82, line 1423: instead of removing it, to avoid breaking clients that use it, rather put it to false, and mark it with a `@Deprecated("No longer in use")`

core/src/main/kotlin/org/evomaster/core/output/service/naming/NamingStrategy.kt
```diff
@@ -0,0 +1,9 @@
+package org.evomaster.core.output.service.naming
```
arcuri82, line 1: `.service.` packages have a special meaning, altough it is not enforced in the build. see `docs/for_developers.md`. rather use for example `org.evomaster.core.output.naming`

core/src/main/kotlin/org/evomaster/core/output/service/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,26 @@
+package org.evomaster.core.output.service.naming
```
arcuri82, line 1: see previous comment. unless this class is a service bean injectable with Guice, shouldn't be uner `.service.` package

core/src/main/kotlin/org/evomaster/core/output/service/naming/TestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,16 @@
+package org.evomaster.core.output.service.naming
+
+import org.evomaster.core.output.NamingHelper
+import org.evomaster.core.output.TestCase
+import org.evomaster.core.search.EvaluatedIndividual
+import org.evomaster.core.search.Solution
+
+abstract class TestCaseNamingStrategy(
```
arcuri82, line 8: for abstract class and interfaces, add JavaDocs for each method

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
Pgarrett: Done

core/src/main/kotlin/org/evomaster/core/output/service/naming/NamingStrategy.kt
Pgarrett: Done

core/src/main/kotlin/org/evomaster/core/output/service/naming/TestCaseNamingStrategy.kt
Pgarrett: Added

arcuri82 (approved):

PR #1085, opened 2024-09-27
Kickoff Action Naming for RestIndividuals

Pgarrett:
- Usage of `namingHelper` moved to a specific `NamingHelperNumberedTestCaseNamingStrategy`. The reason for this is to preserve the original naming strategy which used the naming helper for suffixes in cases where a 500 error was returned. The action based naming strategy should not need this since faults and status codes must be part of the naming strategy itself
- `NumberedTestCaseNamingStrategy` is kept as an `open` and not `abstract` class so as to allow the naming just with numbers. This is used in the fallback mechanism when any issue arises regarding sorting, as it used to behave.
- Since just a minimal implementation for RestIndividual action based naming exists, GraphQL and RPC test naming will default to the numbered strategy even though action based was chosen. This will change in the forthcoming PRs, naturally.


core/src/main/kotlin/org/evomaster/core/output/TestSuiteOrganizer.kt
```diff
@@ -228,8 +227,7 @@ class SortingHelper {
     /**
      *Sorting is done according to the comparator list. If no list is provided, individuals are sorted by max status.
      */
-    private fun sortByComparatorList (namingHelper: NamingHelper,
-                              comparators: List<Comparator<EvaluatedIndividual<*>>> = listOf(statusCode),
+    private fun sortByComparatorList (comparators: List<Comparator<EvaluatedIndividual<*>>> = listOf(statusCode),
```
Pgarrett, line 230: @arcuri82, I kept this just even though it just calls the naming strategy since there's quite an extensive comment explaining sorting and we're not modifying sorting yet. I'm guessing we'll get naming done first and then focus in the sorting bit

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,27 @@
...
+open class ActionTestCaseNamingStrategy(
+    solution: Solution<*>
+) : NumberedTestCaseNamingStrategy(solution)  {
+
+
+    override fun expandName(individual: EvaluatedIndividual<*>): String {
+        var evaluatedAction = individual.evaluatedMainActions().last()
+        var action = evaluatedAction.action as RestCallAction
+        var result = evaluatedAction.result as HttpWsCallResult
+        return "_${action.verb}_on_${sanitizePathName(action.path.nameQualifier)}_returns_${result.getStatusCode()}"
```
Pgarrett, line 17: @arcuri82 this first implementation only focuses in the verb, path and status code for REST individuals. Besides these three elements, which other aspects of a response do we wish to contemplate? 
Top of mind I'm thinking of an array response where we could add the `returns_empty`, `returns_2elements` instead of the status code. Not sure how easy it is to access that information from the individual, we hadn't discussed it in the original doc.

arcuri82: an important aspect is to check if there is any fault, and, if so, put its code

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    private fun sanitizePathName(nameQualifier: String): String {
```
arcuri82, line 20: this type of function should be in a utlity class, like for example `TestWriterUtils`.
Also there is already something that could be used, eg, `safeVariableName()`

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
arcuri82: you can use `DetectedFaultUtils.getDetectedFaultCategories` to get such info

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
Pgarrett: Was actually thinking of adding that later to have more _to the point_ PRs, but sure. I'll add the field to represent the label, as we had originally discussed in the document.

Regarding these modifiers for the test name, do we wish to only name test cases using:
- status code for successful tests
- Fault information for tests containing faults
? Or do we wish to use extra metadata like in the first question? Maybe it can be something to evaluate later on

Pgarrett: How do we wish to handle a test case that has multiple faults? Currently there's no hierarchy, could the `FaultCategory.code` serve that purpose?

wfc/src/main/java/com/webfuzzing/commons/faults/FaultCategory.java
```diff
@@ -72,4 +81,14 @@ public enum FaultCategory {
...
         return "F" + code + ":" + name;
     }
+
+    /**
+     * A camelCase label for this category, to be used in test case naming.
+     * This will enable tests to have more descriptive names and easy detection of tests with faults
+     *
+     * @return the test case label
+     */
+    public String getTestCaseLabel() {
```
arcuri82, line 91: as this is an enumeration, and the fields are public final, there is no need for a direct getter

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    private fun addResult(individual: EvaluatedIndividual<*>): String {
+        val detectedFaults = DetectedFaultUtils.getDetectedFaultCategories(individual)
+        if (detectedFaults.isNotEmpty()) {
+            return fault(detectedFaults)
+        }
+        return statusCode(individual.evaluatedMainActions().last())
+    }
+
+    private fun fault(faults: Set<FaultCategory>): String {
```
arcuri82, line 39: if it is only 1, then yes, using `testCaseLabel` makes sense.
but, if it is more than 1 in the test (and not just last call), then maybe using sorted list of unique `code` could be better, eg something like `Faults_100_101_400_500_801`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
arcuri82: @Pgarrett  yes, use unique sorted codes. see comments below

arcuri82 (approved):

PR #1086, opened 2024-09-27
new statistics fields in statistics.csv file

jgaleotti:
averageNumberOfEvaluatedRowsForSqlHeuristics, sqlHeuristicsEvaluationFailures, 	sqlHeuristicsEvaluationCount, averageNumberOfEvaluatedDocumentsForMongoHeuristics	and mongoHeuristicsEvaluationCount


core/src/main/kotlin/org/evomaster/core/search/service/AverageCalculator.kt
```diff
@@ -0,0 +1,27 @@
+package org.evomaster.core.search.service
+
+class AverageCalculator {
```
arcuri82, line 3: why this class?  wouldn't `org.evomaster.core.utils.IncrementalAverage` handle this?
also, as this is not an injectable service, it should not be under a `*.service` package.

core/src/test/kotlin/org/evomaster/core/search/service/AverageCalculatorTest.kt
```diff
@@ -0,0 +1,53 @@
+package org.evomaster.core.search.service
+
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+class AverageCalculatorTest {
+
+    @Test
+    fun testAddAndGetAverageWithMultipleValues() {
```
arcuri82, line 9: could be merged into `IncrementalAverageTest`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/service/AverageCalculator.kt
jgaleotti: I didn't know about this class. I will replace it with IncrementalAverage.

PR #1087, opened 2024-09-30
refactor_sql_metrics

jgaleotti:
Several non functional changes to improve readibility plus statistics on number of SQLExceptions thrown while executing SQL commands


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExecutedSqlCommand.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Info related to SQL command execution.
+ *
+ * This class is IMMUTABLE
+ */
+public class ExecutedSqlCommand implements Serializable {
+
+
+    public static transient long FAILURE_EXECUTION_TIME = -1L;
```
arcuri82, line 14: what is this for? a mutable static variable is dangerous... if really needed, should be under `org.evomaster.client.java.instrumentation.staticstate`

arcuri82: or was it supposed to be `final` and not `transient`???

arcuri82: either way, should add some JavaDoc for it

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlParserUtils.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * check if the sql is `Select 1`
+     * detected by proxyprint as
+     *      ERROR - FAILED TO COMPUTE HEURISTICS FOR SQL: SELECT 1
+     *
+     * https://stackoverflow.com/questions/3668506/efficient-sql-test-query-or-validation-query-that-will-work-across-all-or-most
+     */
+    public static boolean isSelectOne(String sqlCommand) {
+        return sqlCommand!= null && sqlCommand.trim().toLowerCase().matches("select\\s+1\\s*;?");
```
arcuri82, line 53: wouldn't this work for any number, and not just `1`? ie, have a more generic `select\\s+-?\\d+\\s*;?`

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/ExecutedSqlCommand.java
jgaleotti: I think it was supposed to be final instead of transient. I will add some javadoc for it.

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlParserUtils.java
jgaleotti: makes sense

arcuri82 (approved):

PR #1095, opened 2024-10-08
Adding Init actions to test naming. Rest tests now also include details about the json payload in the response

Pgarrett:
- For SQL, MongoDB and WireMock usages in tests, they are now included in the test name
- For REST test cases, when a GET response receives a 200 json response, the body is parsed to have better naming:
  - `[]` returns empty list
  - `[1]` returns 1 element
  - `[1,2,3]` returns 3 elements
  - `{}` returns empty object
  - `{"key": "value"}` returns object
  - `"myItem"` returns string
  - `<xml/>` returns content


core/src/test/kotlin/org/evomaster/core/output/naming/RestActionNamingStrategyTest.kt
```diff
@@ -92,34 +211,166 @@ class RestActionNamingStrategyTest {
...
+
+        val namingStrategy = RestActionTestCaseNamingStrategy(solution, javaFormatter, EMConfig())
+
+        val testCases = namingStrategy.getTestCases()
+        assertEquals(1, testCases.size)
+        assertEquals("test_0_getOnItemsReturnsEmptyUsingSql", testCases[0].name)
+    }
+
+    //    @Test// TODO check PhG Gene#isLocallyValid with JP/AA
+    fun testIndividualUsingMongo() {
```
Pgarrett, line 239: @arcuri82 @jgaleotti 
I'm having trouble with this test case. When executing I get the following exception:
```
java.lang.AssertionError: Assertion failed

	at org.evomaster.core.Lazy.assert(Lazy.kt:16)
	at org.evomaster.core.search.gene.Gene.doInitialize(Gene.kt:176)
	at org.evomaster.core.search.action.Action.doInitialize(Action.kt:75)
	at org.evomaster.core.search.Individual.doInitialize(Individual.kt:198)
	at org.evomaster.core.TestUtils.doInitializeIndividualForTesting(TestUtils.kt:30)
	at org.evomaster.core.TestUtils.doInitializeIndividualForTesting$default(TestUtils.kt:28)
	at org.evomaster.core.output.naming.RestActionNamingStrategyTest.getEvaluatedIndividualWithFaults(RestActionNamingStrategyTest.kt:337)
	at org.evomaster.core.output.naming.RestActionNamingStrategyTest.getEvaluatedIndividualWith(RestActionNamingStrategyTest.kt:294)
	at org.evomaster.core.output.naming.RestActionNamingStrategyTest.getEvaluatedIndividualWith(RestActionNamingStrategyTest.kt:286)
	at org.evomaster.core.output.naming.RestActionNamingStrategyTest.getEvaluatedIndividualWith$default(RestActionNamingStrategyTest.kt:285)
	at org.evomaster.core.output.naming.RestActionNamingStrategyTest.testIndividualUsingMongo(RestActionNamingStrategyTest.kt:241)
```

Which apparently has something to do with how genes are initialized. The `isLocallyValid` call fails for some reason I can't understand.
Any idea what I could look into to fix it?

arcuri82:
hi @Pgarrett  sorry I am bit busy these days, likely until Tuesday :(  @jgaleotti  you think you can have a first review of this PR? (I ll do mine then afterwards)

Pgarrett:
Don't worry @arcuri82! I'll check with JP tomorrow, thanks!

Pgarrett:
Hi @arcuri82! Test fixed, got help from @jgaleotti. I had missed passing a `Randomness` for test initialization, how shameful 🤦 
Setting the PR as ready for review

Pgarrett:
Hi @arcuri82! Were you able to take a look at this PR?

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -51,6 +62,34 @@ abstract class ActionTestCaseNamingStrategy(
...
+    private fun hasSqlAction(actions: List<EnvironmentAction>): Boolean {
+        return actions.any { it is SqlAction }
+    }
+
+    private fun hasMongoAction(actions: List<EnvironmentAction>): Boolean {
+        return actions.any { it is MongoDbAction }
+    }
+
+    private fun usesWireMock(actions: List<Action>): Boolean {
+        return config.isEnabledExternalServiceMocking() && actions.any { it is HostnameResolutionAction }
```
arcuri82, line 92: there are 2 different types of actions for handling external services `HostnameResolutionAction` and `HttpExternalServiceAction`. the latter is related to WireMock.

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
Pgarrett: My bad, fixed!

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -51,6 +62,34 @@ abstract class ActionTestCaseNamingStrategy(
...
+    private fun hasSqlAction(actions: List<EnvironmentAction>): Boolean {
+        return actions.any { it is SqlAction }
+    }
+
+    private fun hasMongoAction(actions: List<EnvironmentAction>): Boolean {
+        return actions.any { it is MongoDbAction }
+    }
+
+    private fun usesWireMock(actions: List<Action>): Boolean {
+        return config.isEnabledExternalServiceMocking() && actions.any { it is HttpExternalServiceAction }
```
arcuri82, line 92: hi. just noticed it... but, why the `config.isEnabledExternalServiceMocking()`?  if there is any action of type `HttpExternalServiceAction`, then it is implied that WM support is enabled.  isn't  `config.isEnabledExternalServiceMocking()` redundant?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
Pgarrett: @arcuri82, took that from the TestSuiteWriter file here: https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt#L397

Which is, now that I remember, from where I also took the usage of the `HostnameResolutionAction`, since the `needWireMockServers` checks for that: https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/search/Solution.kt#L46

So, if it's redundant, I'll remove it, not very sure from my end

arcuri82: yep, I think it is redundant.  `config.isEnabledExternalServiceMocking()` would imply that WM can be used... but that does not mean that necessarely a test is using it. from a logical perspective, we have a case of `A && B`, where `notA => notB` and `B => A`. so `A` is redundant when checking `A && B` (isn't it?)

Pgarrett: Fair enough, yes. From a logical perspective I fully agree. My next two questions are:
1. Why does `TestSuiteWriter` check both conditions?
2. Why does `Solution.needWireMockServers` use `HostnameResolutionAction` instead of `HttpExternalServiceAction`?

arcuri82: regarding [TestSuiteWriter.kt#L397](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt#L397), those are import statements at the beginning of the test suite. you just need 1 test in whole suite to require executing that code. Indeed `config.isEnabledExternalServiceMocking()` could be considered redundant here, but, it is a cheaper check then looking at whole `solution` (so, if false, it is quicker to compute)

Regarding `solution.needWireMockServers()` might indeed rather check `HttpExternalServiceAction`. that might be old check from time before we had a sink address.  but worst case it just adds imports that are not used. 
@seran  can you comment on this?

thx

Pgarrett: Got it, I'll remove the `config.isEnabledExternalServiceMocking()` and wait for the comments regarding `HttpExternalServiceAction`

seran: 1. `isEnabledExternalServiceMocking()` is used in several other places as a setting to enable or disable functions related to external service handling.
2. There could be cases where we only need to reroute the external service call to the network sinkhole. In such instances, we use `HostnameResolutionAction` to configure the DNS information, and there will be no `HttpExternalServiceAction `.
3. If the test acts upon the response from the external service, then we use the `HttpExternalServiceAction` to configure the stubs, along with `HostnameResolutionAction` to configure DNS information.

As the name suggests, if your method only needs to check for WM usage, then the current logic seems fine without the `isEnabledExternalServiceMocking()` check.

You can assume that if there is a `HttpExternalServiceAction`, there is a `HostnameResolutionAction ` but not vice versa.
Also, all of this is will happen if the `isEnabledExternalServiceMocking()` returns `true.`

@arcuri82 we need to do a clean-up in other places where we have the same check as `isEnabledExternalServiceMocking()`.

Pgarrett: Understood, thanks @seran for the explanation! I believe then we're good to go with the review @arcuri82, I'll leave the `HttpExternalServiceAction` as the check for WM

arcuri82 (approved):

PR #1099, opened 2024-10-16
feat: adding opcode to end of the branch identifier

omursahin:


arcuri82:
hi @omursahin . thanks. when making a PR, you need to wait that CI passes before asking my review :)  if CI fails (eg due to tests failing), you need to fix them first. 
once fixed and CI is green, then you can write a message here tagging me @arcuri82 , so I get notification that the PR is ready for my review ;)

omursahin:
I think It is OK this time @arcuri82

arcuri82 (approved):

PR #1108, opened 2024-10-23
renaming of the stopping criterion 

omursahin:


core/src/main/kotlin/org/evomaster/core/docs/ConfigToMarkdown.kt
```diff
@@ -156,7 +157,7 @@ object ConfigToMarkdown {
         val typeName = if(type.isEnum){
             "Enum"
         } else {
-            type.simpleName.capitalize()
+            type.simpleName.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.ENGLISH) else it.toString() }
```
arcuri82, line 160: why this change?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/docs/ConfigToMarkdown.kt
omursahin: Since I changed the parameter name, I had to generate Markdown again. When I regenerated Markdown, the tests ran correctly on my computer but gave an error in the CI process (https://github.com/WebFuzzing/EvoMaster/actions/runs/11477931721/job/31942213194). Later, I realized that since it capitalized the "ı" character as "İ", the "Int" type was written as "İnt" in the file. The tests pass/fail depending on the computer's default language. With this change, I aimed to remove the dependency on the language. 
Btw, the capitalize function appeared to be deprecated so the IDE suggested me to use the replaceFirstChar function instead.

image

omursahin: I think this time It is OK.

arcuri82 (approved):

PR #1121, opened 2024-10-30
Path disambiguation

Pgarrett:
# What
For REST individuals, whenever two individuals end in the same path, we go up one path level to add disambiguation.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
@@ -17,7 +17,7 @@ open class GraphQLActionTestCaseNamingStrategy(
         val evaluatedAction = individual.evaluatedMainActions().last()
         val action = evaluatedAction.action as GraphQLAction
 
-        nameTokens.add(action.methodType.toString())
+        nameTokens.add(action.methodType.toString().lowercase())
         nameTokens.add(on)
         nameTokens.add(getPath(action.methodName))
         addResult(individual, nameTokens)
@@ -25,8 +25,9 @@ open class GraphQLActionTestCaseNamingStrategy(
         return formatName(nameTokens)
     }
 
-    override fun resolveAmbiguity(individualToName: MutableMap<EvaluatedIndividual<*>, String>, duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>) {
-        // do nothing at the moment. This will be completed with the experimental params disambiguation method
+    override fun resolveAmbiguities(duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>): Map<EvaluatedIndividual<*>, String> {
+        // TODO do nothing at the moment. This will be completed with the experimental params disambiguation method
+        return emptyMap()
     }
 
     override fun addActionResult(evaluatedAction: EvaluatedAction, nameTokens: MutableList<String>) {
--- a/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
@@ -25,7 +25,7 @@ open class GraphQLActionTestCaseNamingStrategy(
         return formatName(nameTokens)
     }
 
-    override fun resolveAmbiguity(individualToName: MutableMap<EvaluatedIndividual<*>, String>, inds: MutableSet<EvaluatedIndividual<*>>) {
+    override fun resolveAmbiguity(individualToName: MutableMap<EvaluatedIndividual<*>, String>, duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>) {
         // do nothing at the moment. This will be completed with the experimental params disambiguation method
     }
 
--- a/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
@@ -4,6 +4,7 @@ import org.evomaster.core.problem.graphql.GraphQLAction
 import org.evomaster.core.problem.graphql.GraphQlCallResult
 import org.evomaster.core.search.EvaluatedIndividual
 import org.evomaster.core.search.Solution
+import org.evomaster.core.search.action.Action
 import org.evomaster.core.search.action.EvaluatedAction
 
 open class GraphQLActionTestCaseNamingStrategy(
@@ -12,7 +13,7 @@ open class GraphQLActionTestCaseNamingStrategy(
 ) : ActionTestCaseNamingStrategy(solution, languageConventionFormatter)  {
 
 
-    override fun expandName(individual: EvaluatedIndividual<*>, nameTokens: MutableList<String>): String {
+    override fun expandName(individual: EvaluatedIndividual<*>, nameTokens: MutableList<String>, ambiguitySolver: ((Action) -> List<String>)?): String {
         val evaluatedAction = individual.evaluatedMainActions().last()
         val action = evaluatedAction.action as GraphQLAction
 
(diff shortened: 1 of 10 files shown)


Pgarrett:
@arcuri82 I've decided to split the PRs regarding the disambiguation feature, so as to have a more to the point PR. Later on I'll send the one with the experimental feature of disambiguation using query/uri params. Grounds for that feature are naturally set here

core/src/main/kotlin/org/evomaster/core/output/naming/GraphQLActionTestCaseNamingStrategy.kt
```diff
@@ -24,6 +25,10 @@ open class GraphQLActionTestCaseNamingStrategy(
         return formatName(nameTokens)
     }
 
+    override fun resolveAmbiguity(individualToName: MutableMap<EvaluatedIndividual<*>, String>, duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>) {
+        // do nothing at the moment. This will be completed with the experimental params disambiguation method
+    }
```
arcuri82, line 30: when comments like this, have a `TODO` in the comment. `TODO` tags are specially handled by IntelliJ and Git

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -21,18 +22,51 @@ open class NumberedTestCaseNamingStrategy(
     }
 
     // numbered strategy will not expand the name unless it is using the namingHelper
-    override fun expandName(individual: EvaluatedIndividual<*>, nameTokens: MutableList<String>): String {
+    override fun expandName(individual: EvaluatedIndividual<*>, nameTokens: MutableList<String>, ambiguitySolver: ((Action) -> List<String>)?): String {
         return ""
     }
 
+    override fun resolveAmbiguity(individualToName: MutableMap<EvaluatedIndividual<*>, String>, duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>) {
+        // do nothing, plain numbered strategy will never have duplicate names
```
arcuri82, line 30: add `TODO`

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -21,18 +22,51 @@ open class NumberedTestCaseNamingStrategy(
...
     private fun getName(counter: Int, individual: EvaluatedIndividual<*>): String {
         return "test_${counter}${expandName(individual, mutableListOf())}"
     }
 
+    private fun concatName(counter: Int, expandedName: String): String {
+        return "test_${counter}${expandedName}"
+    }
+
     private fun generateNames(individuals: List<EvaluatedIndividual<*>>) : List<TestCase> {
+        val individualToName = mutableMapOf<EvaluatedIndividual<*>, String>()
```
arcuri82, line 43: should not have mutable objects (eg `EvaluatedIndividual<*>`) as keys for a map. if for any reason a key is modified, then the map can become unusable

arcuri82: maybe better something like:
```
val individualToName = individuals.map{Pair(it, expandName(it))}
```

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -21,18 +22,51 @@ open class NumberedTestCaseNamingStrategy(
...
+
+        val duplicatedNames = getDuplicateNames(individualToName)
+        duplicatedNames.forEach { entry -> resolveAmbiguity(individualToName, entry.value) }
+
         var counter = 0
-        return individuals.map { ind -> TestCase(ind, getName(counter++, ind)) }
+        return individualToName.map { entry -> TestCase(entry.key, concatName(counter++, entry.value)) }
+    }
+
+    private fun getDuplicateNames(individualToName: MutableMap<EvaluatedIndividual<*>, String>): MutableMap<String, MutableSet<EvaluatedIndividual<*>>> {
```
arcuri82, line 56: if `individualToName` is not modified, should not be passed as a `MutableMap`. also, it would be weird to have side-effect modifications on inputs in a function called `get...`

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -21,18 +22,51 @@ open class NumberedTestCaseNamingStrategy(
...
+        val result = mutableMapOf<String, MutableSet<EvaluatedIndividual<*>>>()
+        individualToName.forEach { entry ->
```
arcuri82, line 58: this can be simplified with `.map` and `.filter` operations.
as a rule of thumb, should avoid a `forEach` that is filling a mutable list created just for it

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -37,6 +43,66 @@ open class RestActionTestCaseNamingStrategy(
         }
     }
 
+    /**
+     * In REST Individuals. Ambiguity will be resolved with the path filter.
```
arcuri82, line 47: `In REST Individuals, ambiguity`

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -37,6 +43,66 @@ open class RestActionTestCaseNamingStrategy(
...
+            var path = (it.evaluatedMainActions().last().action as RestCallAction).path
+            if (path.isLastElementAParameter()) {
+                path = path.parentPath()
+            }
+            val toStringPath = path.toString()
+            val isLastAParam = path.isLastElementAParameter()
+            Pair(toStringPath, isLastAParam)
+        }.filter { it.value.size == 1 && !it.key.second }
+
+        groupByPath.forEach { entry ->
```
arcuri82, line 74: why introducing local variable `groupByPath` and not chain the `forEach` directly after the filter?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: To be honest, readability and debugging purposes at the time

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
Pgarrett: I understand, but the fact is that I need to have a way to bind the `EvaluatedIndividual` with the generated test case name for the latter to be replaced if it needs to be disambiguated. So, eventually I'll need something along the lines of a Map.
The latest commit, does not forward the `individualToName` to the following method, but just modifies it in the scope of the `generateNames` method. Maybe that's a better solution.

I see no reason for an `EvaluatedIndividual` to be modified while the test case names are being generated. That said, a `Lazy.assert { individuals.size == individualToName.size }` could be added just after the `getDuplicateNames(individualToName).forEach ...` call. So as to be sure nobody in the future changes an EvaluatedIndividual and therefore adds an extra element to the map. What do you think?

arcuri82: @Pgarrett : i guess it can be ok here, as the map is not exposed (eg returned from a public method). 
as a rule of thumb, mutable keys should be avoided, and their map never exposed. 
debugging weird bugs related to a corrupted map (eg buckets based on hascode indices, but hashcode of stored elements has changed because their state has changed) is no fun... it is the kind of a bug that is really hard to track down :(

arcuri82 (approved):

PR #1127, opened 2024-11-08
Phg/query params disamb

Pgarrett:
- Adding an extra layer for test naming differentiation for REST scenarios. It takes into account the presence of query params, but not the values. The latter ones will be added for the experimental features, but setting the grounds here.
- Test disambiguation now takes place until no change has been detected. This will happen since the check runs against the size of `duplicatedIndividuals: MutableSet`, either no more disambiguation takes place (no individuals are removed from the set) or eventually all individuals are disambigauted (the set size is now 0 and the next iteration will have the same size)

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
@@ -2,10 +2,12 @@ package org.evomaster.core.output.naming
 
 import com.google.gson.Gson
 import com.google.gson.JsonSyntaxException
+import org.evomaster.core.output.TestWriterUtils.safeVariableName
 import org.evomaster.core.problem.rest.HttpVerb
 import org.evomaster.core.problem.rest.RestCallAction
 import org.evomaster.core.problem.rest.RestCallResult
 import org.evomaster.core.problem.rest.RestPath
+import org.evomaster.core.problem.rest.param.QueryParam
 import org.evomaster.core.search.EvaluatedIndividual
 import org.evomaster.core.search.Solution
 import org.evomaster.core.search.action.Action
@@ -51,7 +53,10 @@ open class RestActionTestCaseNamingStrategy(
      * following solvers.
      */
     override fun resolveAmbiguities(duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>): Map<EvaluatedIndividual<*>, String> {
-        return checkForPath(duplicatedIndividuals)
+        val solvedAmbiguities = mutableMapOf<EvaluatedIndividual<*>, String>()
+        solvedAmbiguities.putAll(checkForPath(duplicatedIndividuals))
+        solvedAmbiguities.putAll(checkForQueryParams(duplicatedIndividuals))
+        return solvedAmbiguities
     }
 
     /*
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -105,6 +109,39 @@ open class RestActionTestCaseNamingStrategy(
         return if (parentPathQualifier == "/") "" else getPath(parentPathQualifier)
     }
 
+    /*
+     * If any test uses query prams, then add withQueryParam(s) after the path to them to make them differ.
```
arcuri82, line 113: `prams -> params`?

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -105,6 +109,39 @@ open class RestActionTestCaseNamingStrategy(
...
+     * The filter call ensures that we are only performing this disambiguation when there's only one individual that
+     * differs and the list of query params is not empty.
+     */
+    private fun checkForQueryParams(duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>): Map<EvaluatedIndividual<*>, String> {
```
arcuri82, line 117: i guess could be specified `<RestIndividual` instead of `<*>`?

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -105,6 +109,39 @@ open class RestActionTestCaseNamingStrategy(
...
+        return duplicatedIndividuals
+            .groupBy {
+                (it.evaluatedMainActions().last().action as RestCallAction).parameters.filterIsInstance<QueryParam>()
+            }
+            .filter { it.value.size == 1 && it.key.isNotEmpty()}
+            .mapNotNull { entry ->
+                val eInd = entry.value[0]
+                duplicatedIndividuals.remove(eInd)
```
arcuri82, line 125: why this? this is a side effect on the input list. quite confusing, considering that the name of the function is `check`. also, this is the same datastructure used on line 118 at the beginning of stream operations (althugh these ones are eagerly evaluated), which complicates the understanding of the function. this needs refactoring/simplification

arcuri82: as rule of thumb, side effects on inputs data structures should be avoided, unless explicitely clarified in the method name (somehow)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: Yep

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: Tried doing so, the Kotlin type system didn't like it much

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: Just in case, it follows the same pattern as the path disambiguation in the previous [PR](https://github.com/WebFuzzing/EvoMaster/pull/1121/files#diff-1635090843debc863feecb95a31a82cbb08c67fd5b7a66b03582062b01405d8dR77). It's also clarified in the [Javadoc](https://github.com/WebFuzzing/EvoMaster/pull/1127/files#diff-1635090843debc863feecb95a31a82cbb08c67fd5b7a66b03582062b01405d8dR51-R52) of the top method.

Still, I'll work on refactor, simplification and naming

Pgarrett: I've just pushed a draft on a possible solution. It transfers the responsibility of removing the solved duplicates to a specific function.
The downside is the duplicate removal happens twice:
1. in the `resolveAmbiguities` function where the local working copy has the resolved duplicates removed
2. in the outer loop which triggers the resolution in [NumberedTestCaseNamingStrategy](https://github.com/WebFuzzing/EvoMaster/pull/1127/files#diff-15b9fc3f018892c963e1bdab6cf898d1e32cfe9a4c3755b4249a025beca324cbR55-R57).

However, this downside could be an advantage if we decide that all ambiguity solvers should be executed for resolution. It could even be tested for the experimentation: "do you like this naming where it only solves by one particular parameter or do you want the whole enchilada?".

Let me know what you think. Once we decide on this I'll also extract the `groupBy` sections of the solvers to more clearly named methods

arcuri82 (approved):

PR #1128, opened 2024-11-11
Collecting data from random walk algorithm

omursahin:


core/src/main/kotlin/org/evomaster/core/search/service/FitnessFunction.kt
```diff
@@ -138,12 +138,14 @@ abstract class FitnessFunction<T>  where T : Individual {
         actionsSize: Int
     ) : EvaluatedIndividual<T>?{
 
+        val allTargetsWithDescriptive = config.processFormat == EMConfig.ProcessDataFormat.TARGET_HEURISTIC
```
arcuri82, line 141: add some comments here, to explain "why" this is done. ie, by default we optimize for performance in collecting coverage values, but for special cases we want to collect full info

arcuri82 (changes requested):

arcuri82 (approved):

Issue #964, opened 2024-04-25
ERROR - Failed to extract the SQL Database Schema: Cannot handle repeated column sid in table addressable


aruvic:
**Below error is produced by JDK17 Spring Boot driver using EvoMaster 3.0.0**
```
ERROR - Failed to extract the SQL Database Schema: Cannot handle repeated column sid in table addressable
java.lang.IllegalArgumentException: Cannot handle repeated column sid in table addressable
        at org.evomaster.client.java.sql.SchemaExtractor.handleTableEntry(SchemaExtractor.java:575)
        at org.evomaster.client.java.sql.SchemaExtractor.extract(SchemaExtractor.java:196)
        at org.evomaster.client.java.controller.internal.SutController.getSqlDatabaseSchema(SutController.java:639)
        at org.evomaster.client.java.controller.internal.SutController.initSqlHandler(SutController.java:296)
        at org.evomaster.client.java.controller.internal.EMController.lambda$runSut$20(EMController.java:372)
        at org.evomaster.client.java.controller.internal.EMController.noKillSwitch(EMController.java:126)
        at org.evomaster.client.java.controller.internal.EMController.runSut(EMController.java:372)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at shaded.org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52)
        at shaded.org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:124)
        at shaded.org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:167)
        at shaded.org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:176)
[...]

Issue #806, opened 2023-09-20
Support for systems with different database schemas


victorhugofr:
Hi @arcuri82,

I'm facing this error when trying to generate whitebox test cases with EvoMaster:

"ERROR - Failed to extract the SQL Database Schema: null java.lang.NullPointerException"

In my analysis, this happens because EvoMaster is only prepared to read tables from the database that are located in the "public" schema. In the system I'm trying to run, the entities are in various schemas, causing it to attempt to find a table in another schema within the "public" schema.

![image](https://github.com/EMResearch/EvoMaster/assets/32314062/f9b8cbe3-0b12-431c-8eb6-6cbcf8c62837)

Is there any support for different schemas? Is it already on the radar? Or could it be some kind of misconfiguration?

Version of EM: 1.5.0
Running from JAR
Command-line options: java -jar evomaster.jar --outputFolder src/test/java
Database: PostgreSQL 14

arcuri82:
Hi @victorhugofr ,
 thanks for reporting this issue. I am not an expert in PostgreSQL, so it might be a misunderstanding on our part.
Can you provide some more details? 

For example, how many different schemas you have and their name? are they all accessed with the same JDBC connection? How is the database structured / initialized? 

is `connection.getSchema()` returning `null` here? or throwing an exception?

is the API you working on open-source, or closed-source? 

thanks

victorhugofr:
> Hi @victorhugofr , thanks for reporting this issue. I am not an expert in PostgreSQL, so it might be a misunderstanding on our part. Can you provide some more details?
> 
> For example, how many different schemas you have and their name? are they all accessed with the same JDBC connection? How is the database structured / initialized?
> 
> is `connection.getSchema()` returning `null` here? or throwing an exception?
> 
> is the API you working on open-source, or closed-source?
> 
> thanks

Thanks for the reply,
The system has 5 schemas, as shown below:
![image](https://github.com/EMResearch/EvoMaster/assets/32314062/6d82595c-a647-463b-82df-feeb1711d6c6)


They are automatically generated when the system is started with Hibernate/JDBC.

The API is closed-source (but I have access to the code).

_Some important notes:_
When I tried to run EvoMaster, the system returned an error indicating that it couldn't find the database.  It is was launched in a new Docker environment, so the database didn't exist. **Solution**: after starting http://localhost:40100/controller/api, I entered the Docker container and ran the script "
```
psql -U postgres;" CREATE DATABASE databasename; \c databasename;
[...]

PR #1130, opened 2024-11-13
starting with multi-dbs support

arcuri82:


arcuri82:
needed for #964 and  #806

PR #1138, opened 2024-11-22
True boolean query parameters are added to rest test case names experimentation

Pgarrett:
# What + Why
Adding an experimental parameter for using the names of boolean query parameters which have a `true` value in test case disambiguation.
When there's more than one eligible query param, they are joined by the `and` keyword for readability purposes

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -475,10 +475,6 @@ class EMConfig {
             throw ConfigProblemException("Python output is used only for black-box testing")
         }
 
-        if (namingStrategy == NamingStrategy.NUMBERED && nameWithQueryParameters) {
-            throw ConfigProblemException("Test case naming with Query Parameters is only allowed for Action based naming strategy")
-        }
-
         when (stoppingCriterion) {
             StoppingCriterion.TIME -> if (maxEvaluations != defaultMaxEvaluations) {
                 throw ConfigProblemException("Changing number of max actions, but stopping criterion is time")
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -475,6 +475,10 @@ class EMConfig {
             throw ConfigProblemException("Python output is used only for black-box testing")
         }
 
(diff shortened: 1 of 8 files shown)


Pgarrett:
Hi @arcuri82, I've been thinking. It seems to me that these two questions we went through the other day are rather similar:
1. Could you provide extra information that would help naming?
2. Would you name it differently? If so, how?

Maybe we could merge them and have another which asks about String or number query parameters, in a more direct way.

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -147,9 +147,28 @@ open class RestActionTestCaseNamingStrategy(
...
         result.add(if (queryParams.size > 1) "${queryParam}s" else queryParam)
+        if (nameWithQueryParameters) {
+            addQueryParameterNames(queryParams, result)
+        }
         return result
     }
 
+    private fun addQueryParameterNames(queryParams: List<QueryParam>, result: MutableList<String>) {
+        val booleanQueryParams = getBooleanQueryParams(queryParams)
+        if (booleanQueryParams.isNotEmpty()) {
```
arcuri82, line 158: redundant if statement. there is no issue in doing a `forEachIndexed` on a empty collection

arcuri82 (changes requested):

arcuri82:
hi @Pgarrett 
  regarding your questions, that's ok. but, those should not be discussed here on GitHub :) I ll add you to Slack

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -475,6 +475,10 @@ class EMConfig {
             throw ConfigProblemException("Python output is used only for black-box testing")
         }
 
+        if (namingStrategy == NamingStrategy.NUMBERED && nameWithQueryParameters) {
```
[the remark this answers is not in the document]
Pgarrett: Understood!

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: 🤦‍♂️ sorry

arcuri82 (approved):

PR #1140, opened 2024-12-03
sql-multidb-v3

arcuri82:


man-zhang:
hi @arcuri82 , the problem related to `ResourceNodeWithDbTest` is fixed. but now it failed due to timeout :( . could you have a look at the problem?

PR #1142, opened 2024-12-09
Phg/number string query param

Pgarrett:
What's in the box:
- When using `nameWithQueryParameters` option, empty strings and negative number query parameters are also added to the name.
- Instead of performing one disambiguation: path or query parameters, it will perform any that apply. The idea behind it is that system tests often require more characteristics to differentiate names, since they're executing several features/topics at the same time.


core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -123,17 +147,14 @@ open class RestActionTestCaseNamingStrategy(
...
+                restAction.path.getOnlyQuery(restAction.parameters)
             }
-            .filter { it.value.size == 1 && it.key.isNotEmpty()}
-            .mapNotNull { entry ->
-                val eInd = entry.value[0]
-                eInd to expandName(eInd, mutableListOf(), ::queryParamsAmbiguitySolver)
-            }
-            .toMap()
+            .filter { it.value.isNotEmpty() && it.key.isNotEmpty()}
+            .flatMap { it.value }.toList()
```
arcuri82, line 157: best to have each stream operation on its own line, ie, in this case the `.toList()` can go on new line

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +181,52 @@ open class RestActionTestCaseNamingStrategy(
...
+        emptyStringQueryParams.forEachIndexed { index, queryParam ->
+            result.add("empty")
+            result.add(queryParam.name)
+            if (index != emptyStringQueryParams.lastIndex) {
+                result.add(and)
+            }
+        }
     }
 
+    // TODO: need to check if the BooleanGene is not enclosed in an OptionalGene
```
arcuri82, line 204: to handle this, you can use the check `staticCheckIfImpactPhenotype()` to see if the boolean gene is in actual use

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +181,52 @@ open class RestActionTestCaseNamingStrategy(
...
+        return queryParams.filter {
+            val wrappedGene = getWrappedGene(it)
+            wrappedGene is BooleanGene && wrappedGene.value
+        }
+    }
+
+    // TODO: need to check if the NumberGene is not enclosed in an OptionalGene
+    private fun getNegativeNumberQueryParams(queryParams: List<QueryParam>): List<QueryParam> {
+        return queryParams.filter {
+            val wrappedGene = getWrappedGene(it)
```
arcuri82, line 215: use `staticCheckIfImpactPhenotype()`

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +181,52 @@ open class RestActionTestCaseNamingStrategy(
...
+            wrappedGene is NumberGene<*> && wrappedGene.value.toLong() < 0
+        }
+    }
+
+    // TODO: need to check if the StringGene is not enclosed in an OptionalGene
+    private fun getEmptyStringQueryParams(queryParams: List<QueryParam>): List<QueryParam> {
+        return queryParams.filter {
+            val wrappedGene = getWrappedGene(it)
+            wrappedGene is StringGene && wrappedGene.getValueAsRawString().trim().isEmpty()
```
arcuri82, line 224: use `staticCheckIfImpactPhenotype()`

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +181,52 @@ open class RestActionTestCaseNamingStrategy(
...
+        }
     }
 
-    private fun removeSolvedDuplicates(duplicatedIndividuals: MutableSet<EvaluatedIndividual<*>>, disambiguatedIndividuals: Set<EvaluatedIndividual<*>>) {
-        duplicatedIndividuals.removeAll(disambiguatedIndividuals)
+    private fun getWrappedGene(queryParam: QueryParam): Gene {
+        return (queryParam.getGeneForQuery() as OptionalGene).gene
```
arcuri82, line 229: there is an already `Gene.getWrappedGene`. use that. the situation can be even more complex, including null and choice genes

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: Hi @arcuri82! Sorry for the delay, last week I was busy working on the presentation for the thesis. All went well, so that's finished.

Regarding this, it's returning `true` for both `OptionalGene`s and `(Boolean/Integer/String)Gene`s. I assume that's actually not expected, right? Maybe it's got something to do with my unit test set up

arcuri82: hi @Pgarrett phenotype is what is expressed and will impact the fitness. if an optional gene is deactivated, it is still in the genotype, but will have no impact on the phenotype. 
a gene X (regardless of its type) inside an OptionalGene that is deactivated will return false in that function, true otherwise. 
we don't want names being affected by genes that are not used. 
if this is not clear, we can have a chat about it

Pgarrett: Ok, isn't that taken care of by using the `RestPath.getOnlyQuery` [function](https://github.com/WebFuzzing/EvoMaster/blob/e637061636dc6d22a793b5d69cf2371e15786c18/core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt#L300)? Since it uses [usableQueryParamsFunction](https://github.com/WebFuzzing/EvoMaster/blob/e637061636dc6d22a793b5d69cf2371e15786c18/core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt#L292C17-L294) which already checks if an `OptionalGene` is active.

Pgarrett: Ok, then I'll add the phenotype check above, to the section in which the query params are obtained ([here](https://github.com/WebFuzzing/EvoMaster/pull/1142/files#diff-1635090843debc863feecb95a31a82cbb08c67fd5b7a66b03582062b01405d8dR168)). That way we can be certain the params being evaluated are query params that are active/have an impact on the phenotype

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -54,18 +60,38 @@ open class RestActionTestCaseNamingStrategy(
...
 
-        return solvedAmbiguities
+        return collectSolvedNames(ambiguitySolversPerIndividual)
+    }
+
+    /*
+     * When two or more individuals share a name, no disambiguation is performed.
+     * Otherwise, we would just be increasing test case name length without having actually disambiguated.
+     */
+    private fun collectSolvedNames(ambiguitySolversPerIndividual: MutableMap<EvaluatedIndividual<*>, MutableList<(Action) -> List<String>>>): Map<EvaluatedIndividual<*>, String> {
```
arcuri82, line 88: as you are not planning to modify `ambiguitySolversPerIndividual` in this function, it should be passed as a `Map` and not a `MutableMap`

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +183,52 @@ open class RestActionTestCaseNamingStrategy(
...
+            }
+        }
     }
 
+    // TODO: need to check if the BooleanGene is not enclosed in an OptionalGene
     private fun getBooleanQueryParams(queryParams: List<QueryParam>): List<QueryParam> {
-        return queryParams.filter { it.getGeneForQuery() is BooleanGene && (it.getGeneForQuery() as BooleanGene).value }
+        return queryParams.filter {
+            val wrappedGene = getWrappedGene(it)
+            wrappedGene is BooleanGene && wrappedGene.value
```
arcuri82, line 210: i think this is still problematic... as you are assuming that the target gene, possibly boolean, is directly under Optional. there can be other genes in between. Maybe better to have something like:
```
val b = it.getGeneForQuery().getWrappedGene(Boolean::class.java)
b != null && b.staticCheckIfImpactPhenotype() && b.value
```

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
```diff
@@ -161,14 +183,52 @@ open class RestActionTestCaseNamingStrategy(
...
+        }
+    }
+
+    // TODO: need to check if the NumberGene is not enclosed in an OptionalGene
+    private fun getNegativeNumberQueryParams(queryParams: List<QueryParam>): List<QueryParam> {
+        return queryParams.filter {
+            val wrappedGene = getWrappedGene(it)
+            wrappedGene is NumberGene<*> && wrappedGene.value.toLong() < 0
```
arcuri82, line 218: same as previous comment

core/src/main/kotlin/org/evomaster/core/problem/rest/RestPath.kt
```diff
@@ -297,11 +297,14 @@ class RestPath(path: String) {
         return params.filter(usableQueryParamsFunction()).size
     }
 
-    fun resolveOnlyQuery(params: List<Param>): List<String> {
-
+    fun getOnlyQuery(params: List<Param>): List<QueryParam> {
```
arcuri82, line 300: maybe this should be called `getOnlyUsableQueries`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/RestActionTestCaseNamingStrategy.kt
Pgarrett: Hi @arcuri82! Glad you brought this up, I was actually going to ask which way you thought best to extract the correct gene in my next PR.
Tested this and it works great for Boolean and String genes. Number genes, not so much since the check in [Gene](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt#L305) is performed by class equality and a numeric gene is instantiated as an IntegerGene or any other. The two options I can think of are:
1. Add a `hasWrappedGeneOf` method in `Gene` that handles these type of subclasses. Code would then have something like `gene.hasWrappedGeneOf(NumberGene);   val numberGene = gene as NumberGene`
2. Call `getWrappedGene` with all the subclasses of NumberGene until one returns not null or none do (in which case it was not a numeric gene at all)

What do you think? The second one does not scale much, but I'm not sure on how keen we're on adding functions like in the first option to classes like Gene

arcuri82: hi @Pgarrett  I just pushed a change to `getWrappedGene`, but i haven't tested it. you can try it out

Pgarrett: Thanks @arcuri82, it works like charm! Just removed the `TODO`s

arcuri82 (approved):

PR #1145, opened 2024-12-11
handling of join queries without WHERE/ON clauses

jgaleotti:


client-java/sql/src/main/java/org/evomaster/client/java/sql/QueryResultSet.java
```diff
@@ -0,0 +1,58 @@
+package org.evomaster.client.java.sql;
+
+import java.util.TreeMap;
+
+public class QueryResultSet {
```
arcuri82, line 5: add JavaDoc, specifying what this class is for

client-java/sql/src/main/java/org/evomaster/client/java/sql/QueryResultSet.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final TreeMap<String, QueryResult> queryResults;
```
arcuri82, line 7: any specific reason to use a TreeMap? even if the concrete class is going to be a TreeMap, shouldn't the reference be a Map? also, need to add comment specifying what the String keys are

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlHeuristicsCalculator.java
```diff
@@ -3,28 +3,96 @@
...
-import static org.evomaster.client.java.sql.internal.SqlParserUtils.getWhere;
+import java.util.Collection;
+import java.util.List;
+
+import static org.evomaster.client.java.sql.internal.SqlParserUtils.*;
 
 public class SqlHeuristicsCalculator {
 
+    public static double C = 0.1;
+    public static double C_BETTER = C + C / 2;
```
arcuri82, line 21: put parentheses... i never remember which has precednce... is this equivalent to `C + (C/2)` ?

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlHeuristicsCalculator.java
```diff
@@ -3,28 +3,96 @@
...
+    public static Truthness TRUE_TRUTHNESS = new Truthness(1, C);
+
+    private QueryResultSet queryResultSet;
+
+    private SqlHeuristicsCalculator(QueryResult[] data) {
+        final boolean isCaseSensitive = false;
```
arcuri82, line 27: a final boolean given as input? why not passing `false` directly? is this for readability?

arcuri82 (changes requested):

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlHeuristicsCalculator.java
jgaleotti: ok

client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlHeuristicsCalculator.java
jgaleotti: yup, just readability

client-java/sql/src/main/java/org/evomaster/client/java/sql/QueryResultSet.java
jgaleotti: I will improve the understability of the code.

client-java/sql/src/main/java/org/evomaster/client/java/sql/QueryResultSet.java
jgaleotti: done

arcuri82 (approved):

PR #1155, opened 2025-01-11
Phg/test case sorting

Pgarrett:
# What's in the box:
- Sorting strategy divided into two options: `COVERED_TARGETS` (current one), `TARGET_INCREMENTAL` (new one)
- `COVERED_TARGETS` will be set as default strategy
- `TARGET_INCREMENTAL` focuses on tests being organized in an incremental criteria:
    - REST: Paths with least levels first, then status codes (500 first, then from 200..), last HTTP verb  
    - GraphQL: Method name first, method type second (`query`/`mutation`), amount of parameters
    - RPC: Tested class name, tested method name, amount of parameters

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
@@ -1,0 +1,11 @@
+package org.evomaster.core.output.sorting
+
+enum class SortingStrategy {
+
+    COVERED_TARGETS,
+    TARGET_INCREMENTAL
+    ;
+
+    fun isCoveredTargets() = this.name.equals("covered_targets", true)
+    fun isTargetIncremental() = this.name.equals("target_incremental", true)
+}
--- a/core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
@@ -1,11 +1,30 @@
 package org.evomaster.core.output.sorting
 
+import org.evomaster.core.search.EvaluatedIndividual
+
+/**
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
```diff
@@ -0,0 +1,11 @@
+package org.evomaster.core.output.sorting
+
+enum class SortingStrategy {
+
+    COVERED_TARGETS,
+    TARGET_INCREMENTAL
```
arcuri82, line 6: add bit of JavaDocs for those 2 options

core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
```diff
@@ -0,0 +1,11 @@
...
+    ;
+
+    fun isCoveredTargets() = this.name.equals("covered_targets", true)
+    fun isTargetIncremental() = this.name.equals("target_incremental", true)
```
arcuri82, line 10: is this needed? we have something similar in `OutputFormat`, but, there, more than 1 enum value could make a predicate true. if it is an exact match, why not checking the enum value directly?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
Pgarrett: Yep, you're right

core/src/main/kotlin/org/evomaster/core/output/sorting/SortingStrategy.kt
Pgarrett: Done

arcuri82 (approved):

PR #1164, opened 2025-01-29
Vulnerability Analyser (Experimental branch)

seran:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2407,6 +2411,28 @@ class EMConfig {
     @Cfg("Apply a security testing phase after functional test cases have been generated.")
     var security = true
 
+    @Experimental
+    @Cfg("Apply vulnerability hunter as part of security testing.")
+    var vulnerabilityAnalyser = false
```
arcuri82: name is quite vague. shouldn't just call it something like `ssrf`?

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -232,6 +233,7 @@ class Main {
             //apply new phases
             solution = phaseHttpOracle(injector, config, solution)
             solution = phaseSecurity(injector, config, epc, solution)
+            solution = phaseVulnerabilityAnalyser(injector, config, solution)
```
arcuri82: shouldn't be a different phase, but rather inside `phaseSecurity`

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -408,6 +410,31 @@ class Main {
...
+        private fun phaseVulnerabilityAnalyser(
+            injector: Injector,
+            config: EMConfig,
+            solution: Solution<*>,
+        ): Solution<*> {
+            if (!config.vulnerabilityAnalyser) {
+                return solution
+            }
+
+            LoggingUtil.getInfoLogger().info("Starting to apply Vulnerability Analyser")
```
arcuri82: confusing name... unless you are doing something more than what we discussed for SSRF

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -272,7 +279,7 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
                     val obj = ObjectGene("body", listOf())
                     // TODO could look at "Accept" header instead of defaulting to JSON
                     val enumGene = EnumGene("contentType", listOf("application/json"))
-                    val body = BodyParam(obj,enumGene)
+                    val body = BodyParam(obj, enumGene)
```
arcuri82: please revert this file. 
you should not format code you have not modified, as it makes reviewing PRs much harder.
unfortunately, at the moment we don't have a common checkstyle enforced in the codebase :( (we should...)

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+import org.evomaster.core.search.gene.Gene
+import org.evomaster.core.search.service.Archive
+import org.evomaster.core.search.service.FitnessFunction
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import javax.annotation.PostConstruct
+import javax.annotation.PreDestroy
+import kotlin.collections.set
+
+class VulnerabilityAnalyser {
```
arcuri82: the name is quite vague, if we deal with SSRF only. if you had others in mind as well, we need to discuss them

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+
+    /**
+     * [HttpCallbackVerifier] to verify HTTP callbacks for vulnerability classes
+     * related to HTTP calls.
+     * i.e., SSRF, XXE
+     */
+    @Inject
+    private lateinit var httpCallbackVerifier: HttpCallbackVerifier
+
+    private var actionVulnerabilityMapping: MutableMap<String, ActionMapping> = mutableMapOf()
```
arcuri82: what is the `key` here? add javadoc.
also, why a `var` and not a `val`?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+    private lateinit var individualsInSolution: List<EvaluatedIndividual<RestIndividual>>
+
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(VulnerabilityAnalyser::class.java)
+    }
+
+    @PostConstruct
+    fun init() {
+        if (config.vulnerabilityAnalyser) {
+//            httpCallbackVerifier = HttpCallbackVerifier()
```
arcuri82: remove?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+
+    fun applyVulnerabilityAnalyser(): Solution<RestIndividual> {
+        LoggingUtil.Companion.getInfoLogger().info("Applying {}", VulnerabilityAnalyser::class.simpleName)
+
+        // extract individuals from the archive
+        val individuals = this.archive.extractSolution().individuals
+
+        individualsInSolution =
```
arcuri82: shouldn't filter on endpoint paths to avoid duplications?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+
+        // execute
+        analyse()
+
+        val individualsAfterExecution = RestIndividualSelectorUtils.findIndividuals(
+            this.archive.extractSolution().individuals,
+            statusCodes = listOf(200, 201)
+        )
+
+        LoggingUtil.getInfoLogger().info("Total individuals after vulnerability analysis: {}", individualsAfterExecution.size)
```
arcuri82: this shouldn;t be in infologger. maybe in class log in debug level?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+
+        return archive.extractSolution()
+    }
+
+    fun hasVulnerabilities(
```
arcuri82: is `vulnerability` correct here? or rather something like `hasMaliciousInput`?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+        // TODO: Can use the extracted CSV to map the parameter name
+        //  to the vulnerability class.
+    }
+
+
+    /**
+     * Private method to classify parameters using a large language model.
+     */
+    private fun llmClassifier() {
+        if (!config.languageModelConnector) {
```
arcuri82: shouldn't this check rather be present in `EMConfig`?

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+    /**
+     * Extract descriptions from the Gene of body payloads.
+     */
+    private fun extractBodyParameters(
+        parameters: List<Param>
+    ): MutableMap<String, ParamMapping> {
+        val output = mutableMapOf<String, ParamMapping>()
+
+        parameters.forEach { param ->
+            when (param) {
```
arcuri82: no `QueryParam`???

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+     * analyzers
+     */
+    private fun analyse() {
+        if (config.problemType == EMConfig.ProblemType.REST) {
+
+            individualsInSolution.forEach { evaluatedIndividual ->
+                evaluatedIndividual.evaluatedMainActions().forEach { a ->
+                    val action = a.action
+                    if (action is RestCallAction) {
+                        if (actionVulnerabilityMapping.containsKey(action.getName())
```
arcuri82: this function has too much nesting. refactor

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityClass.kt
```diff
@@ -0,0 +1,5 @@
+package org.evomaster.core.problem.security
+
+enum class VulnerabilityClass {
```
arcuri82: this class can be removed

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityVerifier.kt
```diff
@@ -0,0 +1,30 @@
+package org.evomaster.core.problem.security
+
+import org.evomaster.core.problem.rest.service.SecurityRest
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+abstract class VulnerabilityVerifier {
```
arcuri82: SSRFVerifier

core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionMapping.kt
```diff
@@ -0,0 +1,30 @@
+package org.evomaster.core.problem.security.data
+
+import org.evomaster.core.problem.security.VulnerabilityClass
+
+class ActionMapping (
```
arcuri82: add JavaDoc

core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionMapping.kt
```diff
@@ -0,0 +1,30 @@
...
+    val name: String,
+) {
+
+    var params: Map<String, ParamMapping> = mutableMapOf()
```
arcuri82: what is the key?

core/src/main/kotlin/org/evomaster/core/problem/security/data/ParamMapping.kt
```diff
@@ -0,0 +1,25 @@
+package org.evomaster.core.problem.security.data
+
+import org.evomaster.core.problem.security.VulnerabilityClass
+
+class ParamMapping (
```
arcuri82: using the term `Param` can be confusing. a `BodyParam` might have several fields, and recursively via nested objects... would each field be considered for SSRF analysis? if so, maybe something like `InputMapping`?

core/src/main/kotlin/org/evomaster/core/problem/security/verifiers/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.util.UUID
+import javax.annotation.PostConstruct
+import javax.annotation.PreDestroy
+
+class HttpCallbackVerifier : VulnerabilityVerifier() {
+
+    @Inject
+    private lateinit var config: EMConfig
```
arcuri82: recall that, if a class is a singleton service, must be under `.service.` package (this applies to all services)

core/src/main/kotlin/org/evomaster/core/problem/security/verifiers/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    private var wireMockServer: WireMockServer? = null
+
+    private var traceTokens: MutableMap<String, String> = mutableMapOf()
```
arcuri82: what is key and value here?

core/src/main/kotlin/org/evomaster/core/problem/security/verifiers/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,148 @@
...
+        wireMockServer!!.stop()
+        wireMockServer = null
+    }
+
+    fun initWireMockServer() {
+        try {
+            val config = WireMockConfiguration()
+                .bindAddress(SecuritySharedUtils.HTTP_CALLBACK_VERIFIER)
+                .extensions(ResponseTemplateTransformer(false))
+                .port(19000) // Changed for testing purposes
```
arcuri82: ? should this be param in `EMConfig`?

core/src/main/kotlin/org/evomaster/core/problem/security/verifiers/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    fun hasTokenForActionName(name: String, value: String): Boolean {
+        return traceTokens.containsKey(name) && traceTokens[name] == value
+    }
+
+    fun isCallbackURL(value: String): Boolean {
+        val pattern =
+            """^http:\/\/${SecuritySharedUtils.HTTP_CALLBACK_VERIFIER}:[0-9]{5}\/sink\/.{36}""".toRegex()
```
arcuri82: explaion the regex

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
seran: I do have plans for other vulnerability detection, but not sure whether I can finish it within the timeline. For now, I'll rename it to SSRFAnalyser.

core/src/main/kotlin/org/evomaster/core/problem/security/VulnerabilityAnalyser.kt
seran: After line 87, it has only one individual per endpoint. I might need your assistance to double check it.

core/src/main/kotlin/org/evomaster/core/problem/security/data/InputFaultMapping.kt
```diff
@@ -0,0 +1,21 @@
...
+    val name: String,
+    val description: String?
+) {
+
+    /**
+     * Holds potential vulnerability class for the [Param].
+     * Key contains the security related faults, and value marks the exploitability
+     * using a [Boolean].
+     */
+    var securityFaults: MutableList<FaultCategory> = mutableListOf()
```
arcuri82, line 13: key? this is a list

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -51,7 +49,7 @@ class VulnerabilityAnalyser {
     @Inject
     private lateinit var httpCallbackVerifier: HttpCallbackVerifier
 
-    private var actionVulnerabilityMapping: MutableMap<String, ActionMapping> = mutableMapOf()
+    private val actionVulnerabilityMapping: MutableMap<String, ActionFaultMapping> = mutableMapOf()
```
arcuri82, line 58: what is the key here?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/security/data/InputFaultMapping.kt
seran: Old comment forget to remove it 🤦‍♂️.

client-java/instrumentation-shared/src/main/java/org/evomaster/client/java/instrumentation/shared/SecuritySharedUtils.java
```diff
@@ -0,0 +1,9 @@
+package org.evomaster.client.java.instrumentation.shared;
+
+public class SecuritySharedUtils {
+
+    /**
+     * Reserved IP address for SSRF verifier.
+     */
+    public static final String HTTP_CALLBACK_VERIFIER = "127.0.0.4";
```
arcuri82: as we are controlling the port (which we could make ephemeral) do we need a different localhost? especially considering it would not work out-of-the-box for Mac, isn't it?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2407,6 +2417,32 @@ class EMConfig {
...
+        MANUAL,
+
+        /**
+         * Use LLMs to select potential vulnerable inputs.
+         */
+        LLM,
+    }
+
+    @Experimental
+    @Cfg("HTTP callback verifier port.")
```
arcuri82: add some more info. eg `A server could be started locally to verify security attacks such as SSRF.`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2407,6 +2417,32 @@ class EMConfig {
...
+    @Min(0.0)
+    @Max(maxTcpPort)
+    var httpCallbackVerifierPort: Int = 19000
+
+    @Experimental
+    @Cfg("Potential vulnerability class associated with a endpoint classification strategy.")
```
arcuri82: description is quite confusing

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+            WRONG: need to check that test call is using a URL, and that this trigger the fault.
+            otherwise, any test with this action type would be marked as faulty
+
+            should check the content of rcr result
+         */
+
+        var hasCallbackURL = false
+
+        action.parameters.forEach { param ->
+            param.primaryGene().getViewOfChildren().forEach { gene ->
```
arcuri82: what if the primaryGene itself contains the payload and there is no children?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -756,6 +761,12 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
             responseClassifier.updateModel(a, rcr)
         }
 
+        if (config.security && config.ssrf) {
+            if (SSRFAnalyser.hasVulnerableInputs(a)) {
+                rcr.setVulnerableForSSRF(true)
```
arcuri82, line 769: is this correct?  you are check if there is any malicious payload in the action, not that, once action is executed, then a call was made to WM, isn't it?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1156,6 +1171,26 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+        individual.seeMainExecutableActions().forEach {
+            val ar = actionResults.find { r -> r.sourceLocalId == it.getLocalId() } as RestCallResult
+
+            if (ar != null && ar.getResultValue(HttpWsCallResult.VULNERABLE_SSRF)?.toBoolean() ?: false) {
+                val scenarioId = idMapper.handleLocalTarget(
+                    idMapper.getFaultDescriptiveId(DefinedFaultCategory.SSRF, it.getName())
+                )
+                fv.updateTarget(scenarioId, 1.0, it.positionAmongMainActions())
+
+                ar.addFault(DetectedFault(DefinedFaultCategory.SSRF, it.getName(), null))
```
arcuri82: instead of `null`, as context here you could use the name of the parameter that leads to the SSRF. (or could just leave it as a TODO comment)

core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionFaultMapping.kt
```diff
@@ -0,0 +1,37 @@
...
+     * TODO: This is temporary to use it in the test cases.
+     */
+    var httpCallbackURL: String? = null
+
+    /**
+     * Holds potential security faults for the [Action].
+     * Key contains the faults related to security, and value marks the exploitability
+     * using a [Boolean].
+     */
+    var securityFaults: MutableMap<FaultCategory, Boolean> = mutableMapOf()
```
arcuri82: don't use objects as key unless you can guarantee they are immutable

core/src/main/kotlin/org/evomaster/core/problem/security/service/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,140 @@
...
+    fun isCallbackURL(value: String): Boolean {
+        // Regex pattern looks for URL contains [HTTP_CALLBACK_VERIFIER] address and [HTTPCallbackVerifier]
+        // port, along with the path /sink/ and UUID as token generated to make the callback URL unique.
+        val pattern =
+            """^http:\/\/${SecuritySharedUtils.HTTP_CALLBACK_VERIFIER}:${config.httpCallbackVerifierPort}\/sink\/.{36}""".toRegex()
+
+        return pattern.matches(value)
+    }
+
+    fun generateCallbackLink(name: String): String {
```
arcuri82, line 77: provide some explanation for this / documentation

core/src/main/kotlin/org/evomaster/core/problem/security/service/HttpCallbackVerifier.kt
```diff
@@ -0,0 +1,140 @@
...
+        val token = UUID.randomUUID().toString()
```
arcuri82: why a UUID? explain motivation in comment for why we need a non-deterministic here

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+
+    /**
+     * Private method to classify parameters using a large language model.
+     */
+    private fun llmClassifier() {
+        // For now, we consider only the individuals selected from [Archive]
+        // TODO: This can be isolated to classify at the beginning of the search
+        individualsInSolution.forEach { evaluatedIndividual ->
```
arcuri82, line 202: as discussed, likely we ll not need LLM here inside EM for this.
however, if we, isn't here possible duplication? eg, if some action is present in different test cases (which I guess it is a common occurence)

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+    private fun extractBodyParameters(
+        parameters: List<Param>
+    ): MutableMap<String, InputFaultMapping> {
+        val output = mutableMapOf<String, InputFaultMapping>()
+
+        parameters.forEach { param ->
+            when (param) {
+                is BodyParam -> {
+                    param.seeGenes().filter { it.name == "body" }.forEach { gene ->
+                        gene.getAllGenesInIndividual().forEach { geneInIndividual ->
```
arcuri82: `getAllGenesInIndividual()`? this takes everything, also from the other actions...

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -0,0 +1,332 @@
...
+            actionMapping.securityFaults[DefinedFaultCategory.SSRF] = true
+            // Create a testing target
+            archive.addIfNeeded(executedIndividual)
+        }
+    }
+
+    private fun updateGeneWithCallbackURL(actionName: String, primaryGene: Gene, callBackUrl: String) {
+        primaryGene.getViewOfChildren().forEach { gene ->
+            if (actionVulnerabilityMapping.containsKey(actionName)) {
+                val g = actionVulnerabilityMapping[actionName]!!.params[gene.name]
```
arcuri82: on possible issue is that gene names within a tree might not be unique... i guess ll see when running experiiments/e2e

arcuri82 (changes requested):

client-java/instrumentation-shared/src/main/java/org/evomaster/client/java/instrumentation/shared/SecuritySharedUtils.java
seran: Yes, for macOS it won't work. Initially I thought we may expand this in future so allocated a new IP. However, it makes sense to run it in `localhost`. I'll revert the changes.

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
seran: `setFromStringValue` updates the children also recursively as per the comments. So far, this is not creating any problems. Here `gene.name` is used to track the parameter. However, I thinking of a different way to do this and running more experiments to see whether this will be a problem. I can see the tests generated with the payload.

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -0,0 +1,368 @@
...
+                if (g.securityFaults.contains(DefinedFaultCategory.SSRF)) {
+                    // Only change the param marked for SSRF
+                    // This updates the children also recursively
+                    gene.setFromStringValue(callBackUrl)
+                }
+            }
+        }
+    }
+
+    private fun getStringGenesFromParam(genes: List<Gene>): List<Gene> {
```
seran, line 347: @arcuri82 I need you comment on this approach.

PR #1165, opened 2025-01-30
Implementing hard limit for test case name length, default 80 chars

Pgarrett:
# What's in the box
- Implementing a hard limit on test case name length.
- Default value is 80 chars, though it is possible to set a different one with the `maxTestCaseNameLength` parameter
- At start, the amount of chars available will be the name length minus the length of the `test_#N_` string, taking N as the size of the test suite.
- If adding a token or set of tokens to the name makes it exceed the allowed size, then they are not added.
  - All of the added name tokens go through this process of checking name size in order to have an updated value of the available amount of chars for the name

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
@@ -14,6 +14,8 @@ import org.evomaster.core.sql.SqlAction
 abstract class ActionTestCaseNamingStrategy(
     solution: Solution<*>,
     private val languageConventionFormatter: LanguageConventionFormatter,
+    protected val maxTestCaseNameLength: Int,
+    private val testCasesSize: Int = solution.individuals.size,
 ) : NumberedTestCaseNamingStrategy(solution)  {
 
     protected val on = "on"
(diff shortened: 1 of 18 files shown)


core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
```diff
@@ -14,6 +14,8 @@ import org.evomaster.core.sql.SqlAction
 abstract class ActionTestCaseNamingStrategy(
     solution: Solution<*>,
     private val languageConventionFormatter: LanguageConventionFormatter,
+    protected val maxTestCaseNameLength: Int,
+    private val testCasesSize: Int = solution.individuals.size,
```
arcuri82, line 18: having it here a input parameter feels quite weird, considering it depends on `solution`. maybe use a field that is initialized in an `init{}` block?

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
```diff
@@ -6,7 +6,7 @@ import org.evomaster.core.search.Solution
 import java.util.Collections.singletonList
 
 open class NumberedTestCaseNamingStrategy(
-    solution: Solution<*>
+    solution: Solution<*>,
```
arcuri82, line 9: why the `,`?

core/src/test/kotlin/org/evomaster/core/EMConfigTest.kt
```diff
@@ -641,4 +641,25 @@ internal class EMConfigTest{
         assertFalse(config.nameWithQueryParameters)
     }
 
+    @Test
+    fun testMaxCharsLengthIs80ByDefault() {
```
arcuri82, line 645: what is the point of a unit test checking the default value?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/NumberedTestCaseNamingStrategy.kt
Pgarrett: typo

core/src/main/kotlin/org/evomaster/core/output/naming/ActionTestCaseNamingStrategy.kt
Pgarrett: Yep, hadn't though of it, but will do so

arcuri82 (approved):

PR #1170, opened 2025-02-07
Environment actions suffix are only added to test case names if less than half

Pgarrett:
# What's in the box
Environment actions such as SQL, Mongo and WireMock will not be added as a test case name suffix if either:
- All test cases share the same action
- In a suite of more than 10 test cases, more than half of them share the same action

Number 10 was chosen thinking of a small test suite. If 6 out of 10 test cases have the `UsingMongo` suffix, repetition is not so terrible.


arcuri82 (approved):

arcuri82:
@Pgarrett   ingore my approval... looks like u were not done yet and didn't ask my review yet... :)

arcuri82 (approved):

PR #1179, opened 2025-02-25
Swagger Description inside Param extension for Security

seran:


arcuri82:
@seran  hi, should I review this PR, or are u still working on it?

seran:
> @seran hi, should I review this PR, or are u still working on it?

I made a last minute change. I'll send you a PR review request once the CI is completed

core/src/main/kotlin/org/evomaster/core/problem/api/param/Param.kt
```diff
@@ -9,16 +9,26 @@ abstract class Param(
         val genes : MutableList<Gene>
 ) : StructuralElement(genes){
 
+    /**
+     * Contains the description of the parameter.
+     */
+    var description: String? = null
+        set(value) {
+            if (!value.isNullOrEmpty() && field.isNullOrEmpty()) {
```
arcuri82, line 17: i guess if a `field` is already set we should throw an IllegalStateException, stating can only be set once? also, add comment here to specify why we do this. also, if this is just on Param, we could force it to be set on constructor time (ie Param is not so widely used as Gene, requiring lot of changes). ah, never mind, there is issue with copy() :) still clarify it in comments

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
```diff
@@ -621,16 +624,14 @@ object RestActionBuilderV3 {
...
-            }
+            "query" -> params.add(QueryParam(name, gene, p.explode ?: true, p.style ?: Parameter.StyleEnum.FORM)
+                    .apply { this.description = description })
             /*
                 a path is inside a Disruptive Gene, because there are cases in which we want to prevent
                 mutation. Note that 1.0 means can always be mutated
              */
             "path" -> params.add(PathParam(name, CustomMutationRateGene("d_", gene, 1.0)))
-            "header" -> params.add(HeaderParam(name, gene))
+            "header" -> params.add(HeaderParam(name, gene).apply { this.description = description })
```
arcuri82, line 634: shouldn't this apply to all kinds of `param`?  so could have the `apply` outside of the `when`

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -77,6 +77,16 @@ abstract class Gene(
         private val log: Logger = LoggerFactory.getLogger(Gene::class.java)
     }
 
+    /**
+     * Description about the gene.
+     */
+    var description: String? = null
+        set(value) {
+            if (!value.isNullOrEmpty() && field.isNullOrEmpty()) {
```
arcuri82, line 85: same comment as in `Param`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
seran: We are directly add the `Param` to the `MutableList`. So I added the description directly to other types also.

arcuri82 (approved):

PR #1188, opened 2025-03-02
Sending endpoint url as a parameter for auth util

omursahin:
After merging this PR, I will update the EMB SUTs using getForDefaultSpringFormLogin function.

### DIFF
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
@@ -67,28 +67,7 @@ public class AuthUtils {
      * @return a DTO
      */
     public static AuthenticationDto getForDefaultSpringFormLogin(String dtoName, String username, String password){
-
-        LoginEndpointDto cookie = new LoginEndpointDto();
-
-        cookie.endpoint = "/login";
-        cookie.verb = HttpVerb.POST;
-        cookie.contentType = "application/x-www-form-urlencoded";
-        cookie.expectCookies = true;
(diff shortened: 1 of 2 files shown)


client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
```diff
@@ -66,11 +66,11 @@ public static AuthenticationDto getForAuthorizationHeader(String dtoName, String
      * @param password  password for that user
      * @return a DTO
      */
-    public static AuthenticationDto getForDefaultSpringFormLogin(String dtoName, String username, String password){
+    public static AuthenticationDto getForDefaultSpringFormLogin(String dtoName, String username, String password, String endpoint){
```
arcuri82, line 69: this would break all existing driver that used this function. you can create a new function with same name and parameters, and have old function without `String endpoint` calling this version with `/login` as input (as Java does not have default values for method inputs)

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
arcuri82: do not copy&paste code. the other function with 1 less parameter should remove its code, and call this longer version with `return getForDefaultSpringFormLogin(dtoName, username, password, endpoint,"/login");`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1215, opened 2025-04-04
LLM Connector for Vulnerability Analyser

seran:


arcuri82:
hi @seran  I see a `marked this pull request as ready for review`, but i was not added as reviewer. in those cases, i don't get a notification... add me as reviewer if you are done with this PR

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2370,10 +2370,27 @@ class EMConfig {
     @Cfg("When generating data, allow in some cases to use invalid values on purpose")
     var allowInvalidData: Boolean = true
 
-    @Experimental
     @Cfg("Apply a security testing phase after functional test cases have been generated.")
+    @Experimental
```
arcuri82, line 2374: why moving this annotation?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+import javax.annotation.PreDestroy
+import kotlin.math.min
+
+/**
+ * A utility service designed to handle large language model server
+ * related functions.
+ *
+ * Designed to work with Ollama (version 0.6.2).
+ */
+class LanguageModelConnector {
```
arcuri82, line 27: if it is a service, should be under a `*/service/*` package

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+
+    @PostConstruct
+    fun initialise() {
```
arcuri82, line 30: apart from static methods, instance methods should be put after the field declarations

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+
+    private var prompts: MutableMap<String, Prompt> = mutableMapOf()
+
+    private val objectMapper = ObjectMapper()
+
+    private var actualFixedThreadPool = 0
+
+    private lateinit var workerPool: ExecutorService
+
+    @PostConstruct
```
arcuri82, line 45: why 2 `@PostConstuct` functions instead of 1?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+    fun initialise() {
+        LoggingUtil.getInfoLogger().info("Initializing {}", LanguageModelConnector::class.simpleName)
+    }
+
+    @Inject
+    private lateinit var config: EMConfig
+
+    private var prompts: MutableMap<String, Prompt> = mutableMapOf()
```
arcuri82, line 37: maps must always specify in comments what the numeric/string keys are

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+    private fun shutdown() {
+        Lazy.assert { config.languageModelConnector }
+        workerPool.shutdown()
+        prompts.clear()
+    }
+
+    /**
+     * To query the large language server asynchronously with a simple prompt.
+     */
+    fun queryAsync(prompt: String) {
```
arcuri82, line 74: shouldn't return a `Future` so the caller can wait on it?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+     * @return null if there is no answer for the prompt
+     */
+    fun getAnswerByPrompt(prompt: String): Prompt? {
+        return prompts.filter { it.value.prompt == prompt && it.value.hasAnswer() }.values.firstOrNull()
+    }
+
+    /**
+     * @return answer for the UUID of the prompt
+     */
+    fun getAnswerById(id: String): Prompt? {
```
arcuri82, line 103: if this is a UUID, why not using a `UUID` object?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+        return prompts[id]
+    }
+
+    /**
+     * To query the large language server with a simple prompt.
+     * @return answer string from the language model server
+     */
+    fun query(prompt: String): String? {
+        if (!config.languageModelConnector) {
+            return null
```
arcuri82, line 113: shouldn't rather throw an `IllegalStateException` ?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * Note: If you are using Ollama as a server, please make sure to set the
+     * CORS origin for Ollama on the host operating system.
+     *
+     * Reference:
+     * https://medium.com/dcoderai/how-to-handle-cors-settings-in-ollama-a-comprehensive-guide-ee2a5a1beef0
+     */
+    private fun call(languageModelServerURL: String, requestBody: String): String? {
+        try {
+            val connection = URL(languageModelServerURL).openConnection() as HttpURLConnection
```
arcuri82, line 162: so each time we create a new TCP socket for the calls? shouldn't be better to re-use the connection objects?

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+            connection.doInput = true
+            connection.doOutput = true
+
+            val writer = DataOutputStream(connection.outputStream)
+            writer.writeBytes(requestBody)
+            writer.flush()
+            writer.close()
+
+            if (connection.responseCode != HttpURLConnection.HTTP_OK) {
+                LoggingUtil.getInfoLogger().warn("Failed to connect to language model server.")
```
arcuri82, line 177: in the log messages, should have the status code returned. 
also this should be the log of this class, and not the info-logger. 
also, would make sense to open a connection on service startup, and provide info-logger message if it fails (eg due to wrong provided URL)

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
arcuri82: also, might had been better to use something like `HttpClientFactory.createTrustingJerseyClient`

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            // This is designed to use the non-stream outputs.
+            // If stream is needed, consider implementing a
+            // different method to handle stream outputs.
+            val response = objectMapper.readValue(connection.inputStream, OllamaResponse::class.java)
+
+            return response.response
+        } catch (e: Exception) {
+            LoggingUtil.getInfoLogger().warn("Failed to connect to language model server: ${e.message}.")
```
arcuri82, line 188: see previous comments on logs

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    /**
+     * Private method, returns the large language model server URL from
+     * EMConfig, otherwise returns the possible default local URL.
+     */
+    private fun getLanguageModelServerURL(): String {
+        val languageModelServerURL = if (config.languageModelServerURL == null) {
+            "http://localhost:11434/api/generate"
```
arcuri82, line 200: is this some test utility? in such case, should not be here. 
otherwise,  what is 11434 representing? is it some defaults of Ollama? clarify

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Private method, returns the large language model server name from
+     * EMConfig, otherwise return llama3.2:latest.
+     *
+     * TODO: Failsafe model name yet to be decided
+     */
+    private fun getLanguageModelName(): String {
+        val languageModelName = if (config.languageModelName == null) {
+            "llama3.2:latest"
```
arcuri82, line 220: those configuration defaults should not be here, but rather in `EMConfig`

core/src/main/kotlin/org/evomaster/core/languagemodel/data/Prompt.kt
```diff
@@ -0,0 +1,14 @@
+package org.evomaster.core.languagemodel.data
+
+class Prompt (
+    val id: String,
+
+    val prompt: String
+) {
+
+    var answer: String? = null
```
arcuri82, line 9: adding `answer` to `Prompt` might be confusing. better to create a new class `AnsweredPrompt` with fields `prompt: Prompt` and `answer: String`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/module/RestModule.kt
```diff
@@ -18,61 +19,61 @@ import org.evomaster.core.search.service.mutator.Mutator
...
+            .asEagerSingleton()
 
         bind(HttpWsExternalServiceHandler::class.java)
-                .asEagerSingleton()
+            .asEagerSingleton()
 
         bind(HarvestActualHttpWsResponseHandler::class.java)
             .asEagerSingleton()
 
+        bind(LanguageModelConnector::class.java)
```
arcuri82, line 75: rather have it under `EnterpriseModule`

core/src/test/kotlin/org/evomaster/core/languagemodel/LanguageModelConnectorTest.kt
```diff
@@ -0,0 +1,50 @@
...
+class LanguageModelConnectorTest {
+
+    private lateinit var config: EMConfig
+    private lateinit var languageModelConnector: LanguageModelConnector
+
+    companion object {
+        @BeforeAll
+        @JvmStatic
+        fun skipOnGA() {
+            CIUtils.skipIfOnGA()
```
arcuri82, line 21: if skipping on GA, add comment to explain why

core/src/test/kotlin/org/evomaster/core/languagemodel/LanguageModelConnectorTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val injector: Injector = LifecycleInjector.builder()
+            .withModules(BaseModule())
+            .build().createInjector()
+
+
+
+        config = injector.getInstance(EMConfig::class.java)
+        config.languageModelConnector = true
+
+        languageModelConnector = injector.getInstance(LanguageModelConnector::class.java)
```
arcuri82, line 37: does this work? you only setup `BaseModule`

core/src/test/kotlin/org/evomaster/core/languagemodel/LanguageModelConnectorTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        // If languageModelName or languageModelURL set to empty, an exception
+        // will the thrown.
+        config.languageModelName = "llama3.2:latest"
+        config.languageModelServerURL = "http://localhost:11434/api/generate"
```
arcuri82, line 42: this will fail if run locally and one does not have Ollama. can Ollama be run with Docker? otherwise, this "test" should be disabled, with some comments to explain situation

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/languagemodel/LanguageModelConnector.kt
seran: this was a mistake :(

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/languagemodel/service/LanguageModelConnector.kt
```diff
@@ -0,0 +1,317 @@
...
+
+    private val httpClients: ConcurrentHashMap<Long, Client> = ConcurrentHashMap()
+
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(LanguageModelConnector::class.java)
+    }
+
+    @PostConstruct
+    fun init() {
+        LoggingUtil.Companion.getInfoLogger().info("Initializing {}", LanguageModelConnector::class.simpleName)
```
arcuri82: this info log should be inside the `if (config.languageModelConnector)`

core/src/main/kotlin/org/evomaster/core/languagemodel/service/LanguageModelConnector.kt
```diff
@@ -0,0 +1,317 @@
...
+
+        if (config.languageModelConnector) {
+
+            if (!this.isModelAvailable()) {
+                LoggingUtil.uniqueWarn(log, "${config.languageModelName} is not available in the provided " +
+                        "language model server URL: ${config.languageModelServerURL}. " +
+                        "Language Model Connector will be disabled.")
+                config.languageModelConnector = false
```
arcuri82: cannot do this. you should not modify `config` values programmatically inside services. rather have a state here in this class, and then a boolean function telling whether the LLM is available

core/src/main/kotlin/org/evomaster/core/languagemodel/service/LanguageModelConnector.kt
```diff
@@ -0,0 +1,332 @@
...
+
+    private val objectMapper = ObjectMapper()
+
+    private var actualFixedThreadPool = 0
+
+    private lateinit var workerPool: ExecutorService
+
+    private val httpClients: ConcurrentHashMap<Long, Client> = ConcurrentHashMap()
+
+    private var isLanguageModelAvailable: Boolean = false
```
arcuri82, line 56: there should be a way to query this from users of this service

arcuri82 (approved):

PR #1216, opened 2025-04-07
added check for when test prematurely stops

arcuri82:


arcuri82:
@omursahin  i ll merge once CI passes

PR #1225, opened 2025-04-16
sql_heuristics_aliases

jgaleotti:
first version of SQL aliases for advanced SQL heuristics


arcuri82:
@jgaleotti  hi. you made some changes after you requested my review, and now the build fails. please fix, and re-ask my review once done ;)

jgaleotti:
> @jgaleotti hi. you made some changes after you requested my review, and now the build fails. please fix, and re-ask my review once done ;)

Thanks, I'm into this.

client-java/sql/src/main/java/org/evomaster/client/java/sql/heuristic/ColumnReference.java
```diff
@@ -0,0 +1,44 @@
+package org.evomaster.client.java.sql.heuristic;
+
+import java.util.Objects;
+
+public class ColumnReference {
```
arcuri82, line 5: provide JavaDoc to explain what this class is for

arcuri82 (changes requested):

client-java/sql/src/main/java/org/evomaster/client/java/sql/heuristic/ColumnReference.java
jgaleotti: added to this class and other classes similar to this one.

arcuri82 (approved):

PR #1231, opened 2025-04-25
fix resetDatabase when executing generated tests

man-zhang:


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -685,6 +690,29 @@ public final DbInfoDto getSqlDatabaseSchema() {
...
+            schemaDto.extraConstraintDtos.addAll(extra);
+        }
+
+        return schemaDto;
+    }
+
+    /**
+     * @return extracted Sql Database Schema. Note that null is returned only if the schema has been extracted or the extraction fails with the given db connection.
+     */
+    public final DbInfoDto extractSqlDbSchemaAndConstraints(){
```
arcuri82, line 709: this is bit confusing... it is a public method. if it is called with an existing `schemaDto` it will return `null`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -685,6 +690,29 @@ public final DbInfoDto getSqlDatabaseSchema() {
             return schemaDto;
         }
 
+        DbInfoDto schema = extractSqlDbSchemaAndConstraints();
```
arcuri82, line 693: is this `schema` object used? need some clarifying comment, or more descriptive function names

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
man-zhang: this function has been changed to return `Boolean` representing whether the extraction has been performed successfully.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
man-zhang: this has been changed now.

PR #1247, opened 2025-05-14
starting to create examples for ai constraint handling

arcuri82:


arcuri82:
@MohsenTaheriShalmani  you can take over this branch

core-it/src/main/kotlin/bar/examples/it/spring/aiconstraint/numeric/AICNumericApplication.kt
```diff
@@ -0,0 +1,95 @@
...
+import io.swagger.v3.oas.annotations.media.Schema
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@SpringBootApplication(exclude = [SecurityAutoConfiguration::class])
+@RequestMapping(path = ["/api"])
+@RestController
+open class AICNumericApplication {
```
arcuri82: here you added new elements like boolean and enum. it is fine, but, then , this is not really any more just a "Numeric" application. should change the name.  
however, recall you can have several different SUTs here, in different package names. it makes sense to have a separate SUT for each base type (to check they work fine by their own), and then other SUTs for combination of types in same endpoint (as here)

core-it/src/test/kotlin/org/evomaster/core/problem/rest/aiconstraint/numeric/AICNumericCheck.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+
+class AICNumericCheck : IntegrationTestRestBase() {
+
+    companion object {
+        @BeforeAll
+        @JvmStatic
```
arcuri82: remove the `@BeforeAll` as it is a JUnit annotation, which will be ignore here

core-it/src/test/kotlin/org/evomaster/core/problem/rest/aiconstraint/numeric/AICNumericCheck.kt
```diff
@@ -0,0 +1,89 @@
...
+        fun main(args: Array<String>) {
+            val test = AICNumericCheck()
+            init()
+            test.initializeTest()
+            test.runClassifierExample()
+        }
+    }
+
+    @BeforeEach
```
arcuri82: remove the `@BeforeEach` as it is a JUnit annotation, which will be ignore here

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/inputEncoder.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import org.evomaster.core.problem.rest.data.RestCallAction
+import org.evomaster.core.search.gene.BooleanGene
+import org.evomaster.core.search.gene.collection.EnumGene
+import org.evomaster.core.search.gene.numeric.DoubleGene
+import org.evomaster.core.search.gene.numeric.IntegerGene
+
+
+// Encoding IntegerGene, DoubleGene, EnumGene, and BooleanGene of a RestCallAction into a double vector
+fun inputEncoder(input: RestCallAction): List<Double> {
```
arcuri82: it is possible to have package-level functions in Kotlin, but it is not something we use in EvoMaster. rather create a `object` utility class, with this function inside it

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/inputEncoder.kt
```diff
@@ -0,0 +1,48 @@
+package org.evomaster.core.problem.rest.classifier
+
+import org.evomaster.core.problem.rest.data.RestCallAction
+import org.evomaster.core.search.gene.BooleanGene
+import org.evomaster.core.search.gene.collection.EnumGene
+import org.evomaster.core.search.gene.numeric.DoubleGene
+import org.evomaster.core.search.gene.numeric.IntegerGene
+
+
+// Encoding IntegerGene, DoubleGene, EnumGene, and BooleanGene of a RestCallAction into a double vector
```
arcuri82: comments on functions and fields should be within a `/**  */` (note the initial double-*), at it would make them into JavaDoc, that are then picked up by tools and IDEs (eg when hovering over such elements with mouse)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/inputEncoder.kt
```diff
@@ -0,0 +1,48 @@
...
+fun inputEncoder(input: RestCallAction): List<Double> {
+
+    // Create enum encodings
+    val enumEncoding = mutableMapOf<String, Map<String, Int>>()
+    for (parameter in input.parameters) {
```
arcuri82: you might want to iterate of sorted parameters, by name, to avoid non-deterministic behavior

arcuri82: also, for now, might just want to consider `QueryParam`

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/inputEncoder.kt
```diff
@@ -0,0 +1,48 @@
...
+            val values = gene.values
+                .map { it.toString() }
+                .filter { it != "EVOMASTER" } //excluding "EVOMASTER" as it appears in Enum lists
+            enumEncoding[parameter.name] = values.withIndex().associate { it.value to it.index }
+        }
+    }
+
+    val encodedFeatures = mutableListOf<Double>()
+    for (gene in input.seeTopGenes()) {
+        when (gene) {
```
arcuri82: could leave TODO note related to handle other types, eg, OptionalGene, and other types of numerics (eg Long and Float)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -0,0 +1,56 @@
...
+    @Inject
+    private lateinit var config : EMConfig
+
+    private lateinit var delegate: AIModel
+
+    // Input dimension of the classifier
+    private var dimension: Int = 0
+    fun setDimension(d: Int) {
+        this.dimension = d
+    }
```
arcuri82: this function and field should not be in this service

arcuri82 (commented):
@MohsenTaheriShalmani @amidgol  look into my new comments

PR #1249, opened 2025-05-21
Phg/dto kickoff

Pgarrett:
# What's in the box?
Kickoff for using DTOs instead of stringified json for REST request payloads.
Current limitations:
- it only supports Java as output format language
- supported fields do not include lists nor custom nested objects
- no strategy yet defined for DTOs defined in a spec path (not as a component with a predefined name)
- DTOs not instantiated for to be used in the test payload
- DTO creation should be done before splitting tests, to avoid re-writting and to have complete DTOs

The current implementation is able to handle wanted field scenarios using optionals and jackson:
- field is non null and sent in the payload: `{"myKey": "myValue"}`/`{"myKey": ""}`
- field is null and sent in the payload: `{"myKey": null}`
- field is not sent in the payload: `{}`

For example, a spec with the following components:
```components:
  schemas:
    Person:
      type: object
      properties:
        name:
          type: string
        age:
          type: integer
        address:
          $ref: "#/components/schemas/Address"
      required:
        - name
        - age
        - address
    Address:
      type: object
      properties:
        street:
          type: string
        country:
          type: string
        voted:
          type: boolean
      required:
        - street
        - country
```

Will generate a class such as:
```
package dto;

import java.util.Optional;

import com.fasterxml.jackson.annotation.JsonInclude;
import shaded.com.fasterxml.jackson.annotation.JsonProperty;

@JsonInclude(JsonInclude.Include.NON_NULL)
public class Person {

    @JsonProperty("name")
    private Optional<String> name;

    @JsonProperty("age")
    private Optional<Integer> age;

    public Optional<String> getName() {
        return name;
    }
    
    public void setName(String name) {
        this.name = Optional.ofNullable(name);
    }

    public Optional<Integer> getAge() {
        return age;
    }
    
    public void setAge(Integer age) {
        this.age = Optional.ofNullable(age);
    }

}
```


arcuri82:
@Pgarrett 

a comment regarding

> field is empty and sent in the payload: `{"myKey": ""}`

an empty string in JSON is a valid string. i guess you should make sure to handle the case of `null`, eg, `{"myKey": null}`

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -818,6 +818,10 @@ class Main {
 
             val writer = injector.getInstance(TestSuiteWriter::class.java)
 
+            if (config.problemType == EMConfig.ProblemType.REST && config.dtoForRequestPayload && config.outputFormat.isJava()) {
```
arcuri82, line 821: add a TODO comment stating we need to add support for Kotlin

core/src/main/kotlin/org/evomaster/core/output/dto/DtoField.kt
```diff
@@ -0,0 +1,8 @@
+package org.evomaster.core.output.dto
+
+class DtoField(
+    val name: String,
+    val type: String,
+){
+
+}
```
arcuri82, line 8: can remove the `{}` if there is no code there

core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoWriter.kt
```diff
@@ -0,0 +1,105 @@
+package org.evomaster.core.output.dto
+
+import org.evomaster.core.output.Lines
+import org.evomaster.core.output.OutputFormat
+import org.evomaster.core.output.TestSuiteFileName
+import org.evomaster.core.utils.StringUtils
+import java.nio.file.Files
+import java.nio.file.Path
+
+class JavaDtoWriter(private val testSuitePath: Path,
```
arcuri82, line 10: add some JavaDoc

core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoWriter.kt
```diff
@@ -0,0 +1,105 @@
...
+                    private val outputFormat: OutputFormat,
+                    private val dtoClass: DtoClass) {
+
+    val dtoFilename: TestSuiteFileName
+
+    init {
+        dtoFilename = TestSuiteFileName(appendDtoPackage(dtoClass.name))
+    }
+
+    fun write() {
```
arcuri82, line 20: any reason to create a class with constructor, instead of having an object (eg `object JavaDtoWriter`) with a parametric `fun write(...)` ?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -201,6 +218,14 @@ class TestSuiteWriter {
         )
     }
 
+    // TODO: extract DTO extraction and writing to a different class
+    fun writeDtos(solutionFilename: String) {
+        val testSuitePath = getTestSuitePath(TestSuiteFileName(solutionFilename), config).parent
+        getDtos().forEach {
+            JavaDtoWriter(testSuitePath, config.outputFormat, it).write()
```
arcuri82, line 225: `JavaDtoWriter.write(testSuitePath, config.outputFormat, it)`

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
...
 
+    private fun getDtos(): List<DtoClass> {
+        val restSampler = sampler as AbstractRestSampler
+        val result = mutableListOf<DtoClass>()
+        restSampler.getActionDefinitions().forEach { action ->
+            action.getViewOfChildren().forEach { child ->
+                if (child is BodyParam) {
+                    val primaryGene = child.primaryGene()
+                    // TODO: Payloads could also be json arrays, analyze ArrayGene
+                    if (primaryGene is ObjectGene) {
```
arcuri82, line 1107: this could be inside an `OptionalGene`, or other kinds of wrapper genes. might want to get reference to the `ObjectGene` (if any) using `primaryGene.getWrappedGene`

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
             .toList()
     }
 
+    private fun getDtos(): List<DtoClass> {
+        val restSampler = sampler as AbstractRestSampler
+        val result = mutableListOf<DtoClass>()
+        restSampler.getActionDefinitions().forEach { action ->
```
arcuri82, line 1102: you could move the code here under some utility function in the `dto` package, and then call it with `restSampler.getActionDefinitions()` as input

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
...
+            action.getViewOfChildren().forEach { child ->
+                if (child is BodyParam) {
+                    val primaryGene = child.primaryGene()
+                    // TODO: Payloads could also be json arrays, analyze ArrayGene
+                    if (primaryGene is ObjectGene) {
+                        // TODO: Determine strategy for objects that are not defined as a component and do not have a name
```
arcuri82, line 1108: could use something related to `action.getName`, and put this as a comment

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
...
+                        val dtoClass = DtoClass(primaryGene.refType?:primaryGene.hashCode().toString())
+                        primaryGene.fixedFields.forEach { field ->
+                            try {
+                                dtoClass.addField(getDtoField(field))
+                            } catch (e: Exception) {
+                                // do nothing for the moment
```
arcuri82, line 1114: at least warn log it

arcuri82: also, might want to add an `assert(false)`. ie, do not crash in production, but fail if it happens in any test

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
...
+                }
+            }
+        }
+        return result
+    }
+
+    private fun getDtoField(field: Gene): DtoField {
+        val wrappedGene = GeneUtils.getWrappedValueGene(field)
+        return DtoField(field.name, when (wrappedGene) {
+            // TODO: handle nested arrays, objects and extend type system for dto fields
```
arcuri82, line 1128: you will need to handle the cases of `OptionalGene` and `NullableGene` (see how wrapping works), as that will impact what kinds of field types you will need to create. also RegexGene is going to be common. anyway, you will see what things crashing when we activate this feature in the E2E...  :)

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1071,4 +1096,47 @@ class TestSuiteWriter {
...
+            is StringGene -> "String"
+            is IntegerGene -> "Integer"
+            is LongGene -> "Long"
+            is DoubleGene -> "Double"
+            is FloatGene -> "Float"
+            is Base64StringGene -> "String"
+            is DateGene -> "String"
+            is TimeGene -> "String"
```
arcuri82, line 1136: you might want to use special types for this, and, in the DTOs, throw an `IllegalArgumentExcpetion` if invalid string. however, for now can leave it as acomment (as we still create invalid strings outside robustness testing, and that is not fixed yet)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoWriter.kt
Pgarrett: Actually, two reasons:
1. Hadn't thought about it, I just got writing and my first impulse was to add that information in the constructor.
2. This writer will most likely change. I was thinking this would change into some kind of delegate/decorator. You should actually ask the `DtoClass` to write itself, which in turn would call the right type of DtoWriter

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: As a matter of fact, we need more E2E tests! Most e2e tests check on tests generated using GET:
- GET: 103+ matches in 36+ files
- POST: 67 matches in 36 files
- PUT: 6 matches in 5 files

And I looking into the POST tests, some do not have a DTO request payload but instead a response one, or are using simple cases. Looked into that for debugging

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Maybe it's something the Gene could answer by itself, something like: `gene.getDtoType()` and have that answer back. That would also save us from these giant ifs

One other question I have is, for these specific cases of Time and Date, should we in the future handle them with jodatime objects or similar? Or just set them as strings and keep it simple? For a start I think strings are fine, not sure what you had in mind for the future

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Could be, yes. We'll most likely come back to this later on

arcuri82 (approved):

PR #1255, opened 2025-05-30
Add support for nested objects

Pgarrett:
- Moved DTO extraction to a specific object
- Support for nested objects


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.evomaster.core.search.gene.numeric.LongGene
+import org.evomaster.core.search.gene.string.Base64StringGene
+import org.evomaster.core.search.gene.string.StringGene
+import org.evomaster.core.search.gene.utils.GeneUtils
+import org.evomaster.core.utils.StringUtils
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.nio.file.Path
+
+object DtoWriter {
```
arcuri82, line 33: provide some /** */ JavaDoc discussing briefly what is the point of writing DTOs

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    private val log: Logger = LoggerFactory.getLogger(DtoWriter::class.java)
+    private val dtoCollector: MutableMap<String, DtoClass> = mutableMapOf()
```
arcuri82, line 41: when having fields that are map, should always have a JavaDoc comment specifying what is the key

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    fun write(testSuitePath: Path, outputFormat: OutputFormat, actionDefinitions: List<Action>) {
+        getDtos(actionDefinitions)
+        dtoCollector.forEach {
+            JavaDtoWriter.write(testSuitePath, outputFormat, it.value)
```
arcuri82: shouldn't here check `outputFormat` to decide if Java version needs to be called?  also, what to do in case of unsupported format? throw an exception maybe?

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    private fun getDtos(actionDefinitions: List<Action>) {
+        actionDefinitions.forEach { action ->
+            action.getViewOfChildren().forEach { child ->
```
arcuri82: instead of a `forEach` with a nested `if`, couldn't just use a `find`?

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+                if (child is BodyParam) {
+                    val primaryGene = GeneUtils.getWrappedValueGene(child.primaryGene())
+                    // TODO: Payloads could also be json arrays, analyze ArrayGene
+                    if (primaryGene is ObjectGene) {
```
arcuri82: what if the body is not required, and so inside a `OptinalGene`?

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+        }
+    }
+
+    private fun getDtoFromObject(gene: ObjectGene, actionName: String) {
+        // TODO: Determine strategy for objects that are not defined as a component and do not have a name
+        val dtoName = gene.refType?:TestWriterUtils.safeVariableName(actionName)
+        val dtoClass = DtoClass(dtoName)
+        gene.fixedFields.forEach { field ->
```
arcuri82, line 71: what about `additionalFields`?  leave a TODO here

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -0,0 +1,90 @@
...
+    private fun getDtoField(fieldName: String, field: Gene?): DtoField {
+        return DtoField(fieldName, when (field) {
+            // TODO: handle nested arrays, objects and extend type system for dto fields
+            is StringGene -> "String"
+            is IntegerGene -> "Integer"
+            is LongGene -> "Long"
+            is DoubleGene -> "Double"
+            is FloatGene -> "Float"
+            is Base64StringGene -> "String"
+            // Time and Date genes will be handled with strings at the moment. In the future we'll evaluate if it's worth having any validation
```
arcuri82: as well as `DateTimeGene` and `RegexGene`

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: Hi @arcuri82! DTO creation will only be executed if the `OutputFormat` is java at the moment (if present in [Main.kt](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/Main.kt#L833-L836)).
In the future will change for `isJavaOrKotlin` and then a more granular check on which dto writer to call will be included here. We could however add a check and exception throw here as well, mainly so as to have a fail-fast if anything changes/breaks in the Main.kt file. Would you prefer that extra check?

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: Shouldn't that be covered by the usage of [GeneUtils.getWrappedValueGene](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt#L773-L774)? Since it handles the `OptionalGene` by looking into it, it should return the body gene eventually.

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: I'm guessing there can be only one `BodyParam` per action, so yes, in that case it would definitely make much more sense to use `find`

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
arcuri82: hi @Pgarrett  yes, extra check here would not hurt :)

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
arcuri82: @Pgarrett  ah, yes, you right, missed that in 53 it is a Param and not a Gene

Pgarrett:
Hi @arcuri82! Whenever you have a moment, please don't forget this PR! Thanks!

arcuri82 (approved):

PR #1259, opened 2025-06-12
Language Model Structured Generate

seran:


core/src/main/kotlin/org/evomaster/core/languagemodel/data/ollama/OllamaResponseFormat.kt
```diff
@@ -0,0 +1,10 @@
+package org.evomaster.core.languagemodel.data.ollama
+
+class OllamaResponseFormat (
```
arcuri82, line 6: need to add some JavaDoc for these properties

arcuri82 (changes requested):

PR #1260, opened 2025-06-13
Ai dependency glm gaussian

MohsenTaheriShalmani:
The GLM classifier is added.
The api is also updated so the EvoMaster sampler can generate both valid and invalid samples.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
@@ -6,6 +6,7 @@ import org.evomaster.core.problem.rest.data.RestCallAction
 import org.evomaster.core.problem.rest.data.RestCallResult
 import org.evomaster.core.problem.rest.classifier.AIModel
 import org.evomaster.core.problem.rest.classifier.AIResponseClassification
+import org.evomaster.core.problem.rest.classifier.GLMOnlineClassifier
 import org.evomaster.core.problem.rest.classifier.GaussianOnlineClassifier
 import javax.annotation.PostConstruct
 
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -17,13 +18,23 @@ class AIResponseClassifier : AIModel {
 
     private lateinit var delegate: AIModel
 
+    // learning rate can be used for different classifiers like glm and NN
+    private var learningRate: Double = 0.01
+    fun setLearningRate(rate: Double) {
+        learningRate = rate
+    }
```
arcuri82: this shouldn't be here. could be a parameter in EMConfig

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -17,13 +18,23 @@ class AIResponseClassifier : AIModel {
...
+
     fun initModel(dimension: Int){
```
arcuri82: likely, this method should be marked with `@PostConstruct`, and `dimension` input removed. also, I would assume each endpoint will end up with different dimensions for the inputs?

arcuri82 (changes requested):

arcuri82:
@MohsenTaheriShalmani  thx! make sure to see that CI is green (ie the build doesn't fail) before asking for a review ;)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
MohsenTaheriShalmani: At the moment, I prefer to avoid using `@PostConstruct` because the classifier initialization depends on the `dimension`, which can vary between endpoints.
However, I agree that this should be properly addressed in the final version.
Based on the current setup, an endpoint has its own classifier, and the classifier has a fixed input dimension.

arcuri82 (approved):

PR #1265, opened 2025-06-23
Ai multi endpoint classifier

MohsenTaheriShalmani:
AIGaussianCheck.kt and AIGlmCheck.kt are updated to show how we can deal with an api with mutiple endpoints


core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -11,46 +11,30 @@ import org.evomaster.core.problem.rest.classifier.GaussianOnlineClassifier
...
-                //TODO
-                delegate = object : AIModel {
-                    override fun updateModel(input: RestCallAction, output: RestCallResult) {}
-                    override fun classify(input: RestCallAction) = AIResponseClassification()
-                }
-            }
+class AIResponseClassifier(
+    private val config: EMConfig = EMConfig(),
+    private val explicitModel: AIModel? = null
+) : AIModel {
```
arcuri82: the injection here shouldn't be touched.  constructor should have no inputs. long story...
recall this class is instantiated as singleton with a DI framework

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -11,46 +11,38 @@ import org.evomaster.core.problem.rest.classifier.GaussianOnlineClassifier
...
 
+
 class AIResponseClassifier : AIModel {
 
     @Inject
-    private lateinit var config : EMConfig
+    private lateinit var config: EMConfig
 
-    private lateinit var delegate: AIModel
+    private lateinit var model: AIModel
```
arcuri82: why changing the name from `delegate` to `model`?  `AIResponseClassifier` itself extends `AIModel`, but delegates its execution to another model.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -11,46 +11,38 @@ import org.evomaster.core.problem.rest.classifier.GaussianOnlineClassifier
...
-        delegate.updateModel(input, output)
+        model.updateModel(input, output)
     }
 
     override fun classify(input: RestCallAction): AIResponseClassification {
-        return delegate.classify(input)
+        return model.classify(input)
     }
 
+    fun getInnerModel(): AIModel = model
```
arcuri82: returning a mutable state reference is usually not a good idea in OO, as it can lead break incapsulation. at times, we need it. in those cases, to make it explicit, i prefer to use the term `view` instead of `get`, eg, `viewInnerModel`

MohsenTaheriShalmani: Tnx, I'll fix it

arcuri82 (approved):

PR #1268, opened 2025-07-02
External pr opensearch

arcuri82:


mmiguerodriguez:
@arcuri82 I wasn't able to detect most of the build errors before pushing my changes. Do you have any suggestion to do so before pushing to the repo?

mmiguerodriguez:
@arcuri82 Thanks for the heads up. 

Do you mind having another look? I fixed the build errors that started showing up, along with manually adding the `OpensearchContainer` implementation since the minimum version would be `1.0.1` or `1.0.1-SNAPSHOT`. 
See [opensearch-testcontainers JDK compatibility](https://github.com/opensearch-project/opensearch-testcontainers?tab=readme-ov-file#testcontainers-and-opensearch-compatibility).

arcuri82:
@mmiguerodriguez  for time being, we build with JDK 8.  as such, should use version `1.0.0` (do not use SNAPSHOT versions for 3rd-party libraries)

mmiguerodriguez:
It should build correctly now, without adding a custom OpensearchContainer class. Let me know if you see something else I could change.

PR #1269, opened 2025-07-02
Add the SQL opertions to the restful API

ytfrank:
In order to test whether evoMaster can insert the data into MySQL DB based on the code analysis, add the SQL operations to the restful API.


arcuri82:
@ytfrank  thx! can you make the PR toward the branch `external-pr-issue-mysql` instead of `master`?  unfortunately, we don't have CI on external PRs

PR #1274, opened 2025-07-08
fixing charset problem

omursahin:


core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -201,7 +201,7 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
     override fun handleVerbEndpoint(baseUrlOfSut: String, _call: HttpWsAction, lines: Lines) {
 
         val call = _call as RestCallAction
-        val verb = call.verb.name.lowercase(Locale.getDefault())
+        val verb = call.verb.name.lowercase(Locale.ENGLISH)
```
arcuri82: looks like several places use `Locale.getDefault()`. can you fix all of the occurrencies in this PR? also, might be good to specify a `setDefault` in the main entry point (ie `Main`) to avoid possible issues in future

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -746,7 +746,7 @@ public final boolean extractSqlDbSchemaAndConstraints(){
                 fkMap.putIfAbsent(t.name, new ArrayList<>());
                 if (t.foreignKeys!=null && !t.foreignKeys.isEmpty()){
                     t.foreignKeys.forEach(f->{
-                        fkMap.get(t.name).add(f.targetTable.toUpperCase(Locale.ENGLISH));
+                        fkMap.get(t.name).add(f.targetTable.toUpperCase(Locale.getDefault()));
```
arcuri82: can just remove the `Locale.getDefault()` in all these cases. ie, calling with no input is the same as calling with `Locale.getDefault()`. ie, simply remove all the `Locale.ENGLISH`, eg `f.targetTable.toUpperCase()`

PR #1276, opened 2025-07-11
python test timeout

omursahin:


core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -289,6 +289,9 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
                 if (bodyParam != null) {
                     lines.append(", data=body")
                 }
+                if(config.testTimeout > 0) {
```
arcuri82, line 292: so the timeout is done in the HTTP call? add a /* */ comment to explain this situation, ie, as timeout at test level does not work reliably in Python, we do timeout as well in each HTTP call

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -344,6 +344,12 @@ abstract class TestCaseWriter {
             lines.indented {
                 lines.add("raise e")
             }
+            if (config.testTimeout > 0) {
+                lines.add("except TimeoutError as e:")
+                lines.indented {
+                    lines.add("unittest.skip(\"Test skipped due to timeout\")")
```
arcuri82: not sure if `skip` is correct here. usually, when a test timeout, it "fails". we might want to do that, and also leave a comment that this is up to discussion.

arcuri82 (approved):

PR #1277, opened 2025-07-11
Redis instrumentation supporting Lettuce and CrudRepository

aszyrej:
- New class replacements added, to support Lettuce integrations (both RedisCommands and RedisAsyncCommands) and CrudRepositories (RedisKeyValueTemplate)
- Unit tests were included calling the replaced methods.
- ExecutionTracer now has the necessary data structures to store Redis commands info.
- ReplacementCategory now includes Redis.


client-java/instrumentation/pom.xml
```diff
@@ -168,6 +168,30 @@
             <artifactId>spring-boot-starter-data-mongodb</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-redis</artifactId>
+            <version>2.7.5</version>
```
arcuri82: version numbers shouldn't be here, but rather in root pom file. make sure to read `docs/for_developers.md`

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/RedisCommand.java
```diff
@@ -0,0 +1,71 @@
...
+public class RedisCommand implements Serializable {
+    /**
+     * Redis commands we'd like to capture. Extendable to any other command in Redis that may
+     */
+    public enum RedisCommandType {
+        GET,
+        HGET,
+        HGETALL,
+        SMEMBERS,
+        ZRANGE
```
arcuri82: add some JavaDoc for all these enums, with brief explanation of each (possibly with url link to Redis documentation). eg, what is difference between `GET` and `HGET`?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/RedisKeyValueTemplateClassReplacement.java
```diff
@@ -0,0 +1,85 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Optional;
+
+/**
+ * This replacement captures operations when Redis is integrated using a CrudRepository.
+ * For other libraries such as Lettuce other replacements were included.
+ * RedisCommandsClassReplacement and RedisAsyncCommandsClassReplacement are examples of this.
+ */
+public class RedisKeyValueTemplateClassReplacement extends ThirdPartyMethodReplacementClass {
```
arcuri82: @jgaleotti  not sure about this... are we going to support Spring directly? wasn't enough to support directly the low level library of Redis?  need some more discussion on this

arcuri82:
hi @aszyrej  thanks for your first PR ;) see comments

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/RedisKeyValueTemplateClassReplacement.java
aszyrej: @arcuri82 I'll review this. If possible we'll change it to only replace low level Redis library.

client-java/instrumentation/pom.xml
aszyrej: Ok, I'll make sure versions are only declared in root pom file

aszyrej: Versions in this file were removed in the last commit (previous to merging master latest updates).
And those versions were added in the root pom file.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/RedisCommand.java
aszyrej: I've added a brief comment from the documentation for each command and then  a link to that specific command in the documentation.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/RedisKeyValueTemplateClassReplacement.java
aszyrej: This is the most important update in the last commit. I changed the strategy for Redis class replacement, and we're now only focusing on StatefulConnection from Lettuce core (which is also used in CrudRepository from Spring).

We no longer support Spring directly, but the low level interface in Redis library (Lettuce in particular).

PR #1279, opened 2025-07-20
Different content type support

omursahin:


client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
```diff
@@ -132,22 +132,23 @@ public static AuthenticationDto getForJsonTokenBearer(
             String payload,
             String extractFromField){
 
-        return  getForJsonToken(dtoName, postEndpoint, payload, extractFromField, "Bearer ");
+        return  getForJsonToken(dtoName, postEndpoint, payload, extractFromField, "Bearer ", "application/json");
     }
 
     public static AuthenticationDto getForJsonToken(
```
arcuri82, line 148: this is a breaking change. need to be careful about this utility functions that are used in the drivers. in this case, to avoid breaking current caller, should keep original signature of method, which call this new method with default parameter, ie, something like:

```
    public static AuthenticationDto getForJsonToken(
            String dtoName,
            String postEndpoint,
            String payload,
            String extractFromField,
            String headerPrefix
    ){
       return getForJsonToken(dtoName, postEndpoint,  ... , "application/json")
```

arcuri82 (changes requested):

arcuri82 (approved):

PR #1280, opened 2025-07-29
AI models generalization

MohsenTaheriShalmani:
The updated `AIGaussianCheck` and `AIGLMCheck` demonstrate that the models are compatible with any API. If the API includes complex endpoints with unsupported genes, the classifier will return `null`, and testing will fall back to random sampling.

The petShop API has been updated to include such an endpoint for validation purposes.

In `AIResponseClassification`, the term `probability` has been replaced by `score`, which is a more general term representing the likelihood of occurrence depending on the selected model.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/AIResponseClassification.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/AIResponseClassification.kt
@@ -2,17 +2,18 @@ package org.evomaster.core.problem.rest.classifier
 
 class AIResponseClassification(
     /**
-     * Map from HTTP Status Code to the probability of getting such code as response.
-     * If code is not defined, then expect probability to be 0.
+     * Map from HTTP Status Code to the scores of getting such code as response.
+     * Score is a positive value reflecting the likelihood or probability of occurrence based on the applied model
+     * If code is not defined, then expect the score as 0.
      */
-    val probabilities : Map<Int, Double> = mapOf()
+    val scores : Map<Int, Double> = mapOf()
 ) {
 
-    fun probabilityOf400() : Double{
+    fun scoreOf400() : Double{
 
-        if(probabilities[400] == null){
+        if(scores[400] == null){
             return 0.0
         }
-        return probabilities[400]!!
+        return scores[400]!!
     }
 }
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/AIResponseClassification.kt
```diff
@@ -2,17 +2,18 @@ package org.evomaster.core.problem.rest.classifier
...
 class AIResponseClassification(
     /**
-     * Map from HTTP Status Code to the probability of getting such code as response.
-     * If code is not defined, then expect probability to be 0.
+     * Map from HTTP Status Code to the scores of getting such code as response.
+     * Score is a positive value reflecting the likelihood or probability of occurrence based on the applied model
+     * If code is not defined, then expect the score as 0.
      */
-    val probabilities : Map<Int, Double> = mapOf()
+    val scores : Map<Int, Double> = mapOf()
```
arcuri82: not sure about this change of name. 
if computing probabilities in [0,1] is problematic, and want to use more generic heuristics score, then need to provide more info on range and semantic of these score values

MohsenTaheriShalmani: I believe probability is not the correct term, as when dealing with continuous numeric data, the probability of observing a single point is always zero. Probability is typically defined over intervals, for example, a < x < b.

I chose score as a general term to allow flexibility for future models. However, I think that **likelihood** is the standard and more appropriate term in this context.

Likelihood is always positive and can be computed for standard models. It's also possible to artificially rescale the likelihood to lie within [0, 1], eventhough this is generally not recommended.

If you agree, I will replace the score with the likelihood.

arcuri82: hi. we are not dealing with continuous numeric data, but categorical data (ie, integer codes in the constrained set 100-599). 
regarding `likelyhood`, based for example on `https://www.reddit.com/r/AskStatistics/comments/18u26cb/what_is_the_difference_between_probability_and/`  the term `probability` might be more correct as we are doing the estimation of an outcome?

MohsenTaheriShalmani: Got it, the idea is to compute the class probabilities. 
So, I need to use the normalized posterior score in the Gaussian model.
I’ll fix it.

arcuri82 (approved):

PR #1281, opened 2025-08-02
External endpoint auth utils

omursahin:


client-java/controller/src/main/java/org/evomaster/client/java/controller/AuthUtils.java
```diff
@@ -156,7 +156,12 @@ public static AuthenticationDto getForJsonToken(
 
         LoginEndpoint le = new LoginEndpoint();
 
-        le.setEndpoint(postEndpoint);
+        if(postEndpoint.startsWith("http://") || postEndpoint.startsWith("https://")){
```
arcuri82, line 174: this might lead to some confusion. add some JavaDoc to this function, to explain how the `postEndpoint` can be interpreted differently based on whether it is a full URL or just a path

arcuri82 (approved):

PR #1284, opened 2025-08-03
Adding unit testing for DTO generation

Pgarrett:
Adding unit tests for DTO generation. The goal is to assert only that given a schema, the expected DTOs are generated in the EM abstraction. Valid code generation of DTOs will be later tested using e2e tests.
Included tests are:
- primitive types in an inline schema
- child object in an inline schema
- simple different objects in the components section
- child object in the components section

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
@@ -66,9 +66,10 @@ class DtoWriter {
 
     private fun getDtoFromObject(gene: ObjectGene, actionName: String) {
         // TODO: Determine strategy for objects that are not defined as a component and do not have a name
+        // TODO: consider an inline schema using more than one possible component: any/one/allOf[object,object]
         val dtoName = gene.refType?:TestWriterUtils.safeVariableName(actionName)
         val dtoClass = DtoClass(dtoName)
-        // TODO: add suport for additionalFields
+        // TODO: add support for additionalFields
         gene.fixedFields.forEach { field ->
             try {
                 val wrappedGene = GeneUtils.getWrappedValueGene(field)
--- a/core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
@@ -1,5 +1,6 @@
 package org.evomaster.core.output.dto
 
+import com.google.common.annotations.VisibleForTesting
 import org.evomaster.core.output.OutputFormat
 import org.evomaster.core.output.TestWriterUtils
 import org.evomaster.core.problem.rest.param.BodyParam
@@ -30,7 +31,7 @@ import java.nio.file.Path
  * Using DTOs provides a major advantage towards code readability and sustainability. It is more flexible and scales
  * better.
  */
-object DtoWriter {
+class DtoWriter {
 
     private val log: Logger = LoggerFactory.getLogger(DtoWriter::class.java)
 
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -102,8 +104,16 @@ object DtoWriter {
...
             else -> throw Exception("Not supported gene at the moment: ${field?.javaClass?.simpleName}")
         })
     }
+
+    /**
+     * @return collected DTOs by the DTO writer. Only for testing, otherwise we're breaking encapsulation.
+     */
+    @VisibleForTesting
+    internal fun getCollectedDtos() : Map<String, DtoClass> {
+        return dtoCollector
```
arcuri82, line 117: i didn't realize it before, but isn't `dtoCollector` a global mutable static state? this create possible issues. either should be removed (and passed as input between the function that first instantiate and then use it), or this class should be refactored into a singleton service

core/src/test/kotlin/org/evomaster/core/output/dto/DtoWriterTest.kt
```diff
@@ -0,0 +1,176 @@
...
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+
+class DtoWriterTest {
+
+    companion object {
+        val tmpTestSuitePath: Path = Paths.get(Files.createTempDirectory("dto-writer-test").toUri())
```
arcuri82: rather then a tmp folder, should be somewhere under `target`, so that it is easy to double-check, and `maven clean` will get rid of it

arcuri82 (changes requested):

core/src/test/kotlin/org/evomaster/core/output/dto/DtoWriterTest.kt
Pgarrett: Hi @arcuri82! Hm, I was actually thinking these tests more for testing the data structures generated in memory and test the `DtoWriter` class, in order to assert that the actions and DTOs match a provided spec. 
Generated IO content is not that relevant here, instead for that I was thinking:
- E2E tests
- Unit tests focused on syntax testing the `JavaDtoWriter` and future `KotlinDtoWriter`

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: No, but it used to be. `DtoWriter` is now a class instead of an object, so the `dtoCollector` is now created every time `DtoWriter()` is instantiated. I actually faced that issue when I started writing my code, realized and changed from object to class

arcuri82: ah, make sense

core/src/test/kotlin/org/evomaster/core/output/dto/DtoWriterTest.kt
arcuri82: but, using `tmpTestSuitePath`, aren't you writing to disk? any tmp file created during a test should go somewhere under `./target`.

arcuri82 (changes requested):

core/src/test/kotlin/org/evomaster/core/output/dto/DtoWriterTest.kt
Pgarrett: Done!

PR #1285, opened 2025-08-04
enable to override externalEndpointURL

arcuri82:


omursahin:
@arcuri82 finally test passed

PR #1288, opened 2025-08-11
Support arrays as DTO payload

Pgarrett:


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -89,8 +101,21 @@ class DtoWriter {
         dtoCollector.put(dtoName, dtoClass)
     }
 
+    private fun getDtoFromArray(gene: ArrayGene<*>, actionName: String) {
```
arcuri82: the name is bit confusing. it is a `get`, but the method does not return anything. if this function has side-effects on state, should use different name than `get`

Pgarrett: Yes, you're right. Renamed it to `calculate`

arcuri82 (approved):

PR #1290, opened 2025-08-18
Manual input identification for SSRF

seran:


core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -63,13 +59,18 @@ class SSRFAnalyser {
      */
     private lateinit var individualsInSolution: List<EvaluatedIndividual<RestIndividual>>
 
+    private val urlRegexPattern: Regex = Regex("\\burl\\b")
```
arcuri82: is this ignoring case? eg, would not match for URL or Url. 
also, add comment specifying what `\b` is

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -184,59 +211,47 @@ class SSRFAnalyser {
...
-     * TODO: Classify based on manual
-     * TODO: Need to rename the word manual to something meaningful later
+     * A private method to identify parameter is a potentially holds URL value
+     * using a Regex based approach.
      */
-    private fun manualClassifier() {
-        // TODO: Can use the extracted CSV to map the parameter name
-        //  to the vulnerability class.
+    private fun manualClassifier(name: String, description: String? = null): Boolean {
+        if (potentialUrlParamNames.contains(name)) {
```
arcuri82: is this ignoring case?

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -184,59 +211,47 @@ class SSRFAnalyser {
...
+            return true
+        }
+
+        if (name.lowercase().matches(urlRegexPattern)) {
```
arcuri82: 2 potential issues here:

1) if `urlRegexPattern` could handle cases, no need to waste computation and memory to call `lowercase()`

2) is the regex correct? for name, it would not be a word boundary with \b. rather want to see url/URL as prefix/postfix, isn;t it?

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -184,59 +211,47 @@ class SSRFAnalyser {
...
+            return true
+        }
+        if (description != null) {
+            if (description.lowercase().matches(urlRegexPattern)) {
```
arcuri82: `description.lowercase()` is expensive if string is long. recall it creates a new string allocation on the heap and throw it away soon this if statement is completed

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
seran: Before comparing I'm making the String lowercase, because of that I'm not checking the case at this point.

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
seran: I'll update the Regex and remove the `lowercase()`

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -135,14 +133,25 @@ class SSRFAnalyser {
...
-            val genes = getStringGenesFromParam(param.seeGenes())
-            genes.any { gene ->
+        val hasCallBackURL = GeneUtils
+            .getAllStringFields(action.parameters)
+            .any { gene ->
                 httpCallbackVerifier.isCallbackURL(gene.getValueAsRawString())
             }
+
+        if (hasCallBackURL) {
+            // FIXME: When the code reaches this point during SSRF phase
```
seran: Hi @arcuri82,

I can see the code reach the AbstractRestFitness during the SSRF detection phase. However, WireMock seems to be null at this point although it was initiated earlier. I'm not sure why this is happening, need your assistance to debug this further.

core/src/main/kotlin/org/evomaster/core/problem/security/service/HttpCallbackVerifier.kt
```diff
@@ -75,7 +80,9 @@ class HttpCallbackVerifier {
      * Method generates a unique callback link to be used as payload for SSRF.
      */
     fun generateCallbackLink(name: String): String {
-        val ssrfPath = "/sink/${counter++}"
+        // FIXME: sink/EM_0 <- slash get replaced with a comma at some point, which fails
```
seran, line 90: This was a problem too, for some reason slash after sink get replaced with a comma when recomputing the fitness.

PR #1294, opened 2025-08-19
initial use of model accuracy

arcuri82:


core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -54,15 +60,31 @@ class AIResponseClassifier : AIModel {
...
+        return delegate.estimateAccuracy(endpoint)
+    }
+
     fun viewInnerModel(): AIModel = delegate
 
     /**
      * If the model thinks this call will lead to a user error (eg 400), then try to repair
      * the action to be able to solve the input constraints, aiming for a 2xx.
      * There is no guarantee that this will work.
      */
```
MohsenTaheriShalmani, line 73: I have no idea about the deterministic model, but I believe repairing could be problematic for the Gaussian and GLM, as it forces the model to change its structure based on its own prediction, which could be an incorrect prediction. Further, I think applying a repair mechanism to a neural network is virtually impossible. 
I prefer to use randomness.nextBoolean(accuracy) on the updating process of classifier.updateModel

arcuri82: hi. the repair is on the action, not the model. i ll make a first implementation (shouldn't take too long). we can discuss it then tomorrow

arcuri82: @MohsenTaheriShalmani  look at the commit I just made. let's discuss it tomorrow

PR #1297, opened 2025-08-21
Excluding endpoints during the search

omursahin:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1101,6 +1101,13 @@ class EMConfig {
             " If no tag is specified here, then such filter is not applied.")
     var endpointTagFilter: String? = null
 
+    @Important(5.2)
```
arcuri82, line 1101: all numbers should be unique, to guarantee deterministic sorting. put a `5.3`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1301, opened 2025-08-25
OpenSearch Distance Calculation Feature

mmiguerodriguez:
This branch extends EvoMaster with distance calculation support for OpenSearch queries.

### Features
- Added OPENSEARCH to Heuristic DTO Entry type.

**Distance Data Structures**
- OpenSearchDistanceWithMetrics: stores distance and number of evaluated documents (non-negative enforced).
- OpenSearchCommandWithDistance: pairs commands with their computed distance.

**OpenSearchHandler**
- Captures OpenSearch queries during execution.
- Retrieves documents from indices.
- Computes query–document distances and records minimum distance + evaluated count.
- Term Query distance calculation (integers).

### Testing
- Example Age app demonstrating equals and range logic.


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -323,6 +327,17 @@ public final void initMongoHandler() {
         }
     }
 
+    public final void initOpenSearchHandler() {
```
arcuri82, line 330: it is fine in this PR. but, as this code seems very similar to what done for Mongo, maybe @jgaleotti  you can refactor it to avoid duplication? 
@mmiguerodriguez   maybe leave a `// TODO` comment stating this code likely needs refactoring with the Mongo one

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -467,6 +490,34 @@ public final void computeMongoHeuristics(ExtraHeuristicsDto dto, List<Additional
...
 
+    public final void computeOpenSearchHeuristics(ExtraHeuristicsDto dto, List<AdditionalInfo> additionalInfoList) {
+        if (openSearchHandler.isCalculateHeuristics()) {
+            if (!additionalInfoList.isEmpty()) {
+                AdditionalInfo last = additionalInfoList.get(additionalInfoList.size() - 1);
+                last.getOpenSearchInfoData().forEach(it -> {
+                    try {
+                        openSearchHandler.handle(it);
+                    } catch (Exception e){
+                        SimpleLogger.error("FAILED TO HANDLE OPENSEARCH COMMAND");
```
arcuri82: add info on the error in this log, eg, `+ e.getMssage()`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/OpenSearchDistanceWithMetrics.java
```diff
@@ -0,0 +1,27 @@
+package org.evomaster.client.java.controller.internal.db;
+
+public class OpenSearchDistanceWithMetrics {
+
+    public final double distance;
+
+    public final int numberOfEvaluatedDocuments;
```
arcuri82: those fields should be private, as anyway you are providing public getters

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/OpenSearchHeuristicsCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+    private QueryOperation getOperation(Object query) {
+        return new OpenSearchQueryParser().parse(query);
+    }
+
+    private double calculateDistance(QueryOperation operation, Object doc) {
+        if (operation instanceof TermOperation<?>) {
+            return calculateDistanceForEquals((TermOperation<?>) operation, doc);
+        }
+
+        return 0;
```
arcuri82: is it correct to return 0 here? if so, it is not so obvious. add some /* */ comments to clarify it

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/operations/ComparisonOperation.java
```diff
@@ -0,0 +1,19 @@
+package org.evomaster.client.java.controller.opensearch.operations;
+
+public abstract class ComparisonOperation<V> extends QueryOperation {
+    private final String fieldName;
+    private final V value;
+
+    ComparisonOperation(String fieldName, V value) {
```
arcuri82: default-scope in Java is very tricky, and possibly counter-intuitive. as such, I usually avoid it. consider if a `protected` makes sense here

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/selectors/QuerySelector.java
```diff
@@ -0,0 +1,37 @@
...
+ * Each selector maps to a unique operation.
+ */
+abstract public class QuerySelector {
+    /**
+     * Returns the operation mapped to the selector if the query satisfy all syntax checks or null if not.
+     */
+    public abstract QueryOperation getOperation(Object query);
+
+    protected Boolean isUniqueEntry(Map<?, ?> map) {
+        return map.size() == 1;
```
arcuri82: what is input map is null?

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/utils/OpenSearchQueryHelper.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Extracts the value from a field by its type (Double, Long, Boolean, String, Null).
+     */
+    public static Object extractTypedFieldValue(Object value) {
+        try {
+            Object kind = value.getClass().getMethod("_kind").invoke(value);
+            String kindName = (String) kind.getClass().getMethod("name").invoke(kind);
+
+            switch (kindName) {
```
arcuri82, line 58: what about `Integer` and `Short`?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/OpenSearchCommand.java
```diff
@@ -1,12 +1,13 @@
...
 
 import java.io.Serializable;
+import java.util.List;
 
 public class OpenSearchCommand implements Serializable {
     /**
      * Name of the index that the operation was applied to
      */
-    private final Object index;
+    private final List<String> index;
```
arcuri82, line 11: update docs based on new type. ie, how can a list of strings be a name?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/OpenSearchClientClassReplacement.java
```diff
@@ -71,13 +73,32 @@ private static Method retrieveMethod(Object openSearchClient, String id) {
...
+            if (result == null) {
+                return null;
+            }
+
+            if (result instanceof List) {
+                return (List<String>) result;
+            } else if (result instanceof String) {
+                return Collections.singletonList((String) result);
+            } else {
+                return Collections.singletonList(result.toString());
```
arcuri82, line 103: is this last case expected in normal behaviour? or is it just a safety check to not crash EM? in this latter case, might want to log such event (so we can see when/if it happens, and fix it if needed)

e2e-tests/spring-rest-opensearch/src/main/java/com/opensearch/age/OpenSearchAgeRest.java
```diff
@@ -0,0 +1,69 @@
...
+        Age s = new Age(18);
+        ageRepo.index(s);
+    }
+
+    @GetMapping("{age}")
+    public List<Age> findByAge(@PathVariable("age") Integer age) throws IOException {
+        return ageRepo.findByAge(age);
+    }
+
+//    @GetMapping("gte/{gte}")
```
arcuri82, line 30: is this code no longer needed? or to be put back in future? either remove it, or add a TODO comment if you ll need it in future

e2e-tests/spring-rest-opensearch/src/test/java/org/evomaster/e2etests/spring/rest/opensearch/age/OpenSearchAgeEMTest.java
```diff
@@ -0,0 +1,35 @@
...
+            "OpenSearchAgeEM",
+            "org.foo.spring.rest.opensearch.OpenSearchAgeEM",
+            1000,
+            true,
+            (args) -> {
+                setOption(args, "instrumentMR_OPENSEARCH", "true");
+
+                Solution<RestIndividual> solution = initAndRun(args);
+
+//                assertHasAtLeastOne(solution, HttpVerb.GET, 404, "/age/{q}", null);
```
arcuri82, line 32: why no assertions? if the techniques work, when you should be able to cover the different endpoints you have designed. if not yet, and tehcniques needs more fixing, it is ok to have a PR now, but: (1) add the needed assertions on the different endpoints; (2) mark the test with `@Disabled`; (3) leave a TODO comment stating this is failing and need more work

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
jgaleotti: please leave a TODO and when the Redis and the OpenSearch implementations are done I will refactor them out.

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/OpenSearchHeuristicsCalculator.java
jgaleotti: I guess this should be Double.MAX_VALUE as this is the case where it is not supported, right?

e2e-tests/spring-rest-opensearch/src/main/java/com/opensearch/age/OpenSearchAgeRest.java
mmiguerodriguez: Will be put back when testing other types of queries (eg: Range). Will add TODO comment.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/OpenSearchClientClassReplacement.java
mmiguerodriguez: I made it this way to consider all cases, but the normal behavior would be for the index to be a single String.

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/utils/OpenSearchQueryHelper.java
mmiguerodriguez: I used OpenSearch [FieldValue](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/FieldValue.java#L74) as reference. Will review if more are required.

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/OpenSearchHeuristicsCalculator.java
mmiguerodriguez: Yes, will modify this.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
mmiguerodriguez: Will add TODO comment regarding refactors

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/OpenSearchCommand.java
mmiguerodriguez: Yeah, this might be confusing, will review how the correct naming should be. 
I think that for an initial implementation we should consider only simple Strings, without lists.

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/selectors/QuerySelector.java
mmiguerodriguez: Removed this method since unused

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/OpenSearchCommand.java
mmiguerodriguez: Updated docs following OpenSearch SearchRequest docs.
```
    /**
     * A comma-separated list of index names to search; use <code>_all</code> or
     * empty string to perform the operation on all indices
     * <p>
     * API name: {@code index}
     */
```

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/OpenSearchClientClassReplacement.java
mmiguerodriguez: Added logs

client-java/controller/src/main/java/org/evomaster/client/java/controller/opensearch/utils/OpenSearchQueryHelper.java
```diff
@@ -0,0 +1,90 @@
...
+                case "String":
+                    if ((Boolean) value.getClass().getMethod("isString").invoke(value)) {
+                        return value.getClass().getMethod("stringValue").invoke(value);
+                    }
+                    break;
+                case "Null":
+                    return null;
+            }
+
+            SimpleLogger.debug("[OpenSearch] [extractTypedFieldValue] found unknown kind: " + kindName);
```
arcuri82: should be a `warn` not `debug`

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/OpenSearchClientClassReplacement.java
```diff
@@ -71,13 +74,34 @@ private static Method retrieveMethod(Object openSearchClient, String id) {
...
+                return null;
+            }
+
+            if (result instanceof String) {
+                return Collections.singletonList((String) result);
+            } else if (result instanceof List) {
+                SimpleLogger.debug("[OpenSearch] [getIndex] returned list of length: " + ((List<?>) result).size());
+                return (List<String>) result;
+            } else {
+                SimpleLogger.debug("[OpenSearch] [getIndex] returned unknown type: " + result.getClass().getName());
```
arcuri82: should be a `warn` not `debug`

arcuri82 (changes requested):

mmiguerodriguez:
I've addressed all of the comments. Regarding the question from https://github.com/WebFuzzing/EvoMaster/pull/1301#discussion_r2310769393, at least for filtering those are the minimal types that should be mapped.

arcuri82 (approved):

PR #1302, opened 2025-08-26
AI models encoder update

MohsenTaheriShalmani:
The encoder has been updated to handle null values.

The models are applied to the SUTs
1. Basic API
2. Multitype API

The results are promising.

Please run `AIGaussianCheck.kt` and `AIGLMCheck.kt` to see the performance.


core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -36,18 +36,37 @@ class GaussianOnlineClassifier : AIModel {
     private var density200: Density? = null
     private var density400: Density? = null
 
+    // classifier state
+    private var cp: Int = 0        // correct predictions
```
arcuri82: 2 things:
1) when describing fields, do not use //, but rather JavaDoc style /** */ (note the first 2 *). this is because documentation is generated, and will be visible when howevering over the variable with mouse in an IDE like IntelliJ
2) do not us acronyms for object level fields. it makes the code harder to read. short acronyms make sense when are for local variables inside a block with short lifespan. the above can be rewritten in: `private var correctPredictions: Int = 0`. note that IntelliJ will autocomplete variable when you start typing them

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -36,18 +36,37 @@ class GaussianOnlineClassifier : AIModel {
...
+    private var ac: Double = 0.0   // accuracy ratio
+
     fun setDimension(d: Int) {
         require(d > 0) { "Dimension must be positive." }
         this.dimension = d
         this.density200 = Density(d)
         this.density400 = Density(d)
+        this.cp = 0
+        this.tot = 1
+        this.ac = 0.0
```
arcuri82: as this can be useful by different models, shouldn't be abstracted into its own class, eg something like `ModelAccuracy`, so it can be re-used in the different model implementations?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -36,18 +36,37 @@ class GaussianOnlineClassifier : AIModel {
...
     }
 
     fun getDimension(): Int {
         check(this.dimension != null) { "Classifier not initialized. Call setDimension first." }
         return this.dimension!!
     }
 
+    fun setCp(cp: Int) {
```
arcuri82: these setters and getters are redundant in Kotlin, as automatically generated by compiler if you declare it like `var correctPredictions: Int = 0` (note the not use of `private` here, as it seems you need get and set both to be public).  have a read at: [https://kotlinlang.org/docs/properties.html#custom-getters-and-setters](https://kotlinlang.org/docs/properties.html#custom-getters-and-setters)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -36,18 +36,37 @@ class GaussianOnlineClassifier : AIModel {
...
+        this.cp = cp
+    }
+    fun setTot(tot: Int){
+        this.tot = tot
+    }
+
+    fun getCorrectPredictions(): Int = cp
+    fun getTotalRequests(): Int = tot
+    fun getAccuracy(): Double = if (tot == 0) 0.0 else cp.toDouble() / tot
```
arcuri82: you keep a variable for `ac` but then recompute it on the fly here? or is the field `ac` redundant?  note that this logic can integrated inside a `ModelAccuracy` to enable reuse

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -100,7 +119,12 @@ class GaussianOnlineClassifier : AIModel {
     }
 
     override fun estimateAccuracy(endpoint: Endpoint): Double {
-        TODO("Not yet implemented")
+        // Strict accuracy
+        //TODO this accuracy can be dynamic for example based on the last 100 requests
+        if (this.tot == 0) {
+            return 0.0
+        }
+        return this.cp.toDouble() / this.tot
```
arcuri82: what is the difference of this with `getAccuracy`? shouldn't this method just call instead of duplicating the logic?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GeneralizedLinearModel.kt
```diff
@@ -26,12 +26,20 @@ class GLMOnlineClassifier(
     private var weights: MutableList<Double>? = null
     private var bias: Double = 0.0
 
+    // classifier state
+    private var cp: Int = 0        // correct predictions
+    private var tot: Int = 1       // total requests
+    private var ac: Double = 0.0   // accuracy ratio
```
arcuri82: if you are using exactly the same code as in another class, that is good hint for 2 possible solutions: (1) a shared parent class in which the common logic is pushed up; (2) or extract the functionality in a new class instantiated in both places (eg `ModelAccuracy`)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GeneralizedLinearModel.kt
```diff
@@ -44,8 +52,19 @@ class GLMOnlineClassifier(
...
+    fun setCp(cp: Int) {
+        this.cp = cp
+    }
+    fun setTot(tot: Int){
+        this.tot = tot
+    }
+
+    fun getCorrectPredictions(): Int = cp
+    fun getTotalRequests(): Int = tot
+    fun getAccuracy(): Double = if (tot == 0) 0.0 else cp.toDouble() / tot
```
arcuri82: duplicated code should be avoiaded.

core/src/test/kotlin/org/evomaster/core/output/TestSuiteOrganizerTest.kt
```diff
@@ -1,239 +0,0 @@
-package org.evomaster.core.output
```
arcuri82: why deleting the content of this test???

MohsenTaheriShalmani: Honestly, I have no idea. Could it be because of the manual merge strategy that I chose?
Anyway, sorry about that.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/AIResponseClassification.kt
```diff
@@ -15,6 +15,13 @@ class AIResponseClassification(
                         " But status code ${it.key} has probability value ${it.value}")
             }
         }
+
+        val sum = probabilities.values.sum()
+        if (kotlin.math.abs(sum - 1.0) > 1e-6) {
```
arcuri82: i would change this check. not to be ==1, but rather <1, with comments to explain why. reason is that, when we classify for a specific output, we might not want or able to classify correctly for all other status codes. might want to classify just for some. for other, it might be unknown or not computed. regardless, sum of these probabilities should never be above 1

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ClassifierPerformance.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Classifier state
+ *
+ * @param correctPrediction the number of correct predictions made by the classifier
+ * @param totalSentRequests the total number of requests sent to the classifier
+ */
+class ClassifierPerformance(
+    val correctPrediction: Int,
+    val totalSentRequests: Int
```
arcuri82: i m all for immutable object... but this one is going to change a lot, at each new classifiation. creating new instance each time of update will be inefficient. better to use `private var` here outside of constructor

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/GaussianOnlineClassifier.kt
```diff
@@ -99,8 +83,38 @@ class GaussianOnlineClassifier : AIModel {
...
+        } else {
+            val predicted = classify(input).prediction()
+            performance.updatePerformance(predicted == trueStatusCode)
+        }
+
+        /**
+         * Updating the density functions based on the real observation
+         */
+        when (trueStatusCode) {
+            200 -> this.density200!!.update(inputVector)
```
arcuri82: recall that "success" is the whole 200-299 range

arcuri82: in the deterministic classifier, im not going to deal with 2xx. but, if needed here, we might need to change some datastructure to handle it better (but could be done in a separated branch/PR)

arcuri82 (changes requested):

core-it/src/test/kotlin/org/evomaster/core/problem/rest/aiclassification/AIGLMCheck.kt
```diff
@@ -198,7 +198,7 @@ class AIGLMCheck : IntegrationTestRestBase() {
 
             // Probabilistic decision-making based on Bernoulli(prob = aci)
             val sendOrNot: Boolean
-            if (predictionOfStatusCode != 400) {
+            if (predictionOfStatusCode !in 400..499) {
```
arcuri82: just keep in mind that, for several status code like 403 and 404, the output depends on database state, for which we have no info here in the classifier. so that was a motivation to deal just with 400 in the classification

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ClassifierPerformance.kt
```diff
@@ -2,14 +2,16 @@ package org.evomaster.core.problem.rest.classifier
...
-    val correctPrediction: Int,
-    val totalSentRequests: Int
-) {
+class ClassifierPerformance {
+
+    private var correctPrediction: Int = 0
+    private var totalSentRequests: Int = 1
+
+    val getCorrectPrediction: Int get() = correctPrediction
+    val getTotalSentRequests: Int get() = totalSentRequests
```
arcuri82: in Kotlin, this can be simplified with exactly same results:
```
var correctPrediction: Int = 0
     private set  
```

core-it/src/test/kotlin/org/evomaster/core/problem/rest/aiclassification/AIGLMCheck.kt
arcuri82: might be easier to discuss this in details in our next meeting

arcuri82:
@MohsenTaheriShalmani  hi. thx. there are conflicts now after some changes i made. pull from master into this branch, resolve the conflicts, push, and then ask for my review again

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ClassifierPerformance.kt
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Classifier state
+ * @property correctPrediction the number of correct predictions made by the classifier
+ * @property totalSentRequests the total number of requests sent to the classifier
+ */
+class ClassifierPerformance {
+
+    var correctPrediction: Int = 0
+        private set
+    var totalSentRequests: Int = 1
```
arcuri82: why does it start with 1 and not 0?

MohsenTaheriShalmani: For two reasons
1. To avoid division by zero 
2. To make sure accuracy never reaches 100% as correctPrediction is always at least one unit less than totalSentRequests

arcuri82: ok, that makes sense. but, then, `totalSentRequests` does not represent the total number of sent requests. It should be either renamed into `totalSentRequestsPlusOne` or, on line 21,  should divide by `totalSentRequests + 1` . also, as it is now, isn't the `if` statement in line 20 always true?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ClassifierPerformance.kt
MohsenTaheriShalmani: Yes, you're right. Based on the `if` statement `totalSentRequests` can start from 0

PR #1303, opened 2025-08-26
setValueBasedOn for other Genes

seran:


core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
```diff
@@ -295,6 +296,42 @@ class ArrayGene<T>(
...
     }
 
+    @Deprecated("Do not call directly outside this package. Call setFromStringValue")
+    /**
+     * To set the Array children from a String.
+     * Use [template] to create child [Gene].
+     * Use comma (,) separated elements as String.
+     */
+    override fun setValueBasedOn(value: String): Boolean {
+        val elements = value.split(",")
```
arcuri82: the split should be based on `separatorTag`, taking care of trimming spaces

arcuri82: also, might have to consider `openingTag` and `closingTag` as well

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
```diff
@@ -295,6 +296,42 @@ class ArrayGene<T>(
...
+                is BooleanGene -> {
+                    addChildren(
+                        elements.map {
+                            BooleanGene(name, it.toBoolean())
+                                .apply { doInitialize(getSearchGlobalState()?.randomness)}
+                        }.toList()
+                    )
+                }
+                else -> {
+                    // TODO: Handle other types
```
arcuri82, line 331: don't silently ignore it. either warn log it, or throw an exception

PR #1304, opened 2025-08-27
Added support for anyOf/oneOf/allOf for DTOs

Pgarrett:
API Specs with `anyOf`, `oneOf` and `allOf` now generate DTOs. For each of them, the resulting DTO is a merge of all the options in the spec


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -56,29 +56,66 @@ class DtoWriter {
...
-                when (primaryGene) {
-                    is ObjectGene -> {
-                        calculateDtoFromObject(primaryGene, action.getName())
-                    }
-                    is ArrayGene<*> -> {
-                        calculateDtoFromArray(primaryGene, action.getName())
-                    }
-                    else -> {
-                        throw IllegalStateException("Gene $primaryGene is not supported for DTO payloads")
+                val primaryGene = (it as BodyParam).primaryGene()
```
arcuri82, line 60: what happens if the body is optional, and the primary gene is a wrapper `OptionalGene`?

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -56,29 +56,66 @@ class DtoWriter {
...
         }
+        dtoCollector.put(dtoName, dtoClass)
+    }
+
+    private fun calculateDtoFromNonChoiceGene(gene: Gene, actionName: String) {
+        when (gene) {
+            is ObjectGene -> calculateDtoFromObject(gene, actionName)
+            is ArrayGene<*> -> calculateDtoFromArray(gene, actionName)
+            else -> {
+                throw IllegalStateException("Gene $gene is not supported for DTO payloads")
```
arcuri82: add some more info to help debugging if it happens, eg put the `actionName` somewhere in this error message

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -56,29 +56,66 @@ class DtoWriter {
...
+    }
+
+    private fun calculateDtoFromArray(gene: ArrayGene<*>, actionName: String) {
+        val template = gene.template
+        // TODO consider ChoiceGene. Primitive types won't be considered, an array of strings should not be wrapped
+        //  into a DTO but just use List<String> for setting the payload.
+        if (template is ObjectGene) {
+            calculateDtoFromObject(template, actionName)
+        } else {
+            log.debug("Arrays of non custom objects are not collected as DTOs")
```
arcuri82: maybe use a unique warn? `LoggingUtil.uniqueWarn`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: Good question, which would be the best way to handle those cases? I saw you removed the `getWrappedValueGene` function. The `getLeafNode` of a `ChoiceGene` wrapped in an `OptionalGene` returns only one of the options

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: Added! This exception throwing should actually be short lived, I was thinking it's main purpose will be providing info for when we execute against Dataset to see which cases we're missing

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: Done! Same explanation as above

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
Pgarrett: @arcuri82

arcuri82: if it is a `ChoiceGene` inside an `OptionalGene`, you will not see it here. `ChoiceGene` complicates a few things, not just here. so in long run i need to think about some new functionality/fixes for it. for now, maybe an approach could be that, on `primaryGene`, you do a call on `getWrappedGene(ChoiceGene::class.java)`, and see if null or not

Pgarrett: Done!

PR #1305, opened 2025-08-28
Redis instrumentation supporting Lettuce and CrudRepository

aszyrej:
- New class replacement added for class io.lettuce.core.api.StatefulConnection. Both Spring CrudRepository and Lettuce integrations use this class, so this single class replacement will work with both solutions. The replaced method dispatch should capture every single Redis command that's being used.

- Unit tests were included calling the replaced method (dispatch).
- Integration tests were added calling both a CrudRepository and a Lettuce connection.

- ExecutionTracer now has the necessary data structures to store Redis commands info.

- ReplacementCategory now includes Redis.
- ReplacementList now includes the new replacement.


aszyrej:
@arcuri82 I've created this new pull request and closed the previous one, in order to keep a cleaner commit history.
I've addressed the things we discussed and replied to your comments in the closed pull request.

PR #1308, opened 2025-08-29
SSRF Test Writer

seran:


core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionStubMapping.kt
```diff
@@ -0,0 +1,14 @@
...
+class ActionStubMapping (
+    val id: Long,
+    val actionName: String,
+    val stub: String,
+    val port: Int,
+    val url: String
+) {
+
+    fun getVerifierName(): String {
+        return "httpCallbackVerifier${id}"
```
arcuri82, line 16: is this related to constant variable `httpCallbackVerifierName` ? if so, should reuse instead of duplicate the same string prefix

core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionStubMapping.kt
```diff
@@ -0,0 +1,14 @@
+package org.evomaster.core.problem.security.data
+
+class ActionStubMapping (
```
arcuri82, line 7: add some JavaDoc for this class

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -80,17 +90,16 @@ class SSRFAnalyser {
...
+        LoggingUtil.getInfoLogger().info("Applying {}", SSRFAnalyser::class.simpleName)
 
         // extract individuals from the archive
         val individuals = this.archive.extractSolution().individuals
 
         individualsInSolution =
             RestIndividualSelectorUtils.findIndividuals(
                 individuals,
-                statusCodes = listOf(200, 201)
+                statusCodes = listOf(200, 201, 204)
```
arcuri82: why these specific codes and not generic 2xx?

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -101,25 +110,23 @@ class SSRFAnalyser {
...
-        }
-
         // execute
         analyse()
 
         // TODO: This is for development, remove it later
         val individualsAfterExecution = RestIndividualSelectorUtils.findIndividuals(
             this.archive.extractSolution().individuals,
-            statusCodes = listOf(200, 201)
+            statusCodes = listOf(200, 201, 204)
```
arcuri82: isn't this duplicating what at line 102? if really need, should refactor out in a common constant list to reuse

core/src/main/kotlin/org/evomaster/core/problem/security/data/ActionStubMapping.kt
seran: failed to clean up `httpCallbackVerifierName `

PR #1317, opened 2025-09-07
Ai model update for testing

MohsenTaheriShalmani:
The encoder is updated to handle ObjectGene
Some additional AI models were added for performance comparison


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1286,7 +1286,19 @@ class EMConfig {
...
+         * "NONE: No Classification"
+         * "GAUSSIAN: Gaussian"
+         * "GM: Gaussian Mixture"
+         * "KDE: Kernel Density Estimation"
+         * "KNN: K-Nearest Neighbors"
+         * "NN: Neural Network"
+         * "GLM: Generalized Linear Model"
+         * "DETERMINISTIC: Rule-Based Deterministic Model
+         */
+        NONE, GAUSSIAN, GM, KDE, KNN, NN, GLM, DETERMINISTIC
```
arcuri82: this `/** */` JavaDoc would apply to the entry `NONE`, not the others, isn't it? should be moved on the `enum` declaration, or split each line in a new `/** */` on each enum entry

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/InputEncoderUtils.kt
```diff
@@ -1,81 +1,120 @@
...
+import org.evomaster.core.search.gene.string.StringGene
+import kotlin.reflect.KClass
 
 /**
  * Utility object for encoding the genes of a [RestCallAction] into a numerical representation.
  * This is primarily used in AI-based models that require fixed-length numeric feature vectors
  * as input.
  */
-object InputEncoderUtils {
+class InputEncoderUtils(
```
arcuri82: `...Utils` postfix is usually for sets of static, state-free utility functions. but here create instance instead. to avoid confusion, should be renamed. maybe something like `InputEncoderUtilWrapper`?

arcuri82 (changes requested):

arcuri82 (approved):

PR #1322, opened 2025-09-10
DTOs used in test case writing

Pgarrett:
DTOs for Java payloads:
- Created `DtoOutput` interface to handle the different language specific writing
- DTOs are now being used as the payload and instantiated in the test case source
- Mapping of `Gene` to DTO is in a specific class to avoid adding extra complexity to `Gene`s.

Example test cases generated:
```
@Test(timeout = 60000)
public void test_1_postOnArrayReturns200() throws Exception {
    
    List<Person> list_Person_1 = new ArrayList<Person>();
    Person dto_Person_1 = new Person();
    dto_Person_1.setName("Rz3VEN6");
    dto_Person_1.setAge(525);
    Address dto_Address_1 = new Address();
    dto_Address_1.setStreet("sawnfEXRe03");
    dto_Address_1.setCountry("yAqG3aF");
    dto_Person_1.setAddress(dto_Address_1);
    list_Person_1.add(dto_Person_1);
    
    given().accept("application/json")
            .body(list_Person_1)
            .post(baseUrlOfSut + "/array")
            .then()
            .statusCode(200)
            .assertThat()
            .body("'data'.size()", equalTo(1))
            .body("'data'[0].'address'.'country'", containsString("yAqG3aF"))
            .body("'data'[0].'address'.'street'", containsString("sawnfEXRe03"))
            .body("'data'[0].'age'", numberMatches(525.0))
            .body("'data'[0].'name'", containsString("Rz3VEN6"))
            .body("'message'", containsString("Received array"));
}

@Test(timeout = 60000)
public void test_2_postOnObjectReturns200() throws Exception {
    
    Person dto_Person_1 = new Person();
    dto_Person_1.setName("jtHq5faoLSZ");
    dto_Person_1.setAge(231);
    Address dto_Address_1 = new Address();
    dto_Address_1.setStreet("qeBfTWn6kE");
    dto_Address_1.setCountry("dyWzxJCTsPmg");
    dto_Address_1.setVoted(false);
    dto_Person_1.setAddress(dto_Address_1);
    
    given().accept("application/json")
            .body(dto_Person_1)
            .post(baseUrlOfSut + "/object")
            .then()
            .statusCode(200)
            .assertThat()
            .body("'data'.'address'.'country'", containsString("dyWzxJCTsPmg"))
            .body("'data'.'address'.'street'", containsString("qeBfTWn6kE"))
            .body("'data'.'address'.'voted'", equalTo(false))
            .body("'data'.'age'", numberMatches(231.0))
            .body("'data'.'name'", containsString("jtHq5faoLSZ"))
            .body("'message'", containsString("Received object"));
}
```


core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -113,6 +117,43 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
 
+    private fun writeDto(call: HttpWsAction, lines: Lines): String {
+        val bodyParam = call.parameters.find { p -> p is BodyParam } as BodyParam?
+        if (bodyParam != null && bodyParam.isJson()) {
+
+            val primaryGene = bodyParam.primaryGene()
+            val choiceGene = primaryGene.getWrappedGene(ChoiceGene::class.java)
+            val actionName = call.getName()
+            if (choiceGene != null) {
+                throw IllegalStateException("Choice genes not yet supported for dto payload")
```
Pgarrett, line 127: Wanted to focus on a simple execution first, get an agreement on the code structure for it and will implement the `ChoiceGene` section

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
```diff
@@ -257,6 +268,51 @@ class ArrayGene<T>(
...
+        } else {
+            elements.forEach {
+                result.add("$listVarName.add(${it.getValueAsPrintableString(targetFormat = null)});")
+            }
+        }
+
+        return DtoCall(listVarName, result)
+    }
+
+    // this should actually be something we can ask the gene, like gene.getDtoType()
```
Pgarrett: What are your thoughts regarding this comment?

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -113,6 +117,43 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
+                            val template = leafGene.template
+                            if (template is ObjectGene) {
+                                 template.refType?:TestWriterUtils.safeVariableName(actionName)
+                            } else {
+                                throw IllegalStateException("Arrays of non custom objects are not collected as DTOs for root level. Attempted at $actionName")
+                            }
+                        }
+                        else -> throw IllegalStateException("Gene $leafGene is not supported for DTO payloads for action: $actionName")
+                    }
+                    val dtoCall = leafGene.getDtoCall(dtoName, 1)
```
Pgarrett: The `1` value here is the counter that will be appended to the variable name, so as to have a unique variable name. It will be useful whenever we have more than one request using the same object. Maybe the counter should be one level up? Or do you have any other idea to generate unique var names? Without using UUIDs

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
arcuri82: ok

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
arcuri82: putting inside Gene might be tricky, as it only apply to some genes, not all. also this is very specific  for DTO-based problems. i think "for now" this function could go in a utility class for DTO (and not inside `ArrayGene`). future might want to refactor it (eg inside `Gene`), but unsure now

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -54,6 +54,7 @@ object CookieWriter {
             }
 
             if (!format.isPython()) {
+                // TODO: should we use DTO for cookie related requests?
```
arcuri82, line 57: good question. i think for now "no". a comment like this is enough for now

core/src/main/kotlin/org/evomaster/core/output/dto/DtoCall.kt
```diff
@@ -0,0 +1,6 @@
+package org.evomaster.core.output.dto
+
+class DtoCall(
```
arcuri82, line 7: provide some JavaDoc for this class and its fields

core/src/main/kotlin/org/evomaster/core/output/service/GraphQLTestCaseWriter.kt
```diff
@@ -44,7 +44,7 @@ class GraphQLTestCaseWriter : HttpWsTestCaseWriter() {
         handleResponseAfterTheCall(call, result, responseVariableName, lines)
     }
 
-    override fun handleBody(call: HttpWsAction, lines: Lines) {
+    override fun handleBody(call: HttpWsAction, lines: Lines, dtoVar: String) {
```
arcuri82: what if there is no DTO involved? eg, when outputting for Python/JS. are you passing an empty`""` string? passing null would be better

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -113,6 +117,43 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
+                val leafGene = primaryGene.getLeafGene()
+                if (format.isJava()) {
+                    val dtoName = when (leafGene) {
+                        is ObjectGene -> leafGene.refType?:TestWriterUtils.safeVariableName(actionName)
+                        is ArrayGene<*> -> {
+                            val template = leafGene.template
+                            if (template is ObjectGene) {
+                                 template.refType?:TestWriterUtils.safeVariableName(actionName)
+                            } else {
+                                throw IllegalStateException("Arrays of non custom objects are not collected as DTOs for root level. Attempted at $actionName")
```
arcuri82: but this will happen. can have a TODO to remind we ll need to handle it at a certain point

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
arcuri82: you already have a `counter` for this:

```
abstract class TestCaseWriter {

    @Inject
    protected lateinit var config: EMConfig


    /**
     * In the tests, we might need to generate new variables.
     * We must guarantee that no 2 variables have the same name.
     * Easiest approach is to just use a counter that is incremented
     * at each new generated variable
     */
    protected var counter = 0
```

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -343,13 +384,19 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
             lines.add(getAcceptHeader(call, res))
         }
 
+        // TODO add support for kotlin
+        var dtoVar = ""
```
arcuri82: in this case, having null would be better.

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -496,7 +543,13 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
                 format.isPython() -> {
                     lines.add("body = ${bodyLines.first()}")
                 }
-                else -> lines.add(".$send(${bodyLines.first()})")
+                else -> {
+                    if (config.dtoForRequestPayload && format.isJava()) {
+                        lines.add(".$send(${dtoVar})")
```
arcuri82: leave a TODO that we might need to check type of individual as well, when we ll introduce Robustness Testing

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -521,12 +574,16 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
                 }
                 else -> {
-                    lines.add(".$send(${bodyLines.first()} + ")
-                    lines.indented {
-                        (1 until bodyLines.lastIndex).forEach { i ->
-                            lines.add("${bodyLines[i]} + ")
+                    if (config.dtoForRequestPayload && format.isJava()) {
+                        lines.add(".$send(${dtoVar})")
+                    } else {
+                        lines.add(".$send(${bodyLines.first()} + ")
```
arcuri82: looks like here there is some duplicated code from other branch? refactor/simplify

core/src/main/kotlin/org/evomaster/core/search/gene/Gene.kt
```diff
@@ -1240,5 +1234,9 @@ abstract class Gene(
             throw IllegalArgumentException("Not a child of current gene: $child - $this")
         }
     }
+
+    open fun getDtoCall(actionName: String, counter: Int): DtoCall {
```
arcuri82: add some JavaDoc

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
```diff
@@ -257,6 +268,51 @@ class ArrayGene<T>(
                 closingTag
     }
 
+    override fun getDtoCall(dtoName: String, counter: Int): DtoCall {
```
arcuri82: here we are dealing with code generation. you have declaration of varibles and statement ending with `;`. it should not be inside `Gene`. these functionalities should be refactored out somewhere under the `output` package

core/src/main/kotlin/org/evomaster/core/output/dto/DtoCall.kt
Pgarrett: Yeah, didn't add any JavaDoc since some things could change depending on the draft review

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
Pgarrett: What if, instead of DTO utility class, we create an interface like `DtoOutput` and only specific genes implement it? That would also be a more OOD

core/src/main/kotlin/org/evomaster/core/output/service/GraphQLTestCaseWriter.kt
Pgarrett: Ok!

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
Pgarrett: Yes, it will definitely happen. Had chosen to throw an exception since that would be the easiest way to catch the cases when testing

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
Pgarrett: Why type of individual?

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
Pgarrett: Oh yes, for sure. Draft PR mindset

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
Pgarrett: Yes, the actual code writing should be delegated to some object that has knowledge of the `OutputFormat`, most likely the already existing object `JavaDtoWriter`

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
Pgarrett: For Robustness Testing, wrong data will be sent thus not being able to instantiate a DTO

arcuri82 (approved):

PR #1326, opened 2025-09-20
Add precision

MohsenTaheriShalmani:
Since the main objective is to detect HTTP 400 responses, precision is a more meaningful metric than accuracy. Accuracy can be misleading if 400s are rare, while precision reflects how reliable the model’s 400 predictions are.
Additionally, it may be more sensible to use precision in model verification rather than expecting perfect performance.


core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelAccuracy.kt
```diff
@@ -1,18 +1,25 @@
...
 
     /**
      * Estimate the accuracy of the model
      */
     fun estimateAccuracy() : Double
 
+    /**
+     * Estimate the precision of the model
+     */
+    fun estimatePrecision400(): Double
```
arcuri82: if we provide both measures, we need to expand the comments to clarify what is the difference between accuracy and precision here

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelAccuracy.kt
```diff
@@ -1,18 +1,25 @@
...
     /**
      * A new prediction was made.
      * The estimate of the accuracy will be updated based on whether the prediction was correct or not.
      *
-     * Note: this can only be computed if, after the prediction, the input is evaluated (ie the test case is executed).
+     * Note: this can only be computed if, after the prediction, the input is evaluated (i.e., the test case is executed).
      * If a new input is created based on a rejected prediction, when we would not know if that was correct or not.
      */
-    fun updatePerformance(predictionWasCorrect: Boolean)
+    fun updatePerformance(predictedStatusCode: Int, actualStatusCode: Int? = null)
```
arcuri82: I assume here the semantic is `predictionWasCorrect = (predictedStatusCode == actualStatusCode)` ? if so, how can `actualStatusCode` be nullable?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -19,7 +22,29 @@ class ModelAccuracyWithTimeWindow(
...
 
-    override fun updatePerformance(predictionWasCorrect: Boolean) {
+    /**
+     * Precision for HTTP 400 responses: TP / (TP + FP).
+     */
+    override fun estimatePrecision400(): Double {
+        val tp = truePositive400Queue.count { it }   // count only true entries
+        val fp = falsePositive400Queue.count { it }  // count only true entries
+        return if ((tp + fp) > 0) tp.toDouble() / (tp + fp) else 0.0
+    }
```
arcuri82, line 46: we need some basic unit tests to validate these computations (eg in `ModelAccuracyWithTimeWindowTest.kt`)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -19,7 +22,29 @@ class ModelAccuracyWithTimeWindow(
...
+
+        val predictionWasCorrect = predictedStatusCode == actualStatusCode
+
         queue.add(predictionWasCorrect)
+
+        if (actualStatusCode == 400) {
+            truePositive400Queue.add(predictionWasCorrect)
+        } else {
+            falsePositive400Queue.add(!predictionWasCorrect)
+        }
```
arcuri82: is this correct? at each call of `updatePerformance`, should both data structures `truePositive400Queue` and `falsePositive400Queue` be updated?  regardless, add unit tests to validate the precision calculations

core/src/test/kotlin/org/evomaster/core/problem/rest/classifier/ModelAccuracyWithTimeWindowTest.kt
```diff
@@ -14,25 +14,25 @@ class ModelAccuracyWithTimeWindowTest {
 
         assertEquals(0.0, ma.estimateAccuracy(), delta)
 
-        ma.updatePerformance(true)
+        ma.updatePerformance(predictedStatusCode = 200, actualStatusCode = 200) //true
         assertEquals(1.0, ma.estimateAccuracy(), delta)
```
arcuri82, line 18: i guess here you could add assertions on the estimated precision

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
MohsenTaheriShalmani: Nice catch. I'll fix it.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400Classifier.kt
```diff
@@ -29,19 +30,40 @@ class Deterministic400Classifier(
...
 
-    override fun estimateAccuracy(endpoint: Endpoint): Double {
-        val m = models[endpoint] ?:
-            return 0.0
-
-        return m.estimateAccuracy(endpoint)
+    override fun estimateMetrics(endpoint: Endpoint): ModelEvaluation {
+        val m = models[endpoint] ?: return ModelEvaluation(
+            accuracy = 0.5,
+            precision400 = 0.5,
```
arcuri82: why `0.5` as default values and not `0.0`?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400Classifier.kt
```diff
@@ -29,19 +30,40 @@ class Deterministic400Classifier(
...
+            f1Score400 = 0.0,
+            mcc = 0.0
+        )
+        return m.estimateMetrics(endpoint)
     }
 
-    override fun estimateOverallAccuracy(): Double {
+    override fun estimateOverallMetrics(): ModelEvaluation {
+        if (models.isEmpty()) {
+            return ModelEvaluation(
```
arcuri82: duplicated code. should have a default immutable instance for example declared and returned directly from `ModelEvaluaton`, eg something like `ModelEvaluation.DEFAULT_NO_DATA`

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400Classifier.kt
```diff
@@ -29,19 +30,40 @@ class Deterministic400Classifier(
...
+                f1Score400 = 0.0,
+                mcc = 0.0
+            )
+        }
 
-        //average over all internal models
         val n = models.size.toDouble()
-        val sum = models.values.sumOf { it.estimateOverallAccuracy() }
+        val total = models.values.map {
+            it?.estimateOverallMetrics() ?: ModelEvaluation(0.5, 0.5, 0.0, 0.0, 0.0)
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400EndpointModel.kt
```diff
@@ -101,20 +102,36 @@ class Deterministic400EndpointModel(
...
+    override fun estimateMetrics(endpoint: Endpoint): ModelEvaluation {
         verifyEndpoint(endpoint)
+        if (!initialized) {
+            // hasn’t learned anything yet → return defaults
+            return ModelEvaluation(
+                accuracy = 0.5,
+                precision400 = 0.5,
+                recall400 = 0.0,
+                f1Score400 = 0.0,
+                mcc = 0.0
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400EndpointModel.kt
```diff
@@ -101,20 +102,36 @@ class Deterministic400EndpointModel(
...
+            )
+        }
 
-        return estimateOverallAccuracy()
-    }
+        val acc = modelMetrics.estimateAccuracy()
+        val prec = modelMetrics.estimatePrecision400()
+        val rec = modelMetrics.estimateRecall400()
+        val f1 = if (prec + rec == 0.0) 0.0 else 2 * (prec * rec) / (prec + rec)
```
arcuri82: is `f1` derived from `prec` and `rec`? if so, why having it as input, and not computed directly inside `ModelMetrics`?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400EndpointModel.kt
```diff
@@ -101,20 +102,36 @@ class Deterministic400EndpointModel(
...
+    }
 
-        if(!initialized){
-            //hasn't learned anything yet
-            return 0.0
+    /** Default overall metrics estimates */
+    override fun estimateOverallMetrics(): ModelEvaluation {
+        if (!initialized) {
+            // hasn’t learned anything yet
+            return ModelEvaluation(0.5, 0.5, 0.0, 0.0, 0.0)
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400Classifier.kt
```diff
@@ -62,16 +63,40 @@ abstract class AbstractProbabilistic400Classifier<T : AIModel>(
...
-    override fun estimateAccuracy(endpoint: Endpoint): Double {
-        val m = models[endpoint] ?: return 0.5
-        return m.estimateAccuracy(endpoint)
+    override fun estimateMetrics(endpoint: Endpoint): ModelEvaluation {
+        val m = models[endpoint] ?: return ModelEvaluation(
+            accuracy = 0.5,
+            precision400 = 0.5,
+            recall400 = 0.0,
+            f1Score400 = 0.0,
+            mcc = 0.0
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400Classifier.kt
```diff
@@ -62,16 +63,40 @@ abstract class AbstractProbabilistic400Classifier<T : AIModel>(
...
-    override fun estimateOverallAccuracy(): Double {
-        if (models.isEmpty()) return 0.5
+    override fun estimateOverallMetrics(): ModelEvaluation {
+        if (models.isEmpty()) {
+            return ModelEvaluation(
+                accuracy = 0.5,
+                precision400 = 0.5,
+                recall400 = 0.0,
+                f1Score400 = 0.0,
+                mcc = 0.0
```
arcuri82: duplicated code. if we want to change the default of 1 value, then we would have to hunt down all these places to update.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400Classifier.kt
```diff
@@ -62,16 +63,40 @@ abstract class AbstractProbabilistic400Classifier<T : AIModel>(
...
+            )
+        }
+
         val n = models.size.toDouble()
-        val sum = models.values.sumOf { it?.estimateOverallAccuracy() ?: 0.5 }
-        return sum / n
+        val total = models.values.map {
+            it?.estimateOverallMetrics() ?: ModelEvaluation(0.5, 0.5, 0.0, 0.0, 0.0)
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -54,33 +56,51 @@ abstract class AbstractProbabilistic400EndpointModel(
...
         verifyEndpoint(endpoint)
-        return estimateOverallAccuracy()
+        if (!initialized) {
+            // hasn’t learned anything yet → return defaults
+            return ModelEvaluation(
+                accuracy = 0.5,
+                precision400 = 0.5,
+                recall400 = 0.0,
+                f1Score400 = 0.0,
+                mcc = 0.0
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -54,33 +56,51 @@ abstract class AbstractProbabilistic400EndpointModel(
...
+            )
+        }
+
+        val acc = modelMetrics.estimateAccuracy()
+        val prec = modelMetrics.estimatePrecision400()
+        val rec = modelMetrics.estimateRecall400()
+        val f1 = if (prec + rec == 0.0) 0.0 else 2 * (prec * rec) / (prec + rec)
```
arcuri82: duplicated code. i see same computation in a previous class. might consider move it inside ModelEvaluation

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -54,33 +56,51 @@ abstract class AbstractProbabilistic400EndpointModel(
...
+        return ModelEvaluation(acc, prec, rec, f1, mcc)
     }
 
-    override fun estimateOverallAccuracy(): Double {
+    /** Default overall metrics estimates */
+    override fun estimateOverallMetrics(): ModelEvaluation {
         if (!initialized) {
             // hasn’t learned anything yet
-            return 0.5
+            return ModelEvaluation(0.5, 0.5, 0.0, 0.0, 0.0)
```
arcuri82: duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelEvaluation.kt
```diff
@@ -0,0 +1,10 @@
+package org.evomaster.core.problem.rest.classifier
+
+data class ModelEvaluation(
+    val accuracy: Double,
```
arcuri82, line 19: add javadoc for all these fields

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelEvaluation.kt
```diff
@@ -0,0 +1,10 @@
...
+    val precision400: Double,
+    val recall400: Double,
+    val f1Score400: Double,
```
arcuri82: is this derived from the other fields? if so, can compute it in an `init{}` block instead of passing it as input

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelEvaluation.kt
```diff
@@ -0,0 +1,10 @@
...
+    val mcc: Double
+)
```
arcuri82: should have an `init` block to validate the inputs, ie, each should be in [0,1], or throw IllegalArgumentException otherwise

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelEvaluation.kt
arcuri82: can be brief comment, with a `[ModelMetrics]` link for more details

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetrics.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Estimate the overall accuracy of the model.
+     *
+     * Accuracy = (TP + TN) / (TP + TN + FP + FN)
+     *
+     * The proportion of correct predictions
+     * (both positives and negatives) out of all evaluated predictions.
+     */
+    fun estimateAccuracy(): Double
```
arcuri82, line 31: isn't this just `return estimateMetrics().accuracy`? if so, could provide it here as `default` implementation for this interface. or am i missing something? also, this comment applies to all other functions here in this interface

arcuri82: or, alternatively, it should be `fun estimateMetrics()` that should have a default implementation calling the other methods (this might be a better option)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsFullHistory.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Recall(400) = TP / (TP + FN)
+     *
+     * Of all requests that were actually 400, how many were correctly predicted as 400.
+     */
+    override fun estimateRecall400(): Double {
+        val denominator = truePositive400 + falseNegative400
+        return if (denominator > 0) truePositive400.toDouble() / denominator else 0.0
```
arcuri82, line 52: why in a method it is called `denom` but here `denominator`? be consistent :)  as rule of thumb, I prefer full names, eg `denominator`, unless the code get too cluttered

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsFullHistory.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * F1(400) = 2 * (Precision * Recall) / (Precision + Recall)
+     *
+     * Harmonic mean of precision and recall.
+     */
+    override fun estimateF1Score400(): Double {
+        val p = estimatePrecision400()
+        val r = estimateRecall400()
+        return if ((p + r) > 0.0) 2 * (p * r) / (p + r) else 0.0
```
arcuri82: isn't this computation of F1 repeated in few places? should be written only once, and reused as utility function

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsFullHistory.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * MCC(400) = (TP*TN - FP*FN) /
+     *            sqrt((TP+FP)(TP+FN)(TN+FP)(TN+FN))
+     *
+     * Matthews Correlation Coefficient, robust even with imbalanced data.
+     */
```
arcuri82: when you have this kind of formula, can be nice here in the code to leave a comment with URL to Wikipedia, or other sources that are unlikely to change (ie broken links)

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -0,0 +1,110 @@
...
+        return if ((tp + fn) > 0) tp.toDouble() / (tp + fn) else 0.0
+    }
+
+    /**
+     * F1(400) = 2 * (Precision * Recall) / (Precision + Recall)
+     */
+    override fun estimateF1Score400(): Double {
+        val p = estimatePrecision400()
+        val r = estimateRecall400()
+        return if (p + r > 0.0) 2 * (p * r) / (p + r) else 0.0
```
arcuri82: duplicated equation

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        val denominator = Math.sqrt((tp + fp) * (tp + fn) * (tn + fp) * (tn + fn))
+        return if (denominator > 0) (tp * tn - fp * fn) / denominator else 0.0
+    }
+
+    /**
+     * Unified metrics estimate packaged into a [ModelEvaluation].
+     */
+    override fun estimateMetrics(): ModelEvaluation {
+        return ModelEvaluation(
```
arcuri82: this could be the default implementation directly in the interface `ModelEvaluation`.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -0,0 +1,110 @@
...
+    override fun updatePerformance(predictedStatusCode: Int, actualStatusCode: Int) {
+        val predictionWasCorrect = predictedStatusCode == actualStatusCode
+        queue.add(predictionWasCorrect)
+
+        when {
+            actualStatusCode == 400 && predictedStatusCode == 400 -> truePositive400Queue.add(true)
+            actualStatusCode == 400 && predictedStatusCode != 400 -> falseNegative400Queue.add(true)
+            actualStatusCode != 400 && predictedStatusCode == 400 -> falsePositive400Queue.add(true)
+            actualStatusCode != 400 && predictedStatusCode != 400 -> trueNegative400Queue.add(true)
+        }
```
arcuri82: i m not sure these computations are correct, if we want to have a timewindow of size N. 
regardless of the number of queues, only half of them get new data at each call of this function. and only with value `true`. 
i think all queues should be updated (eg, consider the `false` cases as well)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -56,8 +57,8 @@ class AIResponseClassifier : AIModel {
             else -> object : AIModel {
                 override fun updateModel(input: RestCallAction, output: RestCallResult) {}
                 override fun classify(input: RestCallAction) = AIResponseClassification()
-                override fun estimateAccuracy(endpoint: Endpoint): Double  = 0.0
-                override fun estimateOverallAccuracy(): Double = 0.0
+                override fun estimateMetrics(endpoint: Endpoint): ModelEvaluation  = ModelEvaluation(accuracy = 0.5,precision400 = 0.5,recall400 = 0.0,f1Score400 = 0.0,mcc = 0.0)
+                override fun estimateOverallMetrics(): ModelEvaluation  = ModelEvaluation(accuracy = 0.5,precision400 = 0.5,recall400 = 0.0,f1Score400 = 0.0,mcc = 0.0)
```
arcuri82: duplicated code. this default instance should really be instantiated only once and referred in all thes different places. so that, if we want to change the default of one parameter, we do it only once, and not 20 times :)

core/src/test/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindowTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.api.Test
+
+class ModelMetricsWithTimeWindowTest {
+
+
+    @Test
+    fun estimateModelMetrics() {
+
+        val delta = 0.0001
+        val ma = ModelMetricsWithTimeWindow(10) // buffer large enough to hold all
```
arcuri82, line 13: besides this test, add back previous deleted test, where `val ma = ModelAccuracyWithTimeWindow(4)`.
that was done on purpose to verify the time window.
as rule of thumb, do not delete or modify existing tests, add new ones (of course if needed, changes/deletes can be discussed).
as wrote in a previous comment, I do not think the timewindow implementation is correct. to verify it, need a buffer that is smaller than the data you insert

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400EndpointModel.kt
```diff
@@ -101,20 +102,29 @@ class Deterministic400EndpointModel(
...
+    /** Default metrics estimates */
+    override fun estimateMetrics(endpoint: Endpoint): ModelEvaluation {
         verifyEndpoint(endpoint)
+        if (!initialized) {
+            // hasn’t learned anything yet → return defaults
+            return ModelEvaluation.DEFAULT_NO_DATA
+        }
 
-        return estimateOverallAccuracy()
+        return ModelEvaluation(
```
arcuri82: i don't understand this change. why removing the call to `estimateOverallAccuracy()`?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -54,33 +56,44 @@ abstract class AbstractProbabilistic400EndpointModel(
...
+            // hasn’t learned anything yet → return defaults
+            return ModelEvaluation.DEFAULT_NO_DATA
+        }
+
+        return ModelEvaluation(
+            modelMetrics.estimateAccuracy(),
+            modelMetrics.estimatePrecision400(),
+            modelMetrics.estimateRecall400(),
+            modelMetrics.estimateMCC400()
+        )
```
arcuri82: same as in other case. a model for a specific endpoint will have no difference between `estimateMetrics(endpoint: Endpoint)` and `estimateOverallMetrics()` apart from the endpoint validation in the former. so one function should call the other to avoid code duplication. also, what is the difference between:
```
return ModelEvaluation(
            modelMetrics.estimateAccuracy(),
            modelMetrics.estimatePrecision400(),
            modelMetrics.estimateRecall400(),
            modelMetrics.estimateMCC400()
        )
```
and
```
return modelMetrics.estimateMetrics()
```
?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsFullHistory.kt
```diff
@@ -0,0 +1,97 @@
...
+        val fp = falsePositive400.toDouble()
+        val fn = falseNegative400.toDouble()
+
+        val denominator = sqrt((tp + fp) * (tp + fn) * (tn + fp) * (tn + fn))
+        return if (denominator > 0) (tp * tn - fp * fn) / denominator else 0.0
+    }
+
+    /** Unified metrics estimate packaged into a [ModelEvaluation] */
+    override fun estimateMetrics(): ModelEvaluation {
+        return ModelEvaluation(
```
arcuri82: why does this need to be overridden if the implementation is the same?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -0,0 +1,97 @@
...
+
+        val denominator = Math.sqrt((tp + fp) * (tp + fn) * (tn + fp) * (tn + fn))
+        return if (denominator > 0) (tp * tn - fp * fn) / denominator else 0.0
+    }
+
+    /** Unified metrics estimate packaged into a [ModelEvaluation] */
+    override fun estimateMetrics(): ModelEvaluation {
+        return ModelEvaluation(
+            accuracy = estimateAccuracy(),
+            precision400 = estimatePrecision400(),
```
arcuri82: see previous comment. if the implementation is the same, then no need to override

core/src/test/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindowTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        // 1) correct prediction: 200 vs. 200 -> TN
+        ma.updatePerformance(predictedStatusCode = 200, actualStatusCode = 200)
```
arcuri82, line 16: keep this test. but add a new one (can copy&paste some parts of this test)in which the amount of data is more than the buffer. eg you could have a buffer of size 2, and insert the first 5 elements in this test

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/deterministic/Deterministic400EndpointModel.kt
MohsenTaheriShalmani: I think for an endpoint model, “overall” is the same as the endpoint’s metrics, so I just return its own metrics directly.

arcuri82 (approved):

PR #1327, opened 2025-09-22
Support for non custom data types when using DTOs

Pgarrett:
When a payload is supposed to be a JSON primitive value such as a string, number or boolean, DTO setup will not create a DTO for this cases and instead use the JSON stringified representation of the value.

### DIFF
--- a/core-it/src/main/kotlin/bar/examples/it/spring/callgraphdelete/CallGraphDeleteApplication.kt
+++ b/core-it/src/main/kotlin/bar/examples/it/spring/callgraphdelete/CallGraphDeleteApplication.kt
@@ -1,0 +1,25 @@
+package bar.examples.it.spring.callgraphdelete
+
+import bar.examples.it.spring.body.BodyDto
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+
+@SpringBootApplication(exclude = [SecurityAutoConfiguration::class])
+@RequestMapping(path = ["/api/body"])
+@RestController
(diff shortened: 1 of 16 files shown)


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -69,34 +69,49 @@ class DtoWriter {
...
-        // merge options into a single DTO
-        children.forEach { childGene ->
-            when (childGene) {
-                is ObjectGene -> populateDtoClass(dtoClass, childGene)
-                is ArrayGene<*> -> {
-                    val template = childGene.template
-                    if (template is ObjectGene) {
-                        populateDtoClass(dtoClass, template)
+        // TODO: should we handle EnumGene?
+        if (hasObjectOrArrayGene(gene)) {
```
Pgarrett, line 73: @arcuri82, added this check instead of checking for the first gene inside a `ChoiceGene` since:
1. When testing, `EnumGene` came as the first one (and also the active one)
2. Since we'll be merging `ObjectGene`s and `ArrayGene`s (lines below) it seemed the most reasonable check

What do you think?

arcuri82: ok. seem fine at first look. if it will give issues, we will find out when tyring out on WFD :)

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -124,25 +126,33 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
             val actionName = call.getName()
             if (choiceGene != null) {
                 // TODO add support for payloads from choice genes
-                throw IllegalStateException("Choice genes not yet supported for dto payload")
+                if (hasObjectOrArrayGene(choiceGene)) {
+                    // this because when using `format`, integer and number genes are being represented as ChoiceGene
```
arcuri82: i don't think it is due to `format`, but rather due to `example` and `default` entries

Pgarrett: 🤦 this is an old comment, from before we had chatted. Sorry!

PR #1330, opened 2025-09-23
Custom callback URL to work with Docker

seran:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2537,6 +2537,10 @@ class EMConfig {
     @Cfg("Strategy to classify inputs for potential vulnerability classes related to an REST endpoint.")
     var vulnerableInputClassificationStrategy = VulnerableInputClassificationStrategy.MANUAL
 
+    @Experimental
+    @Cfg("HTTP callback URL hostname")
```
arcuri82: should add more info in this description. for example, explicitely state this is needed for when SUT and tool runs on different machines, or when SUT is inside Docker

arcuri82: also explicitely say this is just the hostname and not a full URL

PR #1331, opened 2025-09-24
Path variable support for SSRF

seran:


core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -70,7 +70,7 @@ class SSRFAnalyser {
      * Possible URL variable names.
      * TODO: Can load from a file.
      */
-    private val potentialUrlParamNames: List<String> = listOf("url", "source", "target", "datasource", "referer")
+    private val potentialUrlParamNames: List<String> = listOf("url", "source", "target", "datasource", "referer", "imageURL", "")
```
arcuri82: why this? wouldn't this entry already be handled by the regex?

PR #1332, opened 2025-09-29
Encoder improvement

MohsenTaheriShalmani:


core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -29,7 +29,7 @@ abstract class AbstractProbabilistic400EndpointModel(
     protected var initialized: Boolean = false
 
     val modelMetricsFullHistory: ModelMetricsFullHistory = ModelMetricsFullHistory()
-    val modelMetrics: ModelMetrics = ModelMetricsWithTimeWindow(20)
+    val modelMetricsWithTimeWindow: ModelMetrics = ModelMetricsWithTimeWindow(1000)
```
arcuri82: is this wise? as this is per endpoint, and there might be 100s in an API, this might endup being the whole history. maybe something like 100 could be more reasonable, unless there was some specific reason for such a value. note that, in typical experiments with 1 hour budget, total number of calls is typically in the range 100k to 300k

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/knn/KNN400EndpointModel.kt
```diff
@@ -28,6 +28,7 @@ class KNN400EndpointModel (
 
     companion object {
         private const val NOT_400 = 200
+        private const val MAX_SAMPLES = 10_000    // Fixed-size buffer for samples to avoid memory overload
```
arcuri82: why in KNN this is a constant, but a variable (having same 10_000 value) in KDE?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/knn/KNN400EndpointModel.kt
MohsenTaheriShalmani: Basically, it must be a constant. I'll fix it.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/knn/KNN400EndpointModel.kt
```diff
@@ -39,6 +34,11 @@ class KNN400EndpointModel (
      */
     val samples = mutableListOf<Pair<List<Double>, Int>>()
 
+    /** Cap on stored samples based on the reservoir-style uniform downsampling to avoid memory overload.*/
+    companion object {
+        private const val MAX_SAMPLES = 20_000
```
arcuri82: why some have `10_000` and others `20_000`? is bit confusing. also it might open to critiscims from reviewers, unless we can document good reasons for this (and, in such case, those good reeasons should be briefly mentioned here)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -105,10 +105,14 @@ class AIResponseClassifier : AIModel {
     fun attemptRepair(call: RestCallAction){
 
         val metrics = estimateMetrics(call.endpoint)
-        val accuracy = metrics.accuracy
 
-        //TODO any better way to use this accuracy?
-        if(!randomness.nextBoolean(accuracy)){
+        /**
+         * Skips repair if the classifier is still weak, as indicated by low accuracy and F1-score
```
arcuri82, line 134: `and` or `or`?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/knn/KNN400EndpointModel.kt
MohsenTaheriShalmani: Agree, I'll add it to the Config with a clear explanation

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
MohsenTaheriShalmani: I think AND, because the classifier needs to perform well in both accuracy and precision–recall, otherwise, its knowledge of true positives is no better than a random guess.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/AbstractProbabilistic400EndpointModel.kt
```diff
@@ -23,13 +23,21 @@ abstract class AbstractProbabilistic400EndpointModel(
...
-    val modelMetrics: ModelMetrics = ModelMetricsWithTimeWindow(20)
+    companion object {
+        const val NOT_400 = 200
+    }
+
+    /** Performance metrics tracker.*/
+    val modelMetrics: ModelMetrics = when (metricType) {
+        EMConfig.AIClassificationMetrics.TIME_WINDOW -> ModelMetricsWithTimeWindow(100)
+        EMConfig.AIClassificationMetrics.FULL_HISTORY -> ModelMetricsFullHistory()
+    }
```
arcuri82: this looks like duplicated code

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/quantifier/ModelMetricsWithTimeWindow.kt
```diff
@@ -19,70 +19,39 @@ import kotlin.math.sqrt
  * short-term view of performance rather than lifetime history.
  */
 class ModelMetricsWithTimeWindow(
-    bufferSize: Int
+    bufferSize: Int = 100
 ) : ModelMetrics {
 
-    private val queue: EvictingQueue<Boolean> = EvictingQueue.create(bufferSize)
+    /** Lifetime counter of total evaluated requests. Important for warm-up logic. */
+    override var totalSentRequests: Int = 0
```
arcuri82, line 26: why does this one have a public setter? 
also, instead of override, would't make sense to rather have it protected in superclass and access it directly from there?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -117,13 +147,28 @@ class AIResponseClassifier : AIModel {
 
         repeat(n) {
             val classification = classify(call)
-
             val p = classification.probabilityOf400()
 
+            // Stop attempts to repair if the classifier predicts a non-400 response
+            if (classification.prediction()!=400) return
```
arcuri82: why this? isn't this redundant / obsolete the following checks?

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/quantifier/ModelMetricsWithTimeWindow.kt
MohsenTaheriShalmani: I think we can’t make it `protected` since `ModelMetrics` is an interface. I’ll fix it by using a `private set` for the setter.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
MohsenTaheriShalmani: I added that check as an early exit to skip all repair attempts if the classifier predicts a non-400 response. Otherwise, we’d continue trying to repair even when we already have a non-400 in hand.

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/kde/KDE400EndpointModel.kt
```diff
@@ -174,18 +172,19 @@ class KDE400EndpointModel (
...
-                if (samples.size < maxStored) {
-                    samples.add(x)
-                } else {
-                    // reservoir: replace with decreasing probability
-                    val r = kotlin.random.Random.nextLong(seen)
-                    if (r < maxStored) {
-                        samples[r.toInt()] = x
-                    }
+                // reservoir: replace with decreasing probability
+                val r = kotlin.random.Random.nextLong(seen)
```
arcuri82: cannot access `kotlin.random.Random` directly. 
all sources of randomness must come from the same singleton `Randomness`

core/src/main/kotlin/org/evomaster/core/problem/rest/classifier/probabilistic/knn/KNN400EndpointModel.kt
```diff
@@ -112,12 +114,26 @@ class KNN400EndpointModel (
...
+         * - If we have not yet filled the reservoir (samples.size < maxStoredSamples), add the new sample.
+         * - Otherwise, replace an existing sample with decreasing probability to maintain
+         *   a uniform random subset of all seen data.
+         */
+        seen++
+        if (samples.size < maxStoredSamples) {
+            samples.add(inputVector to label)
         } else {
-            samples.add(inputVector to NOT_400)
+            val r = kotlin.random.Random.nextLong(seen)
```
arcuri82: see previous comment on `Randomness`.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
arcuri82: i m not sure I agree with this. we can discuss it on slack

arcuri82 (changes requested):

PR #1333, opened 2025-09-30
Support to handle StringGene inside ChoiceGene in GeneUtils.getAllFields()

seran:


core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
```diff
@@ -626,7 +626,7 @@ object RestActionBuilderV3 {
                 a path is inside a Disruptive Gene, because there are cases in which we want to prevent
                 mutation. Note that 1.0 means can always be mutated
              */
-            "path" -> params.add(PathParam(name, CustomMutationRateGene("d_", gene, 1.0))
+            "path" -> params.add(PathParam(name, CustomMutationRateGene(gene.name, gene, 1.0))
```
arcuri82: shouldn't this be just `name` instead of `gene.name`? or are those different?

core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
```diff
@@ -938,11 +941,20 @@ object GeneUtils {
 
         val leaf = gene.getLeafGene()
 
-        if(klass.isAssignableFrom(leaf.javaClass)){
-            //we are adding the wrapper gene, not the leaf
-            fields.add(gene)
+        val parent = leaf.parent
+
+        if (parent is ChoiceGene<*>) {
+            if (parent.getViewOfChildren().any {it is StringGene}){
```
arcuri82: shouldn't this check for `klass.isAssignableFrom`?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
seran: `name` can be used here.

arcuri82 (approved):

PR #1334, opened 2025-09-30
Fixing TestWriter to avoid duplicate HttpCallbackVerifier

seran:


core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -70,7 +70,7 @@ class SSRFAnalyser {
      * Possible URL variable names.
      * TODO: Can load from a file.
      */
-    private val potentialUrlParamNames: List<String> = listOf("url", "source", "target", "datasource", "referer")
+    private val potentialUrlParamNames: List<String> = listOf("url", "source", "target", "datasource", "referer", "imageURL", "")
```
arcuri82: see comment in other PR

core/src/main/kotlin/org/evomaster/core/problem/security/service/SSRFAnalyser.kt
```diff
@@ -337,9 +335,10 @@ class SSRFAnalyser {
     }
 
     private fun getIndividualsWithStatus2XX(): List<EvaluatedIndividual<RestIndividual>> {
+        // TODO: Status codes changed temporarily for the experimental purposes
         return RestIndividualSelectorUtils.findIndividuals(
             this.archive.extractSolution().individuals,
-            statusGroup = StatusGroup.G_2xx
+            statusCodes = listOf(200, 201, 203, 204, 401, 403, 422, 419)
```
arcuri82: these codes do not match what the name of this function is. 
also, it is very unclear why these 4xx codes are needed.
need a justification. 
we cannot have solutions that are ad-hoc just for a specific case study, unless we can generalize some rules

arcuri82 (approved):

PR #1336, opened 2025-10-03
Stack Trace Oracle

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -262,30 +265,96 @@ class SecurityRest {
...
+                val copyTarget = target.copy()
+
+                copyTarget.evaluatedMainActions().forEach { action ->
+                    // body
+                    val result = (action.result as RestCallResult)
+                    val action = (action.action as RestCallAction)
+
+                    val body = result.getBody()
+
+                    if(body != null && StackTraceUtils.looksLikeStackTrace(body)){
```
arcuri82: somehow, this check should be done in the fitness function as well. otherwise, if a test is cloned and its fitness recomputed, we lose the info. 
in the fitness function, it could be checked only for 500 and for SECURITY type tests

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -262,30 +265,96 @@ class SecurityRest {
...
+
+            for(target in suspicious) {
+                var isFaultFound = false
+
+                val copyTarget = target.copy()
+
+                copyTarget.evaluatedMainActions().forEach { action ->
+                    // body
+                    val result = (action.result as RestCallResult)
+                    val action = (action.action as RestCallAction)
```
arcuri82: doesn't `action` here overshadow the `action` in line `310`? 
also, shouldn't we just look at the ones with same verb and path from 310 instead of all actions in a test?

core/src/main/kotlin/org/evomaster/core/utils/StackTraceUtils.kt
```diff
@@ -0,0 +1,104 @@
...
+         * Determines whether the given text looks like a stack trace.
+         *
+         * Heuristics:
+         * - If there is an error/exception keyword AND at least 2 matching lines → true
+         * - If there is no keyword but at least 3 matching lines → true
+         *
+         * @param text the log or string to analyze
+         * @return true if the text appears to be a stack trace
+         */
+        fun looksLikeStackTrace(text: String): Boolean {
```
arcuri82: we need several unit test for this function, with example of real-world response strings (taking from googling) from different languages (at least the ones we use here) and major frameworks.
we need both "positive" and "negative" examples

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1291,6 +1293,26 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
         }
     }
 
+
+    private fun handleStackTraceCheck(
+        individual: RestIndividual,
+        actionResults: List<ActionResult>,
+        fv: FitnessValue
+    ) {
+        for(index in individual.seeMainExecutableActions().indices){
```
arcuri82, line 1327: shouldn't be a check for `config.getDisabledOracleCodesList()`?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -262,30 +265,94 @@ class SecurityRest {
...
+     *
+     * Note: This is a best-effort oracle that may produce false positives, as some applications might
+     * legitimately return stack traces as part of their business logic.
+     *
+     * This check is performed only at the end of the search, not during each fitness evaluation,
+     * to avoid performance overhead during the main search phase.
+     */
+    private fun handleStackTraceCheck(){
+
+        mainloop@ for(action in actionDefinitions){
```
arcuri82, line 310: check for `config.getDisabledOracleCodesList()` ?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -262,30 +265,94 @@ class SecurityRest {
...
+
+            val suspicious = RestIndividualSelectorUtils.findIndividuals(
```
arcuri82, line 312: we need a bit of code comments to explain why we need BOTH the check here and in the fitness function. 
1) here: because we need to make sure 500 from archive are re-executed and added back with SECURITY tag if fault found
2) in ff: to avoid losing info if test is re-evaluated

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
omursahin: I am checking it in accessControlBasedOnRESTGuidelines.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
omursahin: I added this check for all oracle types in the AbstractRestFitness.

PR #1342, opened 2025-10-08
trying automatically using different options in different E2E, for op…

arcuri82:
…tions that should not impact the results of the tests


arcuri82:
@Pgarrett @jgaleotti  FYI

PR #1343, opened 2025-10-09
Bug fixing in DTOs

Pgarrett:
Issues fixed in this PR:
1. No body param still tries to generate dto.
2. Duplicate variable names in the same test case. This was happening with lists, the approach taken was not to name the list variable with the type but with the target attribute. Thus instead of being `list_String_1` it will be named `list_Keys_1` if the target attribute is `keys`.
3. DTO instantiated with the key name: `Key dto_Key_3 = new Key()` when it should have been: `ContributorKey dto_Key_3 = new ContributorKey()`.
4. For `Long` values, add `L` as last char in the number and for `Float` add `f`.
5. Add imports for lists in DTO classes.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/dto/GeneToDto.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/dto/GeneToDto.kt
@@ -154,6 +154,8 @@ class GeneToDto(
         }
     }
 
+    // According to documentation, a trailing constant is only needed for Long, Hexadecimal and Float
+    // https://kotlinlang.org/docs/numbers.html#literal-constants-for-numbers
     private fun getValueSuffix(gene: Gene): String {
         return when (gene) {
             is LongGene -> "L"


core/src/main/kotlin/org/evomaster/core/output/dto/GeneToDto.kt
```diff
@@ -152,4 +153,12 @@ class GeneToDto(
             else -> throw Exception("Not supported gene at the moment: ${gene?.javaClass?.simpleName} for field $fieldName")
         }
     }
+
+    private fun getValueSuffix(gene: Gene): String {
+        return when (gene) {
+            is LongGene -> "L"
+            is FloatGene -> "f"
```
arcuri82, line 162: what about `DoubleGene`?

Pgarrett: Not needed actually! https://kotlinlang.org/docs/numbers.html#literal-constants-for-numbers

arcuri82: can you add such info (including link) as a comment?

Pgarrett: Done!

PR #1346, opened 2025-10-11
use experimental oracles

omursahin:


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2856,23 +2861,47 @@ class EMConfig {
 
     fun getDisabledOracleCodesList(): List<FaultCategory> {
         if (disabledOracleCodesList == null) {
+            val definedCategories = DefinedFaultCategory.values().asList()
```
arcuri82: this is a long `if` statement, where the `else` is just a `return`. to avoid unnecessary nesting (which reduce code readability), the condition should be negated (eg `if(disabledOracleCodeList != null)`), and return immediately if so

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2856,23 +2861,47 @@ class EMConfig {
...
-                                    allCategories.joinToString("\n") { "${it.code} (${it.name})" }
-                        )
+                    category
                 }
+
+            if (!useExperimentalOracles) {
+                disabledOracleCodesList = (disabledOracleCodesList!! + experimentalCategories).distinct()
+            }
         }
         return disabledOracleCodesList!!
```
arcuri82, line 2948: so, if user specifies some codes to skip, then the content of `useExperimentalOracles` is ignored? i don't think that would be correct. I think here a better approach is to change this function, to do not return a list of codes, but a boolean, something like `isFaultCodeActive(fault): Boolean`, so the logic has to be implemted only once

arcuri82 (changes requested):

PR #1347, opened 2025-10-15
Changes in StandardGeneticAlgorithm: Freeze targets per generation; add GASolutionSource

francastagna:
## PR Summary

We implemented changes to the `StandardGeneticAlgorithm`:

- **With and without archive**: allows selecting the source of the final solution.  
- **Fitness update**: introduces generation-frozen objectives (“frozen targets”).

We implemented changes to the `AbstractGeneticAlgorithm`:
- **GA Recorder/Observer**: adds observer hooks to record GA events (selection, crossover and mutation events).

---

## Main Changes

- **Configuration**: added `EMConfig.GASolutionSource` and `EMConfig.gaSolutionSource` to choose the source of the final solution: `ARCHIVE` or `POPULATION`.  
- **Base GA**: rewrote the core logic in `AbstractGeneticAlgorithm` (population, elitism, tournament, `mutate`, `xover`, `sampleSuite`).  
- **`score` function**: computes fitness using generation-frozen objectives (`frozenTargets`).  
- **Dual support**: different behavior when the final solution is built from the archive vs. from the population.  
- **`StandardGeneticAlgorithm.searchOnce()`**:
  - Freezes objectives at the start of each generation: `frozenTargets = archive.notCoveredTargets()`.  
  - Builds the next population via tournament selection, crossover, and mutation, respecting the configured probabilities.
- **GA Recorder/Observer**:
  - Introduces observer hooks in the GA core to record operator metrics.
  - Enhances debugging and testing capabilities without altering the GA behaviour.

---

### Why

We align with Campos’ paper recommendation to keep objectives fixed within a generation so that all individuals compete under the same set of pending goals.  
This prevents fitness drift when the objective set changes mid-generation (due to new coverages), which would make comparisons between individuals inconsistent.

### How

- At the start of each generation, capture `frozenTargets = archive.notCoveredTargets()`.  
- The `score` of a test suite is computed only over those frozen objectives throughout the generation.  
- At the end of the generation (or at a snapshot), the set can be updated.

### Advantages

- Intra-generation consistency for tournament/selection.  
- Fair and stable comparisons; avoids individual evaluation depending on the exact timing of its assessment.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/BaseModule.kt
+++ b/core/src/main/kotlin/org/evomaster/core/BaseModule.kt
@@ -15,12 +15,6 @@ import org.evomaster.core.search.service.mutator.genemutation.ArchiveGeneMutator
 import org.evomaster.core.search.tracer.ArchiveMutationTrackService
 import org.evomaster.core.search.tracer.TrackService
 import org.evomaster.core.solver.SMTLibZ3DbConstraintSolver
-import org.evomaster.core.search.algorithms.strategy.SelectionStrategy
-import org.evomaster.core.search.algorithms.strategy.TournamentSelectionStrategy
-import org.evomaster.core.search.algorithms.strategy.CrossoverOperator
-import org.evomaster.core.search.algorithms.strategy.DefaultCrossoverOperator
-import org.evomaster.core.search.algorithms.strategy.MutationOperator
-import org.evomaster.core.search.algorithms.strategy.DefaultMutationOperator
 
 
 /**
@@ -93,19 +87,6 @@ class BaseModule(val args: Array<String>, val noTests: Boolean = false) : Abstra
                 .asEagerSingleton()
 
         bind(SMTLibZ3DbConstraintSolver::class.java)
-            .asEagerSingleton()
-
-        // Genetic algorithm strategies (selection & crossover)
-        bind(SelectionStrategy::class.java)
-            .to(TournamentSelectionStrategy::class.java)
-            .asEagerSingleton()
-
-        bind(CrossoverOperator::class.java)
-            .to(DefaultCrossoverOperator::class.java)
-            .asEagerSingleton()
-
-        bind(MutationOperator::class.java)
-            .to(DefaultMutationOperator::class.java)
             .asEagerSingleton()
 
         //no longer needed if TestSuiteWriter is moved out?
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -2853,9 +2853,9 @@ class EMConfig {
     fun isEnabledAIModelForResponseClassification() = aiModelForResponseClassification != AIResponseClassifierModel.NONE
 
     /**
-     * Source to build the final GA solution.
+     * Source to build the final GA solution when evolving full test suites (not single tests).
      * ARCHIVE: use current behavior (take tests from the archive).
-     * POPULATION: for GA algorithms, take the best individual from the final population.
+     * POPULATION: for GA algorithms, take the best suite (individual) from the final population.
      */
     enum class GASolutionSource { ARCHIVE, POPULATION }
 
--- a/core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
@@ -6,9 +6,12 @@ import org.evomaster.core.search.service.SearchAlgorithm
 import org.evomaster.core.EMConfig
 import org.evomaster.core.search.Solution
 import com.google.inject.Inject
-import org.evomaster.core.search.algorithms.strategy.CrossoverOperator
-import org.evomaster.core.search.algorithms.strategy.MutationOperator
-import org.evomaster.core.search.algorithms.strategy.SelectionStrategy
+import org.evomaster.core.search.algorithms.strategy.suite.CrossoverOperator
+import org.evomaster.core.search.algorithms.strategy.suite.MutationEvaluationOperator
+import org.evomaster.core.search.algorithms.strategy.suite.DefaultMutationEvaluationOperator
+import org.evomaster.core.search.algorithms.strategy.suite.SelectionStrategy
+import org.evomaster.core.search.algorithms.strategy.suite.TournamentSelectionStrategy
+import org.evomaster.core.search.algorithms.strategy.suite.DefaultCrossoverOperator
 import org.evomaster.core.search.algorithms.observer.GAObserver
 
 /**
@@ -37,14 +40,11 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
      */
     protected var frozenTargets: Set<Int> = emptySet()
 
-    @Inject
-    lateinit var selectionStrategy: SelectionStrategy
+    protected var selectionStrategy: SelectionStrategy = TournamentSelectionStrategy()
 
-    @Inject
-    lateinit var crossoverOperator: CrossoverOperator
+    protected var crossoverOperator: CrossoverOperator = DefaultCrossoverOperator()
 
-    @Inject
-    lateinit var mutationOperator: MutationOperator
+    protected var mutationOperator: MutationEvaluationOperator = DefaultMutationEvaluationOperator()
 
     /** Optional observers for GA events (test/telemetry). */
     protected val observers: MutableList<GAObserver<T>> = mutableListOf()
@@ -111,7 +111,7 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
      * This method modifies the individual in-place.
      */
     protected fun mutate(wts: WtsEvalIndividual<T>) {
-        mutationOperator.mutateIndividual(
+        mutationOperator.mutateEvaluateAndArchive(
             wts,
             config,
             randomness,
@@ -135,6 +135,13 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
         // notify observers
         observers.forEach { it.onCrossover(x, y) }
     }
+
+    /**
+     * Allows tests or callers to override GA operators without DI.
+     */
+    fun useSelectionStrategy(strategy: SelectionStrategy) { this.selectionStrategy = strategy }
+    fun useCrossoverOperator(operator: CrossoverOperator) { this.crossoverOperator = operator }
+    fun useMutationOperator(operator: MutationEvaluationOperator) { this.mutationOperator = operator }
 
     /**
      * Selects one individual using tournament selection.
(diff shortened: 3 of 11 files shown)


core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/DefaultMutationOperator.kt
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Default mutation operator for GA suites.
+ *
+ * Behavior:
+ * - Randomly applies one of: "del", "add", "mod" (selected randomly)
+ * - del: if size > 1, remove a random element.
+ * - add: if size < maxSearchSuiteSize, sample + evaluate; add to suite (and archive GASolutionSource = ARCHIVE).
+ * - mod: mutate one random element; re-evaluate (or mutateAndSave if GASolutionSource = ARCHIVE).
+ */
+class DefaultMutationOperator : MutationOperator {
```
arcuri82: the name might be misleading, as mutation here is just at test suite level and not test case level, isn't it? if so, either "suite" should be in the package name, or class name, somehow

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/DefaultMutationOperator.kt
```diff
@@ -0,0 +1,70 @@
...
+    companion object {
+        private const val OP_DELETE = "del"
+        private const val OP_ADD = "add"
+        private const val OP_MOD = "mod"
+    }
+
+    override fun <T : Individual> mutateIndividual(
```
arcuri82: i don't like the name of this function. this is doing a lot of things, and not just "mutate". name should be more descriptive

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/MutationOperator.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.evomaster.core.EMConfig
+import org.evomaster.core.search.Individual
+import org.evomaster.core.search.algorithms.wts.WtsEvalIndividual
+import org.evomaster.core.search.service.Archive
+import org.evomaster.core.search.service.FitnessFunction
+import org.evomaster.core.search.service.Randomness
+import org.evomaster.core.search.service.Sampler
+import org.evomaster.core.search.service.mutator.Mutator
+
+interface MutationOperator {
```
arcuri82: is this really a mtation operator? you are also evaluating fitness, and saving to archive based on that

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/TournamentSelectionStrategy.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+class TournamentSelectionStrategy : SelectionStrategy {
+    override fun <T : Individual> select(
+        population: List<WtsEvalIndividual<T>>,
+        tournamentSize: Int,
+        randomness: Randomness,
+        score: (WtsEvalIndividual<T>) -> Double
+    ): WtsEvalIndividual<T> {
+        val selected = randomness.choose(population, tournamentSize)
+        return selected.maxByOrNull { score(it) } ?: randomness.choose(population)
```
arcuri82: how can this be null? would that be a bug, or something that could actually happen? here, if it happens it silently handled

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/DefaultMutationOperator.kt
arcuri82: however, looking at the old code, it was named like that as well... :) but i guess it is a good time now for a name update.

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
```diff
@@ -24,6 +31,32 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
...
+    protected var frozenTargets: Set<Int> = emptySet()
+
+    @Inject
+    lateinit var selectionStrategy: SelectionStrategy
+
+    @Inject
+    lateinit var crossoverOperator: CrossoverOperator
+
+    @Inject
+    lateinit var mutationOperator: MutationOperator
```
arcuri82: i don't think it is a good idea to use injection here. that is mainly done for singleton handling, which is not the case here.

core/src/main/kotlin/org/evomaster/core/BaseModule.kt
```diff
@@ -89,6 +95,19 @@ class BaseModule(val args: Array<String>, val noTests: Boolean = false) : Abstra
...
+            .asEagerSingleton()
+
+        bind(CrossoverOperator::class.java)
+            .to(DefaultCrossoverOperator::class.java)
+            .asEagerSingleton()
+
+        bind(MutationOperator::class.java)
+            .to(DefaultMutationOperator::class.java)
+            .asEagerSingleton()
+
```
arcuri82: please remove those. they are not statefull singletons

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
```diff
@@ -141,13 +154,15 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
...
      */
-    private fun sampleSuite(): WtsEvalIndividual<T> {
+    protected fun sampleSuite(): WtsEvalIndividual<T> {
         val n = 1 + randomness.nextInt(config.maxSearchSuiteSize)
         val suite = WtsEvalIndividual<T>(mutableListOf())
 
         for (i in 1..n) {
             ff.calculateCoverage(sampler.sample(), modifiedSpec = null)?.run {
-                archive.addIfNeeded(this)
+                if (config.gaSolutionSource != EMConfig.GASolutionSource.POPULATION) {
```
arcuri82, line 170: why this? @jgaleotti  did we explicitly speak about this? i don't remember... ie, whether we want to evaluated test-level archive for suite generation

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
```diff
@@ -158,4 +173,55 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
...
+     */
+    protected fun score(w: WtsEvalIndividual<T>): Double {
+        if (w.suite.isEmpty()) return 0.0
+
+        // Explicitly use full combined fitness when solution source is POPULATION
+        if (config.gaSolutionSource == EMConfig.GASolutionSource.POPULATION) {
+            return w.calculateCombinedFitness()
+        }
+
+        if (frozenTargets.isEmpty()) return w.calculateCombinedFitness()
```
arcuri82: always use {} for `if` statement, with block on new line

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
```diff
@@ -158,4 +173,55 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
...
+                listOf(),
+                listOf()
+            )
+        } else {
+            super.buildSolution()
+        }
+    }
+
+    /**
+     * Exposes a snapshot copy of the current population (read-only) para observabilidad/tests.
```
arcuri82: the list might be read-only, but not its content. for this kind of methods, we usually use the prefix `getView...`, eg , `getViewOfPopulation`

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
```diff
@@ -158,4 +173,55 @@ abstract class AbstractGeneticAlgorithm<T> : SearchAlgorithm<T>() where T : Indi
...
+     * Se devuelve una copia inmutable para evitar mutaciones externas del estado interno.
```
arcuri82: only comments in English, por favor :)

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2852,6 +2852,19 @@ class EMConfig {
 
     fun isEnabledAIModelForResponseClassification() = aiModelForResponseClassification != AIResponseClassifierModel.NONE
 
+    /**
+     * Source to build the final GA solution.
+     * ARCHIVE: use current behavior (take tests from the archive).
+     * POPULATION: for GA algorithms, take the best individual from the final population.
```
arcuri82: clarify here when evolving full test suites, and not tests in isolation

core/src/test/kotlin/org/evomaster/core/search/algorithms/StandardGeneticAlgorithmTest.kt
```diff
@@ -12,17 +12,20 @@ import org.evomaster.core.search.algorithms.onemax.OneMaxIndividual
...
+import org.evomaster.core.search.service.Randomness
 import org.junit.jupiter.api.Test
 
 import org.junit.jupiter.api.Assertions.*
+import org.evomaster.core.search.algorithms.strategy.FixedSelectionStrategy
+import org.evomaster.core.search.algorithms.observer.GARecorder
 
 class StandardGeneticAlgorithmTest {
 
     val injector: Injector = LifecycleInjector.builder()
```
arcuri82: if you are having more than 1 test in this class, this is problematic. shared injector creates dependencies. should be recreated for each test, eg, in a `@BeforeEach` method

core/src/test/kotlin/org/evomaster/core/search/algorithms/StandardGeneticAlgorithmTest.kt
```diff
@@ -36,15 +39,225 @@ class StandardGeneticAlgorithmTest {
             config.stoppingCriterion = EMConfig.StoppingCriterion.ACTION_EVALUATIONS
 
             val epc = injector.getInstance(ExecutionPhaseController::class.java)
+            if (epc.isInSearch()) {
+                epc.finishSearch()
```
arcuri82: not needed if injector is recreated per test

core/src/test/kotlin/org/evomaster/core/search/algorithms/StandardGeneticAlgorithmTest.kt
```diff
@@ -36,15 +39,225 @@ class StandardGeneticAlgorithmTest {
...
+            }
+            val solution = try {
             epc.startSearch()
-
-            val solution = standardGeneticAlgorithm.search()
-
+                standardGeneticAlgorithm.search()
+            } finally {
```
arcuri82: likely not needed if injector is recreated per test

core/src/main/kotlin/org/evomaster/core/search/algorithms/AbstractGeneticAlgorithm.kt
francastagna: Haha, good catch. Will do, thanks!

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/DefaultMutationOperator.kt
francastagna: What do u suggest? @arcuri82
I'm thinking... mutateEvaluateAndArchiveIfNeeded? 😬

core/src/main/kotlin/org/evomaster/core/search/algorithms/strategy/DefaultMutationOperator.kt
francastagna: You're right it can be misleading, I'll add suite to the package name.

PR #1353, opened 2025-10-20
Update Statistics.kt

MohsenTaheriShalmani:


core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -126,7 +130,11 @@ class Statistics : SearchListener {
             }
         }
 
-        val headers = "interval," + snapshots.values.first().map { it.header }.joinToString(",")
+        val baseHeaders = snapshots.values.first().map { it.header }.toMutableList()
+        if (config.isEnabledAIModelForResponseClassification()) {
```
arcuri82: this is problematic. headers should always be added. if there is no data, as not used, then can fill it with 0s. 
the reason is that, when running experiments, we merge all final data into a single CSV file. if those have different numbers of columns (and/or in different order), it will not work

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -140,11 +148,16 @@ class Statistics : SearchListener {
...
         snapshots.entries.stream()
-                .sorted { o1, o2 -> o1.key.compareTo(o2.key) }
-                .forEach {
-                    val elements = it.value.map { it.element }.joinToString(",")
-                    path.toFile().appendText("${it.key},$elements\n")
+            .sorted { o1, o2 -> o1.key.compareTo(o2.key) }
+            .forEach {
+                val baseElements = it.value.map { it.element }.toMutableList()
+                if (config.isEnabledAIModelForResponseClassification()) {
+                    baseElements.addAll(getAIData().map { it.element })
```
arcuri82: see previous comment

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -342,6 +355,37 @@ class Statistics : SearchListener {
         return list
     }
 
+    fun getAIData(): List<Pair> {
+        val list = mutableListOf<Pair>()
+
+        // Check if AI classifier is enabled
+        if (!config.isEnabledAIModelForResponseClassification()) {
+            list.add(Pair("ai_model_enabled", "false"))
```
arcuri82: is this necessary? we already have `aiModelForResponseClassification` in the CSV file

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -342,6 +355,37 @@ class Statistics : SearchListener {
...
+            return list
+        }
+
+        list.add(Pair("ai_model_enabled", "true"))
+
+        // Try to get classifier and its metrics
+        val classifier = aiResponseClassifier
+        if (classifier == null) {
+            list.add(Pair("ai_status", "not_initialized"))
```
arcuri82: presence of columns must be constant, and not parametric.
also, if `config.isEnabledAIModelForResponseClassification()` is true, how can `classifier` be null?

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -342,6 +355,37 @@ class Statistics : SearchListener {
...
+
+        val model = classifier.viewInnerModel()
+        val metrics = model.estimateOverallMetrics()
+
+        // Add derived metrics
+        list.add(Pair("ai_accuracy", "%.4f".format(metrics.accuracy)))
+        list.add(Pair("ai_precision400", "%.4f".format(metrics.precision400)))
+        list.add(Pair("ai_recall400", "%.4f".format(metrics.recall400)))
+        list.add(Pair("ai_f1Score400", "%.4f".format(metrics.f1Score400)))
+        list.add(Pair("ai_mcc400", "%.4f".format(metrics.mcc)))
```
arcuri82: these entries should always be filled. if `classifier` is null, I guess 0 could be used

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -126,28 +130,47 @@ class Statistics : SearchListener {
             }
         }
 
-        val headers = "interval," + snapshots.values.first().map { it.header }.joinToString(",")
+        // All headers, including AI-related ones
+        val baseHeaders = snapshots.values.first().map { it.header }.toMutableList()
+        baseHeaders.addAll(
+            listOf(
+                "ai_model_enabled",
```
arcuri82: avoid duplicated strings. if the same string is used more than once (like here it seems at least 3 times) should be in a constant (eg under `companion object`)

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -126,28 +130,47 @@ class Statistics : SearchListener {
             }
         }
 
-        val headers = "interval," + snapshots.values.first().map { it.header }.joinToString(",")
+        // All headers, including AI-related ones
+        val baseHeaders = snapshots.values.first().map { it.header }.toMutableList()
```
arcuri82: the snapshot does not contain the AI data? why wasn't `getAIData()` simply done inside `getData(solution)`?

PR #1377, opened 2025-11-03
refactor: moving mongo classes to internal.db.mongo

jgaleotti:
Specific Mongo classes moved from controller.internal.db into controller.internal.db.mongo


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoCommandWithDistance.java
```diff
@@ -0,0 +1,25 @@
+package org.evomaster.client.java.controller.internal.db.mongo;
```
arcuri82, line 1: files should be "moved" and not "deleted/added", as in this latter case we would lose git history, and makes harder to check if introduced any error by mistake

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoDistanceWithMetrics.java
```diff
@@ -0,0 +1,40 @@
+package org.evomaster.client.java.controller.internal.db.mongo;
```
arcuri82, line 1: files should be "moved" and not "deleted/added", as in this latter case we would lose git history, and makes harder to check if introduced any error by mistake

core/src/test/kotlin/org/evomaster/core/parser/GenePostgresLikeVisitorTest.kt
```diff
@@ -76,4 +76,6 @@ class GenePostgresLikeVisitorTest : RegexTestTemplate(){
     fun testPercentWithAlsoEscape(){
         check("\\%%", "%.*")
     }
-}
\ No newline at end of file
+
+
```
arcuri82, line 80: why this in this PR?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoCommandWithDistance.java
jgaleotti: this is weird. I used the intelliJ IDE to refactor these classes. I am not sure why it is not considered a moved instead of a deleted/added

core/src/test/kotlin/org/evomaster/core/parser/GenePostgresLikeVisitorTest.kt
jgaleotti: I guess this is an involuntary edit. I might have not studied the changed files in detail. I am sorry for this.

PR #1391, opened 2025-11-23
Sqli starter

omursahin:


core-tests/e2e-tests/spring/spring-rest-mysql/src/test/kotlin/org/evomaster/e2etests/spring/rest/sqli/SQLiMySQLBodyEMTest.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    fun testRunEM() {
+
+        runTestHandlingFlakyAndCompilation(
+            "SQLiMySQLBodyEM",
+            100
+        ) { args ->
+            setOption(args, "security", "true")
```
arcuri82, line 29: we will need an experimental EMConfig option for SQLi, which should be off by default (until we run experiments)

core-tests/e2e-tests/spring/spring-rest-openapi-v3/pom.xml
```diff
@@ -150,6 +150,14 @@
...
         </dependency>
 
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+        </dependency>
```
arcuri82: why are those dependencies here? they point of having separated modules for E2E for different databases was to avoid possible conflicts and configuration when starting spring (as it scans the classpath)

core/src/main/kotlin/org/evomaster/core/problem/httpws/HttpWsCallResult.kt
```diff
@@ -126,4 +127,7 @@ abstract class HttpWsCallResult : EnterpriseActionResult {
      */
     fun setVulnerableForSSRF(on: Boolean) = addResultValue(VULNERABLE_SSRF, on.toString())
     fun getVulnerableForSSRF() : Boolean = getResultValue(VULNERABLE_SSRF)?.toBoolean() ?: false
+
+    fun setResponseTime(responseTime: Long) = addResultValue(RESPONSE_TIME, responseTime.toString())
+    fun getResponseTime(): Long = getResultValue(RESPONSE_TIME)?.toLong() ?: 0
```
arcuri82: in case of errors, this is not going to be set (same as status code). so maybe better to return a nullable, ie, `Long?`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/RestSecurityOracle.kt
```diff
@@ -222,6 +222,24 @@ object RestSecurityOracle {
         return true
     }
 
+    // Simple SQLi payloads
```
arcuri82: `/** */`, so it gets into the documentation of the variable

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -592,7 +594,13 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
         val appliedLink = handleLinks(a, all,actionResults)
 
         val response = try {
-            createInvocation(a, chainState, cookies, tokens).invoke()
+            SearchTimeController.measureTimeMillis(
+                { t, res ->
+                    rcr.setResponseTime(t)
+                },
+                {createInvocation(a, chainState, cookies, tokens).invoke()}
```
arcuri82: the `createInvocation()` should be not counted. rather save it to variable `call`, and then measure time of `call.invoke()`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1316,6 +1325,37 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
         }
     }
 
+    private fun handleSQLiCheck(
+        individual: RestIndividual,
+        actionResults: List<ActionResult>,
+        fv: FitnessValue
+    ) {
+        if (!config.isEnabledFaultCategory(DefinedFaultCategory.XSS)) {
```
arcuri82: xss???  also need a check on SQLi, which would be off by default, as still experimental. 
however, maybe we should avoid long checks like:

`!config.xss || !config.isEnabledFaultCategory(DefinedFaultCategory.XSS)`

where the check for `config.xss` (and new `!config.sqli`) should be done inside `isEnabledFaultCategory`. eg something like:

```
    fun isEnabledFaultCategory(category: FaultCategory) : Boolean{
       if(category==DefinedFaultCategory.XSS && !xss){
          return false;
       }  
       return category !in getDisabledOracleCodesList()
    }
```

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1316,6 +1325,37 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+        for(index in individual.seeMainExecutableActions().indices){
+            val a = individual.seeMainExecutableActions()[index]
+            val r = actionResults.find { it.sourceLocalId == a.getLocalId() } as? RestCallResult
+                ?: continue
+
+//            if(!r.getTimedout()){
+//                continue
+//            }
+
+            if(r.getResponseTime() < config.sqlInjectionMaxResponseTimeMs && !r.getTimedout()){
```
arcuri82: this is tricky, as there might be long executing tests regardless of SQLi, and still not timingout. 
first, need to check if SQLi payload is in the test.
then, likely we need a new value in `RestCallResult` to mark the increase in time _after_  the payload. recall what wrote in `notes.txt` (if unclear, we can have a meeting about it)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,108 @@ class SecurityRest {
...
+                genes.forEach {
+                    gene ->
+                    val leafGene = gene.getLeafGene()
+                    if(leafGene !is StringGene) return@forEach
+
+                    // we need to do this way because we need to append our paylod
+                    val hasInvalidChars = leafGene.invalidChars.any { payload.contains(it) }
+                    if(!hasInvalidChars){
+                        // append the SQLi payload value
+                        leafGene.value = leafGene.value + payload
```
arcuri82: the gene might have length constraints. still need to validate such constraints would not break the validity. furthermore, the StringGene might not be be in use due to tainted value. might need to call `getPhenotype()` to make sure that `leafGene.value` is indeed in use

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,108 @@ class SecurityRest {
...
+                }
+
+                if(!anySuccess){
+                    continue
+                }
+
+                copy.modifySampleType(SampleType.SECURITY)
+                copy.ensureFlattenedStructure()
+
+                val evaluatedIndividual = fitness.computeWholeAchievedCoverageForPostProcessing(copy)
```
arcuri82: here we need check the execution time _before_ and _after_ the injection, to see if it was successful. but, then, reexecuting the tests would lose such info :(  we need to think of a way to handle this properly (i don't have an easy solution right now)

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2597,6 +2597,10 @@ class EMConfig {
     @Cfg("To apply SSRF detection as part of security testing.")
     var ssrf = false
 
+    @Experimental
+    @Cfg("Maximum response time (in milliseconds) to consider a potential SQL Injection vulnerability.")
+    var sqlInjectionMaxResponseTimeMs = 2000
```
arcuri82: recall `notes.txt`, there are 2 values to handle (there called N and K)

arcuri82: also need experimental config parameter for SQLi, false by default

core/src/main/kotlin/org/evomaster/core/problem/httpws/HttpWsCallResult.kt
arcuri82: this is still returning `Long` instead of `Long?`

core/src/main/kotlin/org/evomaster/core/problem/httpws/HttpWsCallResult.kt
```diff
@@ -126,4 +127,7 @@ abstract class HttpWsCallResult : EnterpriseActionResult {
      */
     fun setVulnerableForSSRF(on: Boolean) = addResultValue(VULNERABLE_SSRF, on.toString())
     fun getVulnerableForSSRF() : Boolean = getResultValue(VULNERABLE_SSRF)?.toBoolean() ?: false
+
+    fun setResponseTime(responseTime: Long?) = addResultValue(RESPONSE_TIME, responseTime.toString())
```
arcuri82: why `Long?` as input? if it is null, this method should simply not be called. ie, input should be `Long`

core/src/main/kotlin/org/evomaster/core/problem/httpws/HttpWsCallResult.kt
```diff
@@ -25,6 +25,7 @@ abstract class HttpWsCallResult : EnterpriseActionResult {
         const val TCP_PROBLEM = "TCP_PROBLEM"
         const val APPLIED_LINK = "APPLIED_LINK"
         const val LOCATION = "LOCATION"
+        const val RESPONSE_TIME = "RESPONSE_TIME"
```
arcuri82: what is the measure unit? minutes? if it is milliseconds, rename it into `RESPONSE_TIME_MS` (as well as getter/setter)

core/src/main/kotlin/org/evomaster/core/problem/rest/data/RestCallAction.kt
```diff
@@ -434,4 +434,5 @@ class RestCallAction(
         this.weakReference = wr
         return copy
     }
-}
\ No newline at end of file
+    var baseResponseTime: Long? = 0
```
arcuri82: why a nullable variable is initialized with `0` instead of `null`?

arcuri82: also info about responses should NOT be in the call, but rather in the response object

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1346,6 +1359,49 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+
+        for(index in individual.seeMainExecutableActions().indices){
+            val a = individual.seeMainExecutableActions()[index]
+            val r = actionResults.find { it.sourceLocalId == a.getLocalId() } as? RestCallResult
+                ?: continue
+
+            val baseline = a.baseResponseTime
+            val afterPayload = r.getResponseTime()
+            val K = config.sqliBaselineMaxResponseTimeMs        // K: maximum allowed baseline response time
+            val N = config.sqliInjectedSleepDuration * 1000          // N: expected delay introduced by the injected sleep payload
```
arcuri82: isn't confusing that two related configs are using different time units? ie, seconds and millisecond. shouldn't both use milliseconds?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1346,6 +1359,49 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+        if (!config.isEnabledFaultCategory(DefinedFaultCategory.SQL_INJECTION)) {
+            return
+        }
+
+        for(index in individual.seeMainExecutableActions().indices){
+            val a = individual.seeMainExecutableActions()[index]
+            val r = actionResults.find { it.sourceLocalId == a.getLocalId() } as? RestCallResult
+                ?: continue
+
+            val baseline = a.baseResponseTime
```
arcuri82: the `baseline` should not be saved in the action

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1346,6 +1359,49 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+            val afterPayload = r.getResponseTime()
+            val K = config.sqliBaselineMaxResponseTimeMs        // K: maximum allowed baseline response time
+            val N = config.sqliInjectedSleepDuration * 1000          // N: expected delay introduced by the injected sleep payload
+
+            // Baseline must be fast enough (baseline < K)
+            val baselineIsFast = baseline!! < K
```
arcuri82: what if `baseline` is null? this will crash

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1346,6 +1359,49 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+    ) {
+        if (!config.isEnabledFaultCategory(DefinedFaultCategory.SQL_INJECTION)) {
+            return
+        }
+
+        for(index in individual.seeMainExecutableActions().indices){
+            val a = individual.seeMainExecutableActions()[index]
+            val r = actionResults.find { it.sourceLocalId == a.getLocalId() } as? RestCallResult
+                ?: continue
+
```
arcuri82, line 1376: before those time-checks, shouldn't we see if there is a SQLi payload in this action?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -1346,6 +1359,49 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
...
+            val N = config.sqliInjectedSleepDuration * 1000          // N: expected delay introduced by the injected sleep payload
+
+            // Baseline must be fast enough (baseline < K)
+            val baselineIsFast = baseline!! < K
+
+            // Response after injection must be slow enough (response > N)
+            val responseIsSlowEnough = afterPayload > N
+
+            // Timeout is also considered a potential vulnerability indicator
+            val isTimeout = r.getTimedout()
```
arcuri82: i would remove this. timeouts happen a few times, regardless of SQLi

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
             handleStackTraceCheck()
         }
 
+        if (!config.isEnabledFaultCategory(DefinedFaultCategory.SQL_INJECTION)) {
+            LoggingUtil.uniqueUserInfo("Skipping experimental security test for sql injection as disabled in configuration")
```
arcuri82: we should remove these `Skipping ` logs, as they are now flooding the logs. if needed, they can be put as `debug` option, and not UserInfo

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
...
+            if(actionIndex < 0){
+                continue
+            }
+
+            // Slice to keep only up to the target action
+            val sliced = RestIndividualBuilder.sliceAllCallsInIndividualAfterAction(
+                target.individual,
+                actionIndex
+            )
+            val responseTime = target.evaluatedMainActions()[actionIndex].result.getResultValue("RESPONSE_TIME")?.toLong()
```
arcuri82: what if `"RESPONSE_TIME"` is null? this should not happen, because action returned 2xx.
but should still have an `if` check, in which we log a `warn` if it happens, and continue isntead of crashed. in that block, can also have a `assert(false)` to crash tests if it happens, but not production

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
...
+                    val leafGene = gene.getLeafGene()
+                    if(leafGene !is StringGene) return@forEach
+
+                    // we need to do this way because we need to append our payload
+
+                    //check invalid chars
+                    val hasInvalidChars = leafGene.invalidChars.any { payload.contains(it) }
+
+                    //check max length
+                    val hasMaxLength = (leafGene.getPhenotype().getValueAsRawString().length + payload.length) > leafGene.maxLength
```
arcuri82: no, here you are duplicating the internal checks of StringGene. what if those change in future? better to delegate to StringGene, eg using something like `setFromStringValue()`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
...
+
+                    if(!hasInvalidChars && !hasMaxLength){
+                        // append the SQLi payload value
+                        leafGene.getPhenotype().setFromStringValue(leafGene.getPhenotype().getValueAsRawString() + String.format(payload, config.sqliInjectedSleepDuration))
```
arcuri82: `setFromStringValue` returns false if it fails to update

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
...
+                }
+                actionCopy.baseResponseTime = responseTime
+
+                copy.modifySampleType(SampleType.SECURITY)
+                copy.ensureFlattenedStructure()
+
+                val evaluatedIndividual = fitness.computeWholeAchievedCoverageForPostProcessing(copy)
+
+                if (evaluatedIndividual == null) {
+                    log.warn("Failed to evaluate constructed individual in handleStackTraceCheck")
```
arcuri82: `handleStackTraceCheck`???

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -303,10 +301,115 @@ class SecurityRest {
...
+                        // append the SQLi payload value
+                        leafGene.getPhenotype().setFromStringValue(leafGene.getPhenotype().getValueAsRawString() + String.format(payload, config.sqliInjectedSleepDuration))
+                        anySuccess = true
+                    }
+                }
+
+                if(!anySuccess){
+                    continue
+                }
+                actionCopy.baseResponseTime = responseTime
```
arcuri82: this approach is hacky, and would not work if test is re-evaluated for any reason.  i think we need a singleton, in which we keep track, for each action,  the min/mean/max exection time during the search (and NOT during security phase). then, when we need a baseline, we query the singleton.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2605,6 +2605,19 @@ class EMConfig {
     @Cfg("To apply XSS detection as part of security testing.")
     var xss = false
 
+    @Experimental
+    @Cfg("To apply SQLi detection as part of security testing.")
+    var sqli = false
+
+    @Experimental
+    @Cfg("Injected sleep duration (in seconds) used inside the malicious payload to detect time-based vulnerabilities.")
+    var sqliInjectedSleepDuration = 5
```
arcuri82: be consistent with time units

docs/options.md
```diff
@@ -309,6 +309,9 @@ There are 3 types of options:
 |`seedTestCases`| __Boolean__. Whether to seed EvoMaster with some initial test cases. These test cases will be used and evolved throughout the search process. *Default value*: `false`.|
```
arcuri82, line 309: let's not forget about `generated test cases (in Java/Kotlin/Python/Javascript) need to then log the execution time (eg startTime-endTime) and assert it is more than N.` but best to do it in a new PR _after_ this one is merged

core/src/main/kotlin/org/evomaster/core/problem/rest/data/RestCallAction.kt
omursahin: Now, I am keeping them in the ExecutionStats class.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
omursahin: I removed it here as I started using the ExecutionStats class to check execution time.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
omursahin: I added "src/main/kotlin/org/evomaster/core/search/service/ExecutionStats.kt" class for this.

PR #1393, opened 2025-11-27
Add support to xml data type

suarezrominajulieta:


core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -529,8 +529,21 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
                     }
                     else -> lines.add(".$send(\"$body\")")
                 }
+            } else if (bodyParam.isXml()) {
+
+                val xml = bodyParam.getValueAsPrintableString(mode = GeneUtils.EscapeMode.XML, targetFormat = format)
+
+                when {
+
+                    format.isCsharp() -> {
```
arcuri82: don't add for `isCsharp()`, as it is deprecated, and no longer in use (although we haven't cleaned the code to remove its occurrences)

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
```diff
@@ -748,8 +747,12 @@ object RestActionBuilderV3 {
             listOf()
         }
 
-        var gene = getGene("body", obj.schema, schemaHolder,currentSchema, referenceClassDef = null, options = options, messages = messages, examples = examples)
+        val deref = obj.schema.`$ref`?.let { ref -> val name = ref.substringAfterLast("/")
```
arcuri82: hard to read. use better indentation, eg:
```
val deref = obj.schema.`$ref`?.let { ref -> 
                            val name = ref.substringAfterLast("/")
                            SchemaUtils.getReferenceSchema(schemaHolder, currentSchema, ref, messages) 
                   } ?: obj.schema
```
also, is `name` variable used in this block?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -304,6 +309,99 @@ class ObjectGene(
...
     }
 
+    private fun escapeXmlSafe(s: String): String =
+        s.replace(Regex("(?<!&)&(?![a-zA-Z]+;)"), "&amp;")
+            .replace("<", "&lt;")
+            .replace(">", "&gt;")
+            .replace("\"", "&quot;")
+            .replace("'", "&apos;")
+
+    private fun singularize(n: String): String =
```
arcuri82: what is this for? also, it would not be working for irregular terms, eg, foot vs feet. 
also, if really needed, should go into `org.evomaster.core.utils.StringUtils`

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -304,6 +309,99 @@ class ObjectGene(
...
+        when {
+            n.endsWith("s") && n.length > 1 -> n.removeSuffix("s")
+            else -> n
+        }.replaceFirstChar { it.uppercase() }
+
+    private fun unwrap(v: Any?): Any? =
```
arcuri82: you want to unwrap genes, need to use `Gene.getLeafGene()`. 
also if needed, generic utils on genes should be in `GeneUtils`

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -304,6 +309,99 @@ class ObjectGene(
...
+        when (v) {
+            is OptionalGene -> v.gene
+            else -> v
+        }
+
+    public fun cleanXmlValueString(v: String): String =
+        v.removeSurrounding("\"").let(::escapeXmlSafe)
+
+    private fun getPrintedValue(
```
arcuri82: `getPrintedValueForXml`?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -304,6 +309,99 @@ class ObjectGene(
...
+    ): String =
+        cleanXmlValueString(
+            v.getValueAsPrintableString(
+                previousGenes,
+                GeneUtils.EscapeMode.XML,
+                targetFormat
+            )
+        )
+
+    private fun isPrimitiveGene(value: Any?): Boolean =
```
arcuri82: what is the definition of "primitive" here? and is it something specific for XML only? many other genes could have to be included here... eg Regex, Uri, URL, and so on. 
Depending on how it is used, and its definition, might need to consider to add a `Gene.isPrimitive`

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
```diff
@@ -304,6 +309,99 @@ class ObjectGene(
...
+            is ArrayGene<*> -> {
+                val itemName = singularize(name)
+                v.getViewOfElements().joinToString("", "<$name>", "</$name>") {
+                    serializeXml(previousGenes, itemName, it, targetFormat)
+                }
+            }
+
+            is Gene -> "<$name>${getPrintedValue(previousGenes, v, targetFormat)}</$name>"
+
+            else -> "<$name>${cleanXmlValueString(v.toString())}</$name>"
```
arcuri82: are we actually considering cases in which `v` is not a `Gene`? how would that happen? leave comment in the code if indeed it is possible

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        // 2) Child names must be unique (XML does not allow repeated element names at this level)
+        val duplicated = childFields
+            .groupBy { it.name }
+            .filter { it.value.size > 1 }
+            .keys
+
+        if (duplicated.isNotEmpty()) {
+            throw IllegalStateException("Duplicate child elements not allowed in XML: $duplicated")
```
arcuri82: this type of checks on the state of `this` fields should be done at construction time as well. also depends if they are mutable. as this is info from the schema (which can have mistakes), we should not crash, but rather flags errors in `RestActionBuilderV3`

core/src/test/kotlin/org/evomaster/core/search/gene/ObjectGeneTest.kt
```diff
@@ -101,4 +101,40 @@ internal class ObjectGeneTest {
 
         assertEquals("{foo,bar,nested{hello}}", actual)
     }
+
```
arcuri82, line 105: we also need E2E tests (ask @jgaleotti  to show/explain you how they work):
- at least 1 white-box in `core-tests/e2e-tests/spring/spring-rest-openapi-v3`
- 1 in black-box in `core-tests/e2e-tests/spring/spring-rest-bb`

for example, could have API with endpoint same name but supporting incoming payload in both JSON and XML. for XML, could check some attributes are properly handled/parsed

arcuri82 (changes requested):
@suarezrominajulieta  thx for your first PR! see comments

suarezrominajulieta:
thank you for the comments! I'll look into them

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
suarezrominajulieta: Done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
suarezrominajulieta: Done, used the reference of the childs objects instead.

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
suarezrominajulieta: Fixed and absorved

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
suarezrominajulieta: We are not, fixed

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
suarezrominajulieta: Done, I've improved identation and removed the unused "name" variable.

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt
suarezrominajulieta: Done, i verified if the object had childs instead.

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done: i moved the validation to the constructor, and added logs to restBuilderAction

core/src/test/kotlin/org/evomaster/core/search/gene/ObjectGeneTest.kt
suarezrominajulieta: Done. I've added:
- a White-box E2E test in spring-rest-openapi-v3: XMLEMTest
- a Black-box E2E test in spring-rest-bb: BBXMLTest

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/xml/BBXMLTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        @JvmStatic
+        fun init() {
+            val config = EMConfig()
+            initClass(BBXMLController(), config)
+        }
+    }
+
+    @Test
+    fun testRunEM() {
+        runTestHandlingFlakyAndCompilation(
```
arcuri82: this is not how BB tests should be written in this module. they should follow the structure:

```

    @ParameterizedTest
    @EnumSource
    fun testBlackBoxOutput(outputFormat: OutputFormat) {

        executeAndEvaluateBBTest(
```

see all other tests in this module, in particular the use of `CoveredTargets.cover()`

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
```diff
@@ -742,8 +741,10 @@ object RestActionBuilderV3 {
             listOf()
         }
 
-        var gene = getGene("body", obj.schema, schemaHolder,currentSchema, referenceClassDef = null, options = options, messages = messages, examples = examples)
+        val deref = obj.schema.`$ref`?.let { ref -> SchemaUtils.getReferenceSchema(schemaHolder, currentSchema, ref, messages) } ?: obj.schema
```
arcuri82, line 745: this is unclear. need to add some comments to explain "why" this is done.

core/src/main/kotlin/org/evomaster/core/search/gene/collection/ArrayGene.kt
```diff
@@ -225,7 +225,14 @@ class ArrayGene<T>(
     }
 
     override fun getValueAsPrintableString(previousGenes: List<Gene>, mode: GeneUtils.EscapeMode?, targetFormat: OutputFormat?, extraCheck: Boolean): String {
-        return openingTag +
+
+        val(open,close,sep) = if (mode == GeneUtils.EscapeMode.XML){
```
arcuri82, line 229: add a comment to explain this

core/src/test/kotlin/org/evomaster/core/search/gene/ArrayGeneTest.kt
```diff
@@ -123,4 +124,89 @@ class ArrayGeneTest {
...
+                IntegerGene("num2", 20),
+                IntegerGene("num3", 30)
+            )
+        )
+
+        val jsonOutput = intArray.getValueAsPrintableString(mode = GeneUtils.EscapeMode.JSON)
+
+        assertEquals("[10, 20, 30]", jsonOutput)
+
+        assertTrue(jsonOutput.startsWith("["), "JSON debe empezar con corchete de apertura")
```
arcuri82: error messages should not be in Spanish, but English :)

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
```diff
@@ -74,7 +75,39 @@ class GeneRandomizedTest : AbstractGeneTest(){
                 //assertTrue(x.containsSameValueAs(y))
             } else {
                 //they must be different. the same genotype must not lead to different phenotypes
-                assertFalse(x.containsSameValueAs(y))
+                val containsSame = x.containsSameValueAs(y)
+                if (containsSame) {
+                    // Debug info: find which child gene has the inconsistency
+                    val debugInfo = buildString {
```
arcuri82: this could be refactored to a call to a private method

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
```diff
@@ -74,7 +75,39 @@ class GeneRandomizedTest : AbstractGeneTest(){
...
+                                    appendLine("  INCONSISTENT FIELD[$i]: ${xf.javaClass.simpleName}:${xf.name}")
+                                    appendLine("    xfv: $xfv")
+                                    appendLine("    yfv: $yfv")
+                                    appendLine("    containsSameValueAs: $same")
+                                }
+                            }
+                        }
+                        appendLine("=== END DEBUG ===")
+                    }
+                    System.err.println(debugInfo)
```
arcuri82: this `debugInfo` should not go on `System.err.println`, but rather appended to the error message of the `assertFalse`

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
```diff
@@ -110,8 +143,8 @@ class GeneRandomizedTest : AbstractGeneTest(){
         assertTrue(gene.isLocallyValid(), msg)
         //all tree must be valid, regardless of impact on phenotype
         assertTrue(gene.flatView().all {
-                it.isLocallyValid()
-            }
+            it.isLocallyValid()
+        }
```
arcuri82: indentation here does not seem nice. should avoid changing indentation of existing code

core/src/test/kotlin/org/evomaster/core/search/gene/ArrayGeneTest.kt
```diff
@@ -123,4 +124,89 @@ class ArrayGeneTest {
...
+        assertTrue(jsonOutput.endsWith("]"), "JSON debe terminar con corchete de cierre")
+        assertTrue(jsonOutput.contains(", "), "JSON debe contener comas separadoras")
+    }
+
+    @Test
+    fun testXmlSerializationEmptyArray(){
+        // Test para verificar que un array vacío en XML no genera corchetes
```
jgaleotti: comments should only be written in English

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
suarezrominajulieta: i'm Sorry, this file was not supposed to be modified.

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
suarezrominajulieta: i'm Sorry, this file was not supposed to be modified

core/src/test/kotlin/org/evomaster/core/search/gene/GeneRandomizedTest.kt
suarezrominajulieta: i'm Sorry, this file was not supposed to be modified

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
suarezrominajulieta: I added this because $ref schemas do not carry XML-specific metadata (such as xml.name), which is defined on the referenced schema itself.
Without resolving the reference, XML element names for arrays and object items would be generated incorrectly (e.g. using the ref name instead of the schema-defined xml.name).

For example:

people:
  type: array
  items:
    $ref: '#/components/schemas/Person'

Person:
  type: object
  xml:
    name: person


Without dereferencing, the generated XML element would be <Person>, while the OpenAPI contract explicitly defines <person>.

suarezrominajulieta:
Hi, i made the changes requested: i fixed the bbxml test, add the comments, removed the spanish comments, and the changes to the GeneRandomisedTest, that i didn't want to change, it was a mistake. Thank you!

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+import org.evomaster.core.search.gene.collection.PairGene
+import org.evomaster.core.search.gene.placeholder.CycleObjectGene
+import org.evomaster.core.search.gene.string.StringGene
+import org.evomaster.core.search.gene.utils.GeneUtils
+import org.evomaster.core.search.gene.wrapper.OptionalGene
+import org.evomaster.core.utils.CollectionUtils
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class ObjectWithAttributesGene(
```
arcuri82, line 24: add couple of sentences in `/** */` to state differences of this class from ObjectGene

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+    name: String,
+    fixedFields: List<Gene>,
+    refType: String? = null,
+    isFixed: Boolean,
+    template: PairGene<StringGene, Gene>? = null,
+    additionalFields: MutableList<PairGene<StringGene, Gene>>? = null,
+    val attributeNames: Set<String> = emptySet()
```
arcuri82, line 39: specify how/what this `attributeNames` name is

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    init {
+
+        val includedFields = fixedFields
+            .filter { it !is CycleObjectGene }
+            .filter { it !is OptionalGene || (it.isActive && it.gene !is CycleObjectGene) }
+            .filter { it.isPrintable() }
+
+        val attributeFields = includedFields.filter { attributeNames.contains(it.name) }
```
arcuri82: is this `attributeFields` variable used anywhere?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+
+        // "#text" CANNOT be an attribute - warn if schema incorrectly defines it as one
+        if (attributeNames.contains("#text")) {
+            LoggingUtil.uniqueWarn(log, "Invalid XML schema: '#text' cannot be used as an attribute. It will be treated as content instead.")
```
arcuri82: should rather throw an IllegalArgument exception

arcuri82: also, is there any invariant/property that should old for this `attributeNames` input field? eg, should all of its content defined under `fixedFields`? ie, what happen if we have a name in `attirbuteNames` that is no present in the fields of the object?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+            ?.toMutableList()
+
+        return ObjectWithAttributesGene(
+            name,
+            fixedFields.map { it.copy() },
+            refType,
+            isFixed,
+            template,
+            copiedAdditional,
+            attributeNames.toMutableSet()
```
arcuri82: why to a mutable set?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+    override fun containsSameValueAs(other: Gene): Boolean {
+
+        if (other !is ObjectGene) {
+            return false
+        }
+
+        // If other is also ObjectWithAttributesGene, attributeNames must match
+        // If other is plain ObjectGene, this.attributeNames must be empty to produce same XML
+        if (other is ObjectWithAttributesGene) {
+            if (this.attributeNames != other.attributeNames) {
```
arcuri82, line 88: is this correct? isn't it checking reference equality instead of set content?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
```diff
@@ -0,0 +1,160 @@
...
+            return super.getValueAsPrintableString(previousGenes, mode, targetFormat, extraCheck)
+        }
+
+        val includedFields = fixedFields
+            .filter { it !is CycleObjectGene }
+            .filter { it !is OptionalGene || (it.isActive && it.gene !is CycleObjectGene) }
+            .filter { it.isPrintable() }
+
+        val attributeFields = includedFields.filter { attributeNames.contains(it.name) }
+        val childFields = includedFields.filter { !attributeNames.contains(it.name) }
```
arcuri82: should rather use `partition{}` instead of `filter`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Hi @arcuri82 — correct me if I'm wrong, but in Kotlin != calls .equals() under the hood, so this is comparing by content, not by reference (reference identity would require !==). Since Set.equals() checks that both sets contain the same elements regardless of order, I believe the comparison is correct: two ObjectWithAttributesGene instances with the same attribute names — even if constructed separately — should be considered equivalent for the purposes of this method. Please let me know if you had something else in mind!

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Right now, if the object is in the set of attributes but not in fixedFields, nothing happens; it's not printed. Should we throw an exception instead?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
arcuri82: hi @suarezrominajulieta  you might be right. but can you add a JUnit unit test to check this implementation of `containsSameValueAs`, just in case?

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
arcuri82: yep. throw exception

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: No, you are correct. it was wrongly there for the previous validation. 

Now it's there and changed for the throw exception if attribute name is not in fixedFields.

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Fixed.

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done

core/src/main/kotlin/org/evomaster/core/search/gene/ObjectWithAttributesGene.kt
suarezrominajulieta: Done.

PR #1404, opened 2025-12-11
Update the condition

MohsenTaheriShalmani:
The MCC condition is added to detect weak classifiers, especially in class-imbalanced cases where all responses may be identical (e.g., all 400s). In such situations, accuracy and precision can appear artificially high even though the classifier has learned no real discrimination. MCC prevents this by identifying cases where the model is effectively non-informative, avoiding unnecessary or maximal repairs.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
+++ b/core/src/main/kotlin/org/evomaster/core/EMConfig.kt
@@ -1446,8 +1446,7 @@ class EMConfig {
 
     @Experimental
     @Cfg("Determines whether the AI response classifier skips model updates when the response " +
-            "indicates a server-side error with status code 500. Such responses can be considered non-informative, " +
-            "as they can be caused by both valid and invalid requests.")
+            "indicates a server-side error with status code 500.")
     var skipAIModelUpdateWhenResponseIs500 = false
 
     @Cfg("Output a JSON file representing statistics of the fuzzing session, written in the WFC Report format." +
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1444,6 +1444,12 @@ class EMConfig {
     @Cfg("Determines which metric-tracking strategy is used by the AI response classifier.")
     var aIClassificationMetrics = AIClassificationMetrics.TIME_WINDOW
 
+    @Experimental
+    @Cfg("Determines whether the AI response classifier skips model updates when the response " +
+            "indicates a server-side error with status code 500. Such responses can be considered non-informative, " +
+            "as they can be caused by both valid and invalid requests.")
+    var skipAIModelUpdateWhenResponseIs500 = true
```
arcuri82: experimental boolean options should be `false` by default.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1444,6 +1444,12 @@ class EMConfig {
     @Cfg("Determines which metric-tracking strategy is used by the AI response classifier.")
     var aIClassificationMetrics = AIClassificationMetrics.TIME_WINDOW
 
+    @Experimental
+    @Cfg("Determines whether the AI response classifier skips model updates when the response " +
+            "indicates a server-side error with status code 500. Such responses can be considered non-informative, " +
```
arcuri82: remove the sentence `Such responses can be considered non-informative, " +            "as they can be caused by both valid and invalid requests.` as i do not agree with it

PR #1405, opened 2025-12-15
End to end tests for Redis.

aszyrej:
First e2e-test for Redis. 
Includes GET and SET commands, applying Redis heuristics.

### DIFF
--- a/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
+++ b/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
@@ -356,12 +356,6 @@ public abstract class SutController implements SutHandler, CustomizationHandler 
     public final void initRedisHandler() {
         ReflectionBasedRedisClient connection = getRedisConnection();
         redisHandler.setRedisClient(connection);
-
-        List<AdditionalInfo> list = getAdditionalInfoList();
-        if (!list.isEmpty()) {
-            AdditionalInfo last = list.get(list.size() - 1);
-            last.getRedisCommandData().forEach(redisHandler::handle);
-        }
     }
 
     /**
--- a/core-tests/e2e-tests/spring/spring-rest-redis/pom.xml
+++ b/core-tests/e2e-tests/spring/spring-rest-redis/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <artifactId>evomaster-e2e-tests-spring</artifactId>
         <groupId>org.evomaster</groupId>
-        <version>5.0.1-SNAPSHOT</version>
+        <version>5.0.3-SNAPSHOT</version>
     </parent>
 
     <artifactId>evomaster-e2e-tests-spring-rest-redis</artifactId>


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/redis/RedisHandler.java
```diff
@@ -62,29 +56,19 @@ public boolean isCalculateHeuristics() {
...
-        }
+        operations.add(info);
     }
 
     public List<RedisCommandEvaluation> getEvaluatedRedisCommands() {
         operations.stream()
             .filter(command -> command.getType().shouldCalculateHeuristic())
             .forEach(redisCommand -> {
-                RedisDistanceWithMetrics distanceWithMetrics = computeDistance(redisCommand, redisClient);
+                RedisDistanceWithMetrics distanceWithMetrics = computeDistance(redisCommand, (RedisClient) redisClient);
```
arcuri82: hi. thx @aszyrej  i think it is best if @jgaleotti  gives first round of comments, and then, after updating, i ll give my pass. but one thing here that caught my eyes is: why turning `RedisClient` into `Object` reference if anyway you do a cast to `(RedisClient)` here?

aszyrej: Hi @arcuri82 . The reason why I'm using Object is just to be aligned with the solutions for Mongo and OpenSearch where they used Object for the clients. It's not necessary to keep it this way and since I have a dedicated class for encapsulating the client logic, I could refactor it and keep that class instead of Object.
I'll proceed with this if you agree.

arcuri82: the reason for using `private Object mongoClient` is that cannot have dependencies to actual Mongo code, and must do access via reflection. if for Redis you encapsulate all the code for reflection in your own `RedisClient` that is different from `io.lettuce.core.RedisClient`, then no point in storing it as an `Object` reference. Btw, for clarity, your own `RedisClient` could be renamed into `ReflectionBasedRedisClient` (or something like that)

aszyrej: Hi @arcuri82, I've renamed that class into ReflectionBasedRedisClient as we discussed, and changed the type in RedisHandler.

client-java/controller/src/main/java/org/evomaster/client/java/controller/redis/ReflectionBasedRedisClient.java
```diff
@@ -8,25 +10,27 @@
...
         try {
             Class<?> redisClientClass = Class.forName("io.lettuce.core.RedisClient");
             Class<?> redisURIClass = Class.forName("io.lettuce.core.RedisURI");
 
             Method createUri = redisURIClass.getMethod("create", String.class);
             Object uri = createUri.invoke(null, "redis://" + host + ":" + port);
 
+            SimpleLogger.debug("Connecting to Redis with PORT: " + port);
+
             Method createClient = redisClientClass.getMethod("create", redisURIClass);
```
jgaleotti: replace string literals for method names with private string constants

client-java/controller/src/main/java/org/evomaster/client/java/controller/redis/RedisHeuristicsCalculator.java
```diff
@@ -175,6 +186,34 @@ private RedisDistanceWithMetrics calculateDistanceForFieldInHash(
...
+    /**
+     * Computes the distance of target field to each field in hash.
+     *
+     * @param targetField Field searched in query.
+     * @param fields Fields in hash.
+     * @return double
+     */
+    private double calculateDistanceForField(String targetField, Set<String> fields) {
+        if (fields.isEmpty()) {
+            return Long.MAX_VALUE;
```
jgaleotti: Why Long.MAX_VALUE is being used instead of Double.MAX_VALUE ?

client-java/controller/src/main/java/org/evomaster/client/java/controller/redis/ReflectionBasedRedisClient.java
aszyrej: Every method has been replaced with a constant.

client-java/controller/src/main/java/org/evomaster/client/java/controller/redis/RedisHeuristicsCalculator.java
aszyrej: We're now using Double.MAX_VALUE

core-tests/e2e-tests/spring/spring-rest-redis/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+            <version>2.0.1.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-redis</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>redis.clients</groupId>
+            <artifactId>jedis</artifactId>
+            <version>5.1.0</version>
```
jgaleotti: Notice that lettuce, mongoDB and OpenSearch version are defined as properties in the main pom.xml

aszyrej: I've moved this version declaration to the root POM.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -346,6 +350,17 @@ public final void initOpenSearchHandler() {
         }
     }
 
+    public final void initRedisHandler() {
+        ReflectionBasedRedisClient connection = getRedisConnection();
+        redisHandler.setRedisClient(connection);
+
+        List<AdditionalInfo> list = getAdditionalInfoList();
```
arcuri82: @jgaleotti  i see this was done for Mongo and OpenSearch as well... but why? can you leave some comments to explain it? i m a bit confused of what is this `handle` on the `init` of the handler based on additional info of currently executed test case

core-tests/e2e-tests/spring/spring-rest-redis/pom.xml
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>evomaster-e2e-tests-spring</artifactId>
+        <groupId>org.evomaster</groupId>
+        <version>5.0.1-SNAPSHOT</version>
```
arcuri82: new relese was made yesterday. make sure to update all these values to latest `5.0.3-SNAPSHOT`

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
jgaleotti: Hi, thanks for pointing this out.
For Redis/OpenSearch, this is not right since there should be no handling for Redis/OpenSearch commands.
For Mongo, this is done for handling the expected classes to parse the document into. In other words, the call is to handle(MongoCollectionSchema) instead of handle(MongoFindCommand)
Could you remove this code @aszyrej  ?

jgaleotti: I also pushed a change to this branch adding a comment in SutController.initMongoHandler()

aszyrej: I've changed the parent version in spring rest redis POM and removed the handle method in initRedisHandle.
@arcuri82 I think we should be ready now to merge.

PR #1418, opened 2026-01-12
fix existence leakage

omursahin:


omursahin:
/**
    * Calls:
    * 1 - (201) PUT:/api/parents/{pid}
    * 2 - (201) PUT:/api/parents/{pid}/children/{cid}
    * 3 - (403) GET:/api/parents/{pid}/children/{cid}
    * 4 - (404) GET:/api/parents/{pid}/children/{cid}
    * 5 - (403) GET:/api/parents/{pid}
    * Found 1 potential fault of type-code 204
    */
    @Test @Timeout(60)
    fun test_4_getOnParentAllowsUnauthorizedAccessToProtectedResource()  {
        
        val res_0: ValidatableResponse = given().accept("*/*")
                .header("Authorization", "BAR") // BAR
                .header("x-EMextraHeader123", "42")
                .put("${baseUrlOfSut}/api/parents/996")
                .then()
                .statusCode(201)
                .assertThat()
                .body(isEmptyOrNullString())
        val location_pid__2 = res_0.extract().header("location")
        assertTrue(isValidURIorEmpty(location_pid__2));
        
        given().accept("*/*")
                .header("Authorization", "BAR") // BAR
                .header("x-EMextraHeader123", "")
                .put(resolveLocation(location_pid__2, baseUrlOfSut + "/api/parents/996/children/877?EMextraParam123=42"))
                .then()
                .statusCode(201)
                .assertThat()
                .body(isEmptyOrNullString())
        
        given().accept("*/*")
                .header("Authorization", "FOO") // FOO
                .header("x-EMextraHeader123", "42")
                .get("${baseUrlOfSut}/api/parents/996/children/877")
                .then()
                .statusCode(403)
                .assertThat()
                .body(isEmptyOrNullString())
        
        // Fault204. Leakage Information Existence of Protected Resource. null.
        given().accept("*/*")
                .header("Authorization", "BAR") // BAR
                .header("x-EMextraHeader123", "")
                .get("${baseUrlOfSut}/api/parents/132/children/845")
                .then()
                .statusCode(404)
                .assertThat()
                .body(isEmptyOrNullString())
        
        given().accept("*/*")
                .header("Authorization", "BAR") // BAR
                .header("x-EMextraHeader123", "42")
                .get("${baseUrlOfSut}/api/parents/132?EMextraParam123=42")
                .then()
                .statusCode(403)
                .assertThat()
                .body(isEmptyOrNullString())
    }

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/RestSecurityOracle.kt
```diff
@@ -200,6 +200,11 @@ object RestSecurityOracle {
                         (actionResults.find { r -> r.sourceLocalId == it.getLocalId() } as RestCallResult)
                     .getStatusCode() == 404
             }
+        val lastStatusCode = (actionResults.last() as RestCallResult).getStatusCode()
```
arcuri82: few issues here:
1) need to add comments to say we do this last check
2) the last call on a parent resource giving 200 is not mandatory (see algorithm description), as in some cases it is not needed (eg if no parent resource). Would need to verify that is on ancestor resource of `path` (likely the top one), and that is done with same auth info of the call that gives 404

arcuri82: even better, instead of checking last call, could check if any call satisfy those properties

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -727,6 +727,44 @@ class SecurityRest {
...
 
+            // Check if the 404 path is a child resource and add parent GET to check access
+            val path404 = get.path
+            val action404 = final.seeMainExecutableActions().last() as RestCallAction
+            val lastAuth = action404.auth
+
+            // Find the parent path from getOperations (all GET operation definitions)
+            val parentGetOperation = getOperations
+                .filter { it.path.isDirectOrPossibleAncestorOf(path404) && !it.path.isEquivalent(path404) }
+                .maxByOrNull { it.path.toString().length } // Get the most specific parent (closest ancestor)
```
arcuri82: shouldn't be `minByOrNull`? as intermediate resources might not exist either.
also I guess should filter for `isLastElementAParameter()`, to make sure it is a specific resource, and not a collection

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -727,6 +727,44 @@ class SecurityRest {
...
+
+            if (parentGetOperation != null) {
+                val parentGetAction = parentGetOperation.copy() as RestCallAction
+                parentGetAction.resetLocalIdRecursively()
+                parentGetAction.forceNewTaints()
+                parentGetAction.auth = lastAuth
+
+                // Bind to the same path params from the 404 action to ensure same IDs
+                parentGetAction.bindBasedOn(
```
arcuri82: binding is only for static info, not for dynamic data (eg ids) coming from API at runtime. it would work when data is created via SQL, but unsure if works in BB with dynamic IDs

core/src/main/kotlin/org/evomaster/core/problem/rest/service/SecurityRest.kt
```diff
@@ -727,6 +727,44 @@ class SecurityRest {
...
+
+                final.addResourceCall(
+                    restCalls = RestResourceCalls(
+                        actions = mutableListOf(parentGetAction),
+                        sqlActions = listOf()
+                    )
+                )
+                final.seeMainExecutableActions()
+                    .filter { it.verb == HttpVerb.PUT || it.verb == HttpVerb.POST }.forEach {
+                        it.saveCreatedResourceLocation = true
```
arcuri82: this would apply to all actions in the individual, regardless of whether their resources are going to be used

PR #1428, opened 2026-01-21
Update ai stats

MohsenTaheriShalmani:
- Added `skipUpdate` logic for responses other than 200 or 400.
- Removed an unnecessary condition specific to the Neural Network classifier, as it can handle raw data in the same way as the other models.
- Introduced `aIResponseClassifierWeaknessThreshold` in `EMConfig`.
- Updated statistics to track the number of 200, 400, and 500 responses.
- Update ai model statistics for the report and analysis.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
@@ -150,14 +150,18 @@ class AIResponseClassifier : AIModel {
         if(enabledLearning) {
 
             // update the observation counters
-            val trueStatusCode = output.getStatusCode()
-            when (trueStatusCode) {
+            when (val trueStatusCode = output.getStatusCode()) {
                 in 200..299 -> stats.observed2xxCount++
                 in 300..399 -> stats.observed3xxCount++
-                in 400..499 -> stats.observed4xxCount++
-                400 -> stats.observed400Count++
+                in 400..499 -> {
+                    stats.observed4xxCount++
+                    if (trueStatusCode == 400) {
+                        stats.observed400Count++
+                    }
+                }
                 in 500..599 -> stats.observed5xxCount++
             }
+
             /**
              * Updates the maximum metrics when the current endpoint classifier
              * outperforms all other endpoint classifiers.


core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -94,7 +123,22 @@ class AIResponseClassifier : AIModel {
         }
 
         if(enabledLearning) {
+
+            // update counters
+            val trueStatusCode = output.getStatusCode()
+            when (trueStatusCode) {
+                200 -> stats.observed200Count++
```
arcuri82: why ignoring all the other 2xx codes?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -94,7 +123,22 @@ class AIResponseClassifier : AIModel {
...
+                400 -> stats.observed400Count++
+                500 -> stats.observed500Count++
+            }
```
arcuri82, line 163: what about all other status codes?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -130,22 +182,25 @@ class AIResponseClassifier : AIModel {
 
         val metrics = estimateMetrics(call.endpoint)
 
+        // update the max metrics
```
arcuri82: why taking `maxOf` here? unclear. clarify in comments

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
```diff
@@ -211,8 +272,17 @@ class AIResponseClassifier : AIModel {
...
         val trueStatusCode = output.getStatusCode() ?: return true
 
-        // optionally skip if server-side error (500)
-        return trueStatusCode == 500 && config.skipAIModelUpdateWhenResponseIs500
+        // skip conditions
+        val skip500 =
+            trueStatusCode == 500 && config.skipAIModelUpdateWhenResponseIs500
+
+        val skipNot200Or400 =
+            trueStatusCode != 200
```
arcuri82: why are focusing on 200 and not 2xx??? it seems wrong

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -357,15 +357,43 @@ class Statistics : SearchListener {
...
         Pair("ai_precision400", "%.4f".format(precision)),
         Pair("ai_recall400", "%.4f".format(recall)),
         Pair("ai_f1Score400", "%.4f".format(f1)),
-        Pair("ai_mcc400", "%.4f".format(mcc))
+        Pair("ai_mcc400", "%.4f".format(mcc)),
+        // Max metrics among all endpoints
+        Pair("max_Accuracy", "%.4f".format(maxAccuracy)),
+        Pair("max_Precision", "%.4f".format(maxPrecision)),
+        Pair("max_Recall", "%.4f".format(maxRecall)),
+        Pair("max_Mcc", "%.4f".format(maxMcc)),
```
arcuri82: maybe all these new fields should start with `ai_` prefix?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1450,13 +1450,23 @@ class EMConfig {
...
+    var aIClassificationMetrics = AIClassificationMetrics.FULL_HISTORY
 
     @Experimental
     @Cfg("Determines whether the AI response classifier skips model updates when the response " +
             "indicates a server-side error with status code 500.")
     var skipAIModelUpdateWhenResponseIs500 = false
 
+    @Experimental
+    @Cfg("Determines whether the AI response classifier skips model updates " +
+            "when the response is not 200 or 400.")
```
arcuri82: focusing on 400 is good. but not 200, should be 2xx. for example, a DELETE would return 204. a POST would typically return a 201. here, you would ignore them

core/src/main/kotlin/org/evomaster/core/problem/rest/service/AIResponseClassifier.kt
MohsenTaheriShalmani: I agree, it should be right after `skipUpdate(input, output)`

PR #1429, opened 2026-01-22
Add ObjectMapper unescaping ASCII control characters when using DTOs

Pgarrett:
ASCII control characters were being escaped by Jackson when using DTOs. Control characters in a JSON payload generate an invalid payload, which was being tested by EvoMaster. As such, we created a custom object mapper that will not escape the ASCII control characters, allowing for incorrect payloads to be tested.

### DIFF
--- a/core/src/test/kotlin/org/evomaster/core/output/formatter/OutputFormatterTest.kt
+++ b/core/src/test/kotlin/org/evomaster/core/output/formatter/OutputFormatterTest.kt
@@ -151,5 +151,19 @@ class OutputFormatterTest {
         assertFalse(isValid)
     }
 
+    @Test
+    fun testInvalidJsonWithAsciiControlCharacter() {
+        val json = " {\"s0\":\"This is a long string\", \"s1\":\"This isa long string\", \"s2\":\"_EM_3149_XYZ_\"} "
+        val isValid = OutputFormatter.JSON_FORMATTER.isValid(json)
+        assertFalse(isValid)
+    }
+
+    @Test
+    fun testValidJsonWithEscapedAsciiControlCharacter() {
+        val json = " {\"s0\":\"This is a long string\", \"s1\":\"This is\\u001Fa long string\", \"s2\":\"_EM_3149_XYZ_\"} "
+        val isValid = OutputFormatter.JSON_FORMATTER.isValid(json)
+        assertTrue(isValid)
+    }
+
 
 }


core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoOutput.kt
```diff
@@ -6,19 +6,28 @@ import org.evomaster.core.output.TestSuiteFileName
 import org.evomaster.core.utils.StringUtils
 import java.nio.file.Path
 
-class JavaDtoOutput: JvmDtoOutput() {
+class JavaDtoOutput(val outputFormat: OutputFormat): JvmDtoOutput() {
```
arcuri82: is any special reason to keep `outputFormat` here as a field? also, in that case, shouldnt we have a `init{}` block to validate it? (ie throw IllegalArgument if not for Java)

core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoOutput.kt
```diff
@@ -81,4 +90,56 @@ class JavaDtoOutput: JvmDtoOutput() {
...
+
+    private fun addMapperContentClass(lines: Lines) {
+        lines.add("public class $customControlCharMapperFactory implements Jackson2ObjectMapperFactory {")
+        lines.addEmpty()
+        lines.indented {
+            lines.add("@Override")
+            lines.add("public ObjectMapper create(Type cls, String charset) {")
+            lines.indented {
+                lines.add("ObjectMapper mapper = new ObjectMapper();")
+                lines.add("mapper.registerModule(new Jdk8Module());")
```
arcuri82: this needs some comment explanations

core/src/main/kotlin/org/evomaster/core/output/dto/KotlinDtoOutput.kt
```diff
@@ -5,17 +5,26 @@ import org.evomaster.core.output.OutputFormat
 import org.evomaster.core.output.TestSuiteFileName
 import java.nio.file.Path
 
-class KotlinDtoOutput: JvmDtoOutput() {
+class KotlinDtoOutput(val outputFormat: OutputFormat): JvmDtoOutput() {
```
arcuri82: input validation?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -812,6 +813,17 @@ class TestSuiteWriter {
                     lines.indented {
                         lines.add(".jsonConfig(JsonConfig.jsonConfig().numberReturnType(JsonPathConfig.NumberReturnType.DOUBLE))")
                         lines.add(".redirect(redirectConfig().followRedirects(false))")
+                        if (config.dtoSupportedForPayload() && containsDtos) {
```
arcuri82: need some comments here to explain why this is done

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -812,6 +813,17 @@ class TestSuiteWriter {
...
+                            lines.indented {
+                                lines.add(".objectMapperConfig(")
```
arcuri82: is this now always applied? so we go through this custom mapper even if there are no issues with the payloads?

core/src/main/kotlin/org/evomaster/core/output/dto/JavaDtoOutput.kt
Pgarrett: No special reason, but DtoOutput is instantiated depending on the actual OutputFormat: https://github.com/WebFuzzing/EvoMaster/blob/2088b978de8bb78ca4494cae4cd49ee059754ab2/core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt#L52-L56

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
Pgarrett: Yes, but it's only being applied for escaping: https://github.com/WebFuzzing/EvoMaster/blob/2088b978de8bb78ca4494cae4cd49ee059754ab2/core/src/main/kotlin/org/evomaster/core/output/dto/KotlinDtoOutput.kt#L69 and for that operation in particular we're just not escaping the ASCII control characters. Also, I wouldn't know how to apply or not the mapper depending on issues or not with the payload, how could we detect that there could be an issue?

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -147,6 +149,21 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
         return ""
     }
 
+    /*
+     * Control characters break JSON and transform it into an invalid payload. If there's any invalid character
+     * then we'll avoid using DTOs and have the payload in the test case be represented by the raw JSON string.
+     */
+    private fun payloadIsValidJson(bodyParam: BodyParam): Boolean {
```
arcuri82, line 154: this function, or parts of it (if taking as input a string and not a BodyParam), should go to `org.evomaster.core.output.JsonUtils` . also, we need unit tests for it, to make sure it works (eg, you can re-use the exact same JSON payload that was failing the E2E test)

arcuri82: actually, looks like there is also a `TestWriterUtils` in the same package having functions for JSON, including a call to `OutputFormatter.JSON_FORMATTER.isValid(json)` where the same type of check is done. maybe better to call that directly

PR #1432, opened 2026-01-31
Update weakness condition

MohsenTaheriShalmani:
Improved the decision-making criteria so that classifier misclassifications do not cause mutations of 2xx.


core/src/test/kotlin/org/evomaster/core/problem/rest/classifier/ModelMetricsWithTimeWindowSmallBufferTest.kt
```diff
@@ -17,55 +17,55 @@ class ModelMetricsWithTimeWindowTest4SmallBuffer {
         ma.updatePerformance(predictedStatusCode = 200, actualStatusCode = 200)
```
arcuri82, line 17: test class names should always end with `*Test`, otherwise they will not be executed in CI

arcuri82 (changes requested):

PR #1463, opened 2026-03-07
Update EmTest

MohsenTaheriShalmani:
Update thresholds and AIClassificationEMTestBase based on the last set of experiments.

### DIFF
--- a/core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/aiclassification/AIClassificationEMTestBase.kt
+++ b/core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/aiclassification/AIClassificationEMTestBase.kt
@@ -39,31 +39,53 @@ abstract class AIClassificationEMTestBase : SpringTestBase(){
         return ei
     }
 
+    private fun isWeakClassifier(
+        model: AIResponseClassifier,
+        action: RestCallAction,
+        weaknessThreshold: Double
+    ): Boolean {
+
+        val metrics = model.estimateMetrics(action.endpoint)
+
+        return metrics.precision400 <= weaknessThreshold
+                || metrics.sensitivity400 <= weaknessThreshold
+                || metrics.specificity <= weaknessThreshold
+                || metrics.npv <= weaknessThreshold
+    }
+
     protected fun verifyModel(
         injector: Injector,
         ok2xx: List<RestCallAction>,
         fail400: List<RestCallAction>,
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1563,7 +1563,7 @@ class EMConfig {
     @PercentageAsProbability(false)
     @Cfg("If using THRESHOLD for AI Classification Repair, specify its value." +
             " All classifications with probability equal or above such threshold value will be accepted.")
-    var classificationRepairThreshold = 0.8
+    var classificationRepairThreshold = 0.5
```
arcuri82, line 1566: are these changes based on latest experiments?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1602,7 +1602,7 @@ class EMConfig {
     @Experimental
     @Cfg("Minimum confidence threshold required for the AI response classifier to decide" +
             "whether to send a request as-is or attempt a repair.")
-    var aIResponseClassifierWeaknessThreshold = 0.4
+    var aIResponseClassifierWeaknessThreshold = 0.8
```
arcuri82, line 1605: are these changes based on latest experiments?

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/aiclassification/AIClassificationEMTestBase.kt
```diff
@@ -39,31 +39,53 @@ abstract class AIClassificationEMTestBase : SpringTestBase(){
...
+        weaknessThreshold: Double = injector.getInstance(EMConfig::class.java).aIResponseClassifierWeaknessThreshold
     ) {
 
         val model = injector.getInstance(AIResponseClassifier::class.java)
         model.disableLearning() // no side-effects
 
+
         for(ok in ok2xx){
+
+            if (isWeakClassifier(model, ok, weaknessThreshold)) continue
```
arcuri82, line 70: i m unsure about this... we will need to discuss. for example, if the model is always weak, would it mean this test will always pass? that would be against the point of having a E2E. or is guassian not able to reliably solve these simples APIs in these E2Es?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: ah... i see you wrote it in the description of this PR... :)

PR #1471, opened 2026-03-12
http oracle partial update put

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,6 +247,122 @@ object HttpSemanticsOracle {
         return resAfter.getStatusCode() != 404
     }
 
+    /**
+     * Checks the PUT full-replacement oracle:
+     *
+     *   PUT /path  -> 2xx  (successful update/create with body B)
+     *   GET /path  -> 2xx  (must return exactly the fields that were PUT)
+     *
+     * Returns true if any field sent in the PUT body has a different value
```
arcuri82: this is partial. you need also to check that GET is not returning fields not used in the PUT. or, if those are returned, if they are null

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,6 +247,122 @@ object HttpSemanticsOracle {
...
+
+        val bodyParam = put.parameters.find { it is BodyParam } as BodyParam?
+        if (bodyParam != null && !bodyParam.isJson() && !bodyParam.isXml() && !bodyParam.isForm()) {
+            return false
+        }
+
+        val putBody = extractRequestBody(put)
+        val getBody = resGet.getBody()
+
+        if (putBody.isNullOrEmpty() || getBody.isNullOrEmpty()) return false
```
arcuri82: if `putBody` is empty, but `getBody` is not, then we got a bug

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
arcuri82: also, need to check the differences in what declared in the schema, eg, the resource might have fields that are not supposed to be changed via a PUT (eg timestamps or ids)

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,6 +247,122 @@ object HttpSemanticsOracle {
...
+
+        val fieldNames = extractModifiedFieldNames(put)
```
arcuri82: here it should rather check the field names of what declared in the schema of the PUT. i have added some more discussion and examples to `2026/httporacles/notes.txt`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -372,5 +373,42 @@ class HttpSemanticsService {
...
     }
 
+    /**
+     * HTTP_PARTIAL_UPDATE_PUT oracle: PUT makes a full replacement, not a partial update.
+     * If only some fields should be modified, PATCH must be used instead.
+     *
+     * Sequence checked:
+     *   PUT /X  body=B  ->  2xx
+     *   GET /X          ->  response body must match exactly B
+     *                       (no field from a previous state should bleed through)
```
arcuri82, line 383: see updated discussion at `2026/httporacles/notes.txt`

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/httporacle/partialupdateput/HttpPartialUpdatePutEMTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        fun init() {
+            initClass(HttpPartialUpdatePutController())
+        }
+    }
+
+
+    @Test
+    fun testRunEM() {
+
+        runTestHandlingFlakyAndCompilation(
```
arcuri82, line 27: not sure if best in a E2E or an integration test, but check to verify the example discussed in `2026/httporacles/notes.txt`, especially making sure that K does not lead to flag a fault (ie false positive)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
omursahin: I also added this statement: if putBody is not empty, but getBody is. What do you think about that? Is that correct or lead to false positives?

arcuri82: @omursahin  hi, yes, should be fine, i guess shouldn't harm. but, anyway, those cases should still be handled when looking at fields one by one, isn't it?

omursahin: For this, 
`if (putBody.isNullOrEmpty() && !getBody.isNullOrEmpty()) return true
`
yes, you are right. I am updating it now.

But I guess we don't need to check fields for this statement:
`if (!putBody.isNullOrEmpty() && getBody.isNullOrEmpty()) return true
`
Because if we send any value with put, we expect that it should be returned in the get body, right?

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/httporacle/partialupdateput/HttpPartialUpdatePutEMTest.kt
omursahin: I checked it in `\core\src\test\kotlin\org\evomaster\core\problem\rest\oracle\HttpSemanticsOracleTest.kt` but if you want, I can write an integration test too.

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
arcuri82: yep, if send with PUT, we expect in GET.

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/httporacle/partialupdateput/HttpPartialUpdatePutEMTest.kt
arcuri82: as long as it checked, then it is fine :)

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
+        individual: RestIndividual,
+        actionResults: List<ActionResult>,
+        schema: RestSchema? = null
+    ): Boolean {
+
+        val (put, get, resPut, resGet) = findPutGetPair(individual, actionResults) ?: return false
+
+        if (!StatusGroup.G_2xx.isInGroup(resPut.getStatusCode())) return false
+        // if put returned 2xx but entity does not exist afterwards
+        if (resGet.getStatusCode() == 404) return true
```
arcuri82, line 284: that's a good point... can you update `2026/httporacles/notes.txt` mentioning it? it will be helpful in few weeks/months when we are going to use those notes as starting point for a paper :)

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
+        if (!StatusGroup.G_2xx.isInGroup(resGet.getStatusCode())) return false
+
+        val bodyParam = put.parameters.find { it is BodyParam } as BodyParam?
+        if (bodyParam != null && !bodyParam.isJson() && !bodyParam.isXml() && !bodyParam.isForm()) {
```
arcuri82, line 290: add a comment, specifying that for now we only deal with JSON/XML/FORM. also, clarify what would happen when `bodyParam` is null

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
+            return false
+        }
+
+        val putBody = extractRequestBody(put)
+        val getBody = resGet.getBody()
+
+        // PUT sent content but GET body is empty -> sent fields definitely missing
+        if (!putBody.isNullOrEmpty() && getBody.isNullOrEmpty()) return true
+        if (getBody.isNullOrEmpty()) return false
```
arcuri82: refactor for clarity, eg, something like:

```
if (getBody.isNullOrEmpty()) {
     return !putBody.isNullOrEmpty()
}
```

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
+            ?: OutputFormatter.XML_FORMATTER.readFields(getBody, fieldNames)
+    }
+
+    /**
+     * A wiped field is present only if the GET response returns a non-null, non-empty value.
+     */
+    private fun isWipedFieldStillPresent(value: String?): Boolean {
+        if (value == null) return false
+        if (value.isEmpty()) return false
+        if (value == "null") return false
```
arcuri82: i m a bit unsure about this `"null"` check. are you sure that you can properly distinguish between:

`{"x": null}`  (empty value)

and

`{"x": "null"}` (value is present, ie., a string with 4 characters)

make sure to have at least 1 test case to check this

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
      * These are the fields that the client attempted to modify.
      */
     private fun extractModifiedFieldNames(modify: RestCallAction): Set<String> {
 
+        val objectGene = extractBodyObjectGene(modify) ?: return emptySet()
+
+        return objectGene.fields.map { it.name }.toSet()
+    }
+
+    // Extract only the field names that were actually sent in the request body.
```
arcuri82: for method comments, remember to use JavaDoc styel `/** */` and not `//`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -247,24 +251,224 @@ object HttpSemanticsOracle {
...
+    }
+
+    private fun extractBodyObjectGene(modify: RestCallAction): ObjectGene? {
         val bodyParam = modify.parameters.find { it is BodyParam } as BodyParam?
-            ?: return emptySet()
+            ?: return null
 
         val gene = bodyParam.primaryGene()
-        val objectGene = gene.getWrappedGene(ObjectGene::class.java) as ObjectGene?
+        return gene.getWrappedGene(ObjectGene::class.java) as ObjectGene?
```
arcuri82, line 472: keep in mind that a body could be optional (ie, not marked as required), and so inside an `OptionalGene`. if this latter is off (ie, not active), then nothing is sent. should check this

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -365,6 +569,56 @@ object HttpSemanticsOracle {
         return false
     }
 
+    /**
+     * Returns the property names from the PUT request body schema in the OpenAPI spec.
+     * Used as a fallback to determine writable fields when no BodyParam is present on the action.
+     */
+    internal fun extractPutRequestSchemaFields(
```
arcuri82: this feels like a general function... could be extracted in a utility class, as could be re-used in other places as well

arcuri82: maybe in `SchemaUtils`?

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -365,6 +569,56 @@ object HttpSemanticsOracle {
...
+        val pathItem = openAPI.paths?.get(put.path.toString()) ?: return emptySet()
+        val op = pathItem.put ?: return emptySet()
+        val requestBody = op.requestBody ?: return emptySet()
+        val mediaType = requestBody.content?.values?.firstOrNull() ?: return emptySet()
+        val rawSchema = mediaType.schema ?: return emptySet()
+        val resolved = rawSchema.`$ref`?.let {
+            SchemaUtils.getReferenceSchema(schema, schema.main, it, mutableListOf())
+        } ?: rawSchema
+
+        return resolved.properties?.keys?.toSet() ?: emptySet()
```
arcuri82: schema analysis is very TRICKY... as you could have `$ref` at each point in the tree... and those override any sibling node :(   need more checks and unit tests for this

`RestCallAction` is created based on content of schema, but taint analysis could had been involved... so still make sense to look at original schema for these checks

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -365,6 +569,56 @@ object HttpSemanticsOracle {
...
+    }
+
+    /**
+     * Returns the property names from the GET 2xx response schema in the OpenAPI spec.
+     * Empty set if unresolvable, which makes callers skip wiped-field checks.
+     */
+    internal fun extractGetResponseSchemaFields(
+        schema: RestSchema,
+        get: RestCallAction
```
arcuri82: as the other function, could be refactored into `SchemaUtils`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -365,6 +569,56 @@ object HttpSemanticsOracle {
...
+    ): Set<String> {
+
+        val openAPI = schema.main.schemaParsed
+        val pathItem = openAPI.paths?.get(get.path.toString()) ?: return emptySet()
+        val op = pathItem.get ?: return emptySet()
+
+        // pick the first 2xx response, falling back to "default"
+        val response = op.responses?.entries
```
arcuri82: as for other function, need to check for the presence of `$ref` at each point in the tree where it can be present

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
omursahin: I removed this function and started to use just` "isNullOrEmpty"`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
omursahin: I hope I solved the problem and added test cases into `core/src/test/kotlin/org/evomaster/core/problem/rest/schema/SchemaUtilsTest.kt`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -113,6 +113,7 @@ class HttpSemanticsService : TimeBoxedPhase{
 
         if(hasPhaseTimedOut()) return
         sideEffectsOfFailedModification()
+        partialUpdatePut()
```
arcuri82, line 118: add a call to ` if(hasPhaseTimedOut()) return` before this

arcuri82 (changes requested):

arcuri82 (approved):

PR #1474, opened 2026-03-17
Regex support for missing predefined character classes and negated character classes

lmasroca:
Added Java/JS support for negated character classes
Added Java support for missing predefined character classes escapes (\v, \H, etc.) and POSIX character classes (\p{Lower}, etc.)


core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -234,7 +251,7 @@ AtomEscape
 
 fragment CharacterClassEscape
  //one of d D s S w W
```
jgaleotti: The comment does not match the production

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,96 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
+
         ranges.forEach {
             if(it.first.code > it.second.code){
                 LoggingUtil.uniqueWarn(log, "Issue with Regex range, where '${it.first}' is greater than '${it.second}'")
             }
         }
     }
 
-    var value : Char = ranges[0].first
+    var value : Char = internalRanges[0].first
```
jgaleotti: add javadoc of this function

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,96 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
+        }
+
+        internalRanges = newInternalRanges
+    }
 
-    /**
-     * As inputs might be unsorted, we make sure first <= second
-     */
-    val ranges = ranges.map { Pair(min(it.first.code,it.second.code).toChar(), max(it.first.code, it.second.code).toChar()) }
+    private fun remove(toRemove: Pair<Char, Char>) {
```
jgaleotti: add javadoc of this function

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,96 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
             if(it.first.code > it.second.code){
                 LoggingUtil.uniqueWarn(log, "Issue with Regex range, where '${it.first}' is greater than '${it.second}'")
             }
         }
     }
 
-    var value : Char = ranges[0].first
+    var value : Char = internalRanges[0].first
+
+    private fun add(toAdd: Pair<Char, Char>) {
```
jgaleotti: replace Pair<Char,Char> with an internal class for CharacterRange

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -14,6 +15,10 @@ class Randomness {
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(Randomness::class.java)
+
+        private fun stringToListOfCharPairs(s: String) : List<Pair<Char, Char>> {
```
jgaleotti: use a class to replace Pair<Char,Char>

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -34,19 +39,47 @@ class Randomness {
...
     }
 
-    private val digitSet = "0123456789"
-    private val asciiLetterSet = "abcdefghilmnopqrstuvzjkwxyABCDEFGHILMNOPQRSTUVZJKWXY"
-    private val norwegianLetterSet = "æøåÆØÅ"
-
-    private val wordSet = "_$digitSet$asciiLetterSet"
-    private val spaceSet = " \t\r\n"
-    private val punctuationSet = "!@#$%^&*()[]{}<>:;|"
+    private val digitS = listOf('0' to '9')
```
jgaleotti: digitS is digitSet?

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -34,19 +39,47 @@ class Randomness {
...
+    private val asciiLetterS = listOf('a' to 'z', 'A' to 'Z')
```
jgaleotti: what is the goal of having this variables with suffix "S" instead of "Set"?

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -34,19 +39,47 @@ class Randomness {
...
+    private val wordS = listOf('_' to '_') + asciiLetterS + digitS
+    private val spaceS = stringToListOfCharPairs(" \t\r\n\u000C\u000b")
+    private val horizontalSpaceS = listOf(0x2000.toChar() to 0x200a.toChar()) +
+            stringToListOfCharPairs(" \t\u00A0\u1680\u180e\u202f\u205f\u3000")
+    private val verticalSpaceS = stringToListOfCharPairs("\n\u000B\u000C\r\u0085\u2028\u2029")
+    private val punctuationS = stringToListOfCharPairs("""!"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~""")
+
+    private val digitCC = CharacterRangeRxGene(false, digitS)
```
jgaleotti: what is the meaning of suffix "CC"?

jgaleotti (changes requested):

core/src/test/kotlin/org/evomaster/core/search/gene/GeneSamplerForTests.kt
```diff
@@ -461,7 +461,7 @@ object GeneSamplerForTests {
     fun sampleCharacterRangeRxGene(rand: Randomness): CharacterRangeRxGene {
         return CharacterRangeRxGene(
                 negated = false, // TODO update once fixed
```
lmasroca: I suppose this TODO relates to negated character classes, which should now be working properly. However, I am not sure as to what should be updated here.

arcuri82: hi, maybe something like:
`negated = rand.nextBoolean(),`

jgaleotti: sounds reasonable

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Yes i meant to change it back to Set.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Forgot to change them back from S to Set, used S temporarily as at some point both S and Set existed at the same time. Changed S to Set and renamed the two remaining old sets (word and ascii) to "simpleXSet" as they are used elsewhere.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: CC stood for Character Class, changed the suffix to CharClass for more clarity.

core/src/test/kotlin/org/evomaster/core/search/gene/GeneSamplerForTests.kt
lmasroca: Added that change.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: This change was separated into the second commit as the constructor for CharacterRangeRxGene had to be refactored.

lmasroca:
Looking into this failing check, the test generates a SqlCompositeGene which contains a ChoiceGene which contains an ArrayGene which starts empty but its template contains SqlForeignKeyGene. This generates no problems as long as the array is empty, but as soon as the array gets a random element everything stops being globally valid (in this case the new element, SqlForeignKeyGene is not globally valid as none of its parents in the chain are an SqlAction). The test filters globally invalid genes before randomizing but not after.

This happens on GeneRandomizedTest during verifyCopyValueFrom. We could add a globally valid check after randomization the same way it is done with printable checks, this makes the test pass. However, I am not sure if there should be checks in place to make this chain invalid even before randomizing instead.

```kotlin
    private fun verifyCopyValueFrom(
        mutable: List<Gene>,
        rand: Randomness
    ) {
        val printable = mutable.filter { it.isGloballyValid() && it.isPrintable() }

        printable.forEach { root ->

            val x = root.copy()
            val sx = x.getValueAsRawString()

            val y = x.copy().apply { randomize(rand, true) }
            if(!y.isPrintable()){ // we could add a globally valid check here the same way we filter before
                return@forEach
            }
            val sy = y.getValueAsRawString()
```

jgaleotti (approved):

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/main/resources/static/openapi-dto-reflective-assert.yaml
```diff
@@ -97,6 +97,7 @@ paths:
                 - aDouble
                 - aFloat
                 - aBoolean
+                - aNullableString
```
arcuri82, line 100: why this change? 
@Pgarrett  you wrote this file, isn't it? what is your opinion on this change?

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -125,9 +125,26 @@ fragment CharacterEscape
  | HexEscapeSequence
  | UnicodeEscapeSequence
  | OctalEscapeSequence
+ | 'p' BRACE_open PosixCharacterClassLabel BRACE_close
```
arcuri82: add comment that this is specific to Java... possible also add link where \p is explained

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -125,9 +125,26 @@ fragment CharacterEscape
...
  //| IdentityEscape
  ;
 
+fragment PosixCharacterClassLabel
```
arcuri82, line 135: provide reference link from where those values were taken from. recall the comments at the top of this file

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -233,8 +250,8 @@ AtomEscape
  ;
 
 fragment CharacterClassEscape
- //one of d D s S w W
- : [dDsSwW]
+ //one of d D s S w W v V h H
```
arcuri82, line 256: as `v V h H` are non-standard, and specific to Java, add a comment to explain them here

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
 
+    init {
         if(ranges.isEmpty()){
             throw IllegalArgumentException("No defined ranges")
         }
 
+        // this limits the character class complements to 0xffff instead of allowing up to 0x10ffff, but values over
+        // 0xffff are not permitted on Char as they need 2 Chars to be represented; to allow this, we would need to
+        // use String or Int in every possible step as methods which return a single Char cannot return these characters
+        if(negated) internalRanges.add(CharacterRange(Character.MIN_VALUE,Character.MAX_VALUE))
```
arcuri82: always use {} for blocks, even if single statement inside

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
 import org.evomaster.core.search.service.mutator.genemutation.AdditionalGeneMutationInfo
 import org.evomaster.core.search.service.mutator.genemutation.SubsetGeneMutationSelectionStrategy
 import org.slf4j.LoggerFactory
-import kotlin.math.max
-import kotlin.math.min
 
+data class CharacterRange(val start: Char, val end: Char){
+    val size: Int
+        get() = end.code - start.code + 1
```
arcuri82: is this assuming that `end` is greater than `start`? if so, throw an `IllegalArgumentException` in a `init{}` block if constraint not satisfied

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
         if(ranges.isEmpty()){
             throw IllegalArgumentException("No defined ranges")
         }
 
+        // this limits the character class complements to 0xffff instead of allowing up to 0x10ffff, but values over
+        // 0xffff are not permitted on Char as they need 2 Chars to be represented; to allow this, we would need to
+        // use String or Int in every possible step as methods which return a single Char cannot return these characters
+        if(negated) internalRanges.add(CharacterRange(Character.MIN_VALUE,Character.MAX_VALUE))
+        for (range in ranges) {
+            val max = maxOf(range.start, range.end)
```
arcuri82: wouldn't this always be `end`? clarify the invariants on the `CharacterRange` class

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
+                remove(CharacterRange(min, max))
+            } else {
+                add(CharacterRange(min, max))
+            }
+        }
+
         ranges.forEach {
-            if(it.first.code > it.second.code){
-                LoggingUtil.uniqueWarn(log, "Issue with Regex range, where '${it.first}' is greater than '${it.second}'")
+            if(it.start.code > it.end.code){
```
arcuri82: if this is not supposed to ever happen, ie if it happens it is a bug, then here should rather throw an exception

arcuri82: i see that was the old behavior... but now that we create `CharacterRange` classes, such invariant should be handled correctly. eg a range like `z-a` could be transformed in the constructor of `CharacterRange` as `a-z` (under the assumption the 2 are functional equivalent)

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
+    val size: Int
+        get() = end.code - start.code + 1
+    operator fun contains(char: Char): Boolean = char in start..end
+}
 
 class CharacterRangeRxGene(
-        val negated: Boolean,
-        ranges: List<Pair<Char,Char>>
+    val negated: Boolean,
+    val ranges: List<CharacterRange>
```
arcuri82: if `ranges` is not used outside of this class, and internally we use `internalRanges`, why making it a `val`?

arcuri82: if it is just to handle `copyContent`, we could have optimization of a private secondary constructor that takes as input `internalRanges` (which would avoid re-computing the negation at each copy)

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
     companion object{
         private val log = LoggerFactory.getLogger(CharacterRangeRxGene::class.java)
     }
 
-    init {
-        //TODO this will need to be supported
-        if(negated){
-            throw IllegalArgumentException("Negated ranges are not supported yet")
-        }
+    private var internalRanges = mutableListOf<CharacterRange>()
```
arcuri82: add javadoc comment stating these ara valid ranges... if original was to be negated, this contains the transformation of original. otherwise, it is the same

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -11,50 +11,113 @@ import org.evomaster.core.search.service.mutator.MutationWeightControl
...
 
     override fun checkForLocallyValidIgnoringChildren() : Boolean{
-        //TODO negated
-        return ranges.any { value.code >= it.first.code && value.code <= it.second.code }
+        return internalRanges.any { value in it }
     }
 
     override fun isMutable(): Boolean {
-        return ranges.size > 1 || ranges[0].let { it.first != it.second }
+        return internalRanges.size > 1 || internalRanges[0].let { it.start != it.end }
```
arcuri82: maybe the check for `it.start != it.end` should replaced with an equivalent check on size greater than 1?

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -65,48 +128,56 @@ class CharacterRangeRxGene(
     }
 
     override fun setValueWithRawString(value: String) {
```
arcuri82, line 47: this is a deprecated method, that shouldn't be use directly any longer. shouldn't modify it (unless striclty necessary), as it will be removed eventually. see comments in super method. however, repair based on locally valid shouldn't be done here (it is handled in super classes)

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -65,48 +128,56 @@ class CharacterRangeRxGene(
...
+            if (sampledValue < currentRangeMaxValue) {
+                val codePoint = r.start.code + (sampledValue - currentRangeMinValue)
+                // is it necessary to log this?
+                log.trace("using Int {} as character selector for character class, resulting in code point: {}, which is: {}", sampledValue, codePoint, codePoint.toChar())
+                value = codePoint.toChar()
+                return
+            }
+            currentRangeMinValue = currentRangeMaxValue
+        }
+        throw IllegalArgumentException("No defined ranges")
```
arcuri82: this should never happen, if it does it is a bug.
so, maybe replace with an `assert(false)`, with a comment explaining it

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -2,7 +2,8 @@ package org.evomaster.core.search.service
 
 import com.google.inject.Inject
 import org.evomaster.core.EMConfig
-import org.evomaster.core.utils.NumberCalculationUtil
+import org.evomaster.core.search.gene.regex.CharacterRangeRxGene
+import org.evomaster.core.search.gene.regex.CharacterRange
```
arcuri82: if `CharacterRange` is used in different places outside gene, then we should move it package location, maybe under `src/main/kotlin/org/evomaster/core/utils`

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -14,6 +15,10 @@ class Randomness {
 
     companion object {
         private val log: Logger = LoggerFactory.getLogger(Randomness::class.java)
+
+        private fun stringToListOfCharPairs(s: String) : List<CharacterRange> {
```
arcuri82: rather `stringToListOfCharacterRanges` ?

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -34,21 +39,49 @@ class Randomness {
...
-
-    private val nonWordSet = allSet.replace(wordSet,"")
-    private val nonDigitSet = allSet.replace(digitSet, "")
-    private val nonSpaceSet = allSet.replace(spaceSet, "")
-
-    private val wordChars = wordSet.map { it.toInt() }.sorted()
+    private val digitSet = listOf(CharacterRange('0', '9'))
+    private val asciiLetterSet = listOf(CharacterRange('a', 'z'), CharacterRange('A', 'Z'))
+    private val wordSet = listOf(CharacterRange('_', '_')) + asciiLetterSet + digitSet
+    private val spaceSet = stringToListOfCharPairs(" \t\r\n\u000C\u000b")
```
arcuri82: explain in a comment what `\u000C\u000b` are...
also notice that, in some cases, we explicitely skipped some special characters because escaping in the generating test files is bit buggy (need fixing... but never had time)

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -34,21 +39,49 @@ class Randomness {
...
+    private val horizontalSpaceSet = listOf(CharacterRange(0x2000.toChar(), 0x200a.toChar())) +
+            stringToListOfCharPairs(" \t\u00A0\u1680\u180e\u202f\u205f\u3000")
+    private val verticalSpaceSet = stringToListOfCharPairs("\n\u000B\u000C\r\u0085\u2028\u2029")
+    private val punctuationSet = stringToListOfCharPairs("""!"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~""")
+
+    private val digitCharClass = CharacterRangeRxGene(false, digitSet)
```
arcuri82: should not have references to a gene (ie `CharacterRangeRxGene`) here in this package, as it can have architectural side-effects in the long run... whatever logic needs to be re-used, should be refactored and extracted outside of gene

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -276,42 +306,82 @@ class Randomness {
         return set[random.nextInt(set.length)]
     }
 
+    fun nextFromCharClass(cc: CharacterRangeRxGene) : Char{
```
arcuri82: maybe you could introduce a new `MultiCharacterRange` and move out some of business logic of `CharacterRangeRxGene` into it (which then would use `MultiCharacterRange` internally)

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
```diff
@@ -276,42 +306,82 @@ class Randomness {
...
+        cc.randomize(this, false)
+        val k = cc.value
+        // is it necessary to log this?
+        log.trace("nextFromCharClass(): {}", k)
```
arcuri82: yes... we have tests for determinism that check those trace logs. eg look at `runAndCheckDeterminism()`

arcuri82 (changes requested):

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/main/resources/static/openapi-dto-reflective-assert.yaml
Pgarrett: Yes I did, it seemed one of the fields was not being generated in the test and thus causing the test to fail in the reflective assertions

Pgarrett: No concern on my side

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Before these changes there were no references to genes here, but character classes values were sampled from here. The possible values came from strings defined on this file, which provided some basic functionality but missed several characters from predefined character classes. Trying to get these predefined character classes working with these missing characters I implemented them by changing code here. I see now how this could be an issue.

So to fix this, the predefined character classes definitions should be in CharacterClassEscapeRxGene, then with each randomize instead of using Randomness.nextX() we just call CharacterRangeRxGene.randomize(...) by making each CharacterClassEscapeRxGene hold the CharacterRangeRxGene which represents it? Then restore the previous method of getting Randomness.nextWordChar(), etc. (regarding previous comment about skipping some special characters as escaping in the generating test files)

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
lmasroca: \p is allowed in javascript also but only when certain flags are enabled. There is no flag support for Java/JS yet.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Yes it was old behavior, however i just looked into this and it seems z-a is illegal in both Java and JavaScript. There is a test in `GeneRegexJavaVisitorTest` (`testDecreasingRange`) that expects to be able to sample from a decreasing range. There are two commented lines that call `checkSameAsJava` with decreasing ranges and state that it is not valid in java but then calls `checkCanSample` to check if we can sample from a decreasing range.

Should this test and behavior remain regardless? or should we remove this test and instead of correcting the ordering of `CharacterRanges` (either silently or with a warning) just check the ordering in init and if it is decreasing throw en exception?

At the moment (after the last commits) this warning was removed as now `CharacterRanges` (silently) reorders start and end accordingly. On the master branch this the behavior is log this warning and then the ranges were reordered.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Reversed all changes to this function.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Replaced with assert(false), note that this code was moved to MultiCharacterRange.sample method.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Renamed, this code was moved to `CharacterClassEscapeRxGene` along with the predefined character classes definitions.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Added javadoc, however this variable was later renamed to `validRanges`, the javadoc might be a bit redundant now.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Yes this was to handle `copyContent`, this optimization is implemented now. However this optimization was done on a new main private constructor and the old main constructor is now a secondary public constructor (the only public constructor) which then calls the main one after processing. It was done this way because (to my knowledge), all secondary constructors have to call the main constructor at some point. I could not find a simple, clean way to accomplish this the other way around.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Moved the range/negation construction logic (previous `CharacterRangeRxGene` `init`, `add` and `remove` methods) and sampling logic from `CharacterRangeRxGene` to this new `MultiCharacterRange` class.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: This code in particular was removed as now all the randomness modifications were reversed and the logic is now implemented between `MultiCharacterRange`, `CharacterRangeRxGene` and `CharacterClassEscapeRxGene`.

There is however a log regarding character sampling from character classes on `MultiCharacterRange.sample`. Is that log message correct?

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Now it should be, changed this.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Yes, now there is an invoke before constructor which orders them. If there was an exception in init block, a test in `GeneRegexJavaVisitorTest` would fail. In master branch we warn about `start>end` but reorder. Currently we reorder, however this behavior is illegal on both Java and JavaScript as neither of them accept decreasing ranges.

core/src/main/kotlin/org/evomaster/core/search/service/Randomness.kt
lmasroca: Comment added, unsure about which special characters should be skipped, is there a way to figure out which characters cause issues and where? Now, with proper character class negation, the nonX character classes (i.e.: \D, \W, etc.) can generate almost any char it the range 0000-FFFF. This code was moved into CharacterClassEscapeRxGene.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: Removed the mentioned test, exception thrown when a decreasing range appears.

arcuri82 (approved):

PR #1482, opened 2026-03-20
Filter blank failedWhere Queries from mutator

agusaldasoro:
This pull request makes a focused improvement to the mutation logic for handling failed SQL WHERE queries in the `ApiWsStructureMutator` class. The main change ensures that only non-blank failed WHERE queries are processed, which helps prevent unnecessary or invalid operations during mutation. _**We were sending empty strings as a valid query.**_

- Mutation logic improvement:
  * In `ApiWsStructureMutator`, the list of failed WHERE queries is now filtered to exclude blank entries before further processing. This helps avoid issues related to empty or invalid SQL queries during mutation.


core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -277,6 +277,7 @@ abstract class ApiWsStructureMutator : StructureMutator() {
         val oldSqlActions = mutableListOf<EnvironmentAction>().plus(ind.seeInitializingActions())
 
         val failedWhereQueries = evaluatedIndividual.fitness.getViewOfAggregatedFailedWhereQueries()
+            .filter { it.isNotBlank() }
```
jgaleotti: Would that be too complicate to add a test case for checking that no blank queries are sent?

agusaldasoro: Done!

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
arcuri82: how is it possible that a query is empty?  or is it possibly a bug where those queries are collected? see below:

Image

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
agusaldasoro: Hey! You are right, I moved the fix to that function

core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
```diff
@@ -152,14 +163,28 @@ class FitnessValue(
...
-            databaseExecutions.values.flatMap { a -> a.executionInfo }.map{ b -> b.sqlCommand }
+            databaseExecutions.values
+                // Only executions that actually had failing WHERE clauses are relevant;
+                // collecting from all executions would include unrelated INSERT/UPDATE/DELETE/etc.
+                .filter { it.failedWhere.isNotEmpty() }
+                .flatMap { it.executionInfo }
+                .map { it.sqlCommand }
+                // JSQLParser >= 4.9 may produce empty-string SQL commands (it used to throw, now
+                // parses "" to null). Guard here at the source rather than at every call site.
+                .filter { it.isNotBlank() }
```
arcuri82, line 173: was this definition of `aggregatedFailedWhereQueries` correct? i mean, also the previous version. inside `DatabaseExecution` we keep track of all SQL commands done in a test. some have `failedWhere`, some do not. but here it seems like you mark everything under `executionInfo` as `failedWhere`, isn't it?  this would work if and only if all queries had `failedWhere`, isn't it?

arcuri82: just to be on safe side, can you add some tests to check this scenario?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/FitnessValue.kt
agusaldasoro: Shouldn't that be covered by line 169 `.filter { it.failedWhere.isNotEmpty() }`?

I added a test also to check both conditions failed where present and absent

arcuri82: the class `DatabaseExecution` is old, where `executionInfo` was added later
```
/**
 * When a test case is executed, and the SUT does access a SQL database,
 * keep track of which SQL commands were executed.
 *
 * In particular, we keep track of which tables and columns were involved.
 *
 * For the Maps:
 * key -> table name
 * value -> column names, where "*" means all
 *
 *
 * This class MUST be immutable
 */
class DatabaseExecution(
        val queriedData: Map<TableId, Set<String>>,
        val updatedData: Map<TableId, Set<String>>,
        val insertedData: Map<TableId, Set<String>>,
        val failedWhere: Map<TableId, Set<String>>,
        val deletedData: List<TableId>,
        val numberOfSqlCommands: Int,
        val sqlParseFailureCount: Int,
        val executionInfo: List<SqlExecutionInfo>
)
``` 

unless I m mistaken, `executionInfo` includes info for all SQL commands, even the ones that had no failed WHERE.

agusaldasoro: You are right, I used table-name substring matching as a heuristic to keep only commands that reference a table with a failed WHERE clause.

arcuri82:
@jgaleotti @agusaldasoro   sorry, due to exams and sickness, don't think i ll have any time to look at this till next week

arcuri82 (approved):

PR #1494, opened 2026-04-07
Support multi-column foreign keys in SQL

jgaleotti:
support for multi-column foreign keys in SQL.


core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
```diff
@@ -25,26 +25,41 @@ import org.evomaster.core.sql.schema.TableId
  * known beforehand, as primary keys could be dynamically generated by the database.
  */
 class SqlForeignKeyGene(
+        /**
+         * The name of this column in the table.
+         */
         sourceColumn: String,
+        /**
+         * The position in the list of SQL insertion actions
```
arcuri82: `position`??? if that is indeed the case, it seems very brittle, as SQL actions can be inserted/deleted during mutation

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
```diff
@@ -25,26 +25,41 @@ import org.evomaster.core.sql.schema.TableId
...
          * Otherwise, it should be equal to the uniqueId of
          * a previous SqlPrimaryKey
          */
-        var uniqueIdOfPrimaryKey: Long = -1
+        var uniqueIdOfPrimaryKey: Long = -1,
+        /**
+         * names of columns that form a composite foreign key
+         * together with this one.
+         */
+        val otherSourceColumnsInCompositeFK: List<String> = emptyList()
```
arcuri82: does it contain itself? i.e., `sourceColumn`?  clarify

arcuri82: also, shouldn't it rather be a Set?

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
```diff
@@ -110,6 +125,22 @@ class SqlForeignKeyGene(
                 .map { it.uniqueId }
                 .toSet()
 
+        // For composite FKs, we need to make sure all the genes point to the same PK.
+        // We find other FK genes in the same action that point to the same target table AND are part of the same composite FK.
+        val otherFksInSameAction = allGenes.asSequence()
+                .flatMap { it.flatView().asSequence() }
+                .filterIsInstance<SqlForeignKeyGene>()
+                .filter { it.uniqueId == uniqueId && it !== this && it.targetTable == targetTable }
+                .filter { otherSourceColumnsInCompositeFK.contains(it.name) }
```
arcuri82, line 138: is the `name` the same as `sourceColumn`??? i mean, here `it` is a FK. shouldn't be something like `otherTargetColumnsInCompositeFK`?  bit confused here... maybe add some more comments, and rename variables if needed

core/src/main/kotlin/org/evomaster/core/sql/schema/ForeignKey.kt
```diff
@@ -7,7 +7,19 @@ package org.evomaster.core.sql.schema
 
 data class ForeignKey(
 
-        val sourceColumns: Set<Column>,
+        /**
+         * The columns in the source table that are referenced by this foreign key
```
arcuri82: ` are referenced by`? you mean something like `define`?

core/src/main/kotlin/org/evomaster/core/sql/schema/ForeignKey.kt
```diff
@@ -7,7 +7,19 @@ package org.evomaster.core.sql.schema
...
+        /**
+         * The target table that is referenced by this foreign key
+         */
+        val targetTableId: TableId,
+
+        /**
+         * The columns in the target table that are referenced by this foreign key.
+         * The order should match the order of the source columns.
+         */
+        val targetColumns: List<Column>
```
arcuri82, line 25: shouldn't track `sourceTableId` as well?

core/src/main/kotlin/org/evomaster/core/sql/SqlActionUtils.kt
```diff
@@ -140,49 +141,104 @@ object SqlActionUtils {
...
+     * @return A boolean value indicating whether all foreign key constraints are valid.
+     * Returns true if valid, otherwise false.
+     */
     fun isValidForeignKeys(actions: List<SqlAction>, errors: MutableList<String>? = null): Boolean {
 
         for (i in 0 until actions.size) {
-
-            val fks = actions[i].seeTopGenes()
+            val currentAction = actions[i]
+            val currentFkGenes = currentAction.seeTopGenes()
```
arcuri82: can use `seeAllGenes()`

core/src/main/kotlin/org/evomaster/core/sql/SqlActionUtils.kt
```diff
@@ -140,49 +141,104 @@ object SqlActionUtils {
...
-                    }
+            // Collect all unique Ids and their associated table IDs
+            val previousSqlActions = actions.subList(0, i)
+            val previousPkMap = mutableMapOf<Long, TableId>()
+            previousSqlActions.forEach { action ->
+                action.seeTopGenes()
+                    .flatMap { it.flatView() }
+                    .filterIsInstance<SqlPrimaryKeyGene>()
+                    .forEach { previousPkMap[it.uniqueId] = it.tableName }
+            }
```
arcuri82: instead of instantiating a `mutableMapOf` which is then filled with a `forEach`, this code can be simplified with a `flatMap` and then a `associateBy`

client-java/sql/src/main/java/org/evomaster/client/java/sql/DbCleaner.java
```diff
@@ -327,12 +327,15 @@ private static void deleteTables(Statement statement, String table, String schem
...
         if (restartIdentityWhenTruncating) {
-            statement.executeUpdate("TRUNCATE TABLE " + table + " RESTART IDENTITY");
-        } else {
-            statement.executeUpdate("TRUNCATE TABLE " + table);
+            sql += " RESTART IDENTITY";
+        }
+        if (type == DatabaseType.POSTGRES) {
+            sql += " CASCADE";
         }
+        statement.executeUpdate(sql);
```
arcuri82, line 338: thx. this fix seems not related to PK/FK, isn't it? for future, it would had been better to have in a separate PR

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
jgaleotti: you are right. This description does not correctly explains what uniqueId is.

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
jgaleotti: I clarified and changed the field from List to Set.

client-java/sql/src/main/java/org/evomaster/client/java/sql/DbCleaner.java
jgaleotti: agree. Unfortunately it was blocking my progress, so I pushed as a change here. But next time I will try to have a separate small PR.

core/src/main/kotlin/org/evomaster/core/search/gene/sql/SqlForeignKeyGene.kt
jgaleotti: I will refactor this code to make it more readable

core/src/main/kotlin/org/evomaster/core/sql/schema/ForeignKey.kt
jgaleotti: I will improve the Javadoc

core/src/main/kotlin/org/evomaster/core/sql/schema/ForeignKey.kt
jgaleotti: The ForeignKey is contained within a data class Table. Therefore I think it is redundant to include the sourceTableId here.

core/src/main/kotlin/org/evomaster/core/sql/SqlActionUtils.kt
jgaleotti: ok. I will remove the flatview then

core/src/main/kotlin/org/evomaster/core/sql/SqlActionUtils.kt
jgaleotti: Thanks. Code looks so much nicer now!

arcuri82 (approved):

PR #1495, opened 2026-04-09
Introducing DynamoDB interception via SDK client class replacement

aschenzle:
Intercepting main operations (Query, Scan, GetItem, BatchGetItems, PutItem, UpdateItem and Delete Item) for both sync and async DynamoDB SDK clients. The interception happens at the SDK level and I've verified it works via an integration test using Docker. There was a dependency conflict while bringing the AWS SDK through Maven, Netty versions would get mixed up for Redis end to end tests so I've made the decision to exclude Netty from Lettuce package and use the Netty version packaged on AWS SDK. All e2e tests are passing.

As this is my first PR I'm looking for exhaustive feedback to quickly learn any conventions I've might missed.


client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
```diff
@@ -0,0 +1,201 @@
...
+
+        @Replacement(type = ReplacementType.TRACKER, id = DDB_ASYNC_SCAN, usageFilter = UsageFilter.ANY, category = ReplacementCategory.DYNAMODB, castTo = "java.util.concurrent.CompletableFuture")
+        public static Object scan(Object client, @ThirdPartyCast(actualType = "software.amazon.awssdk.services.dynamodb.model.ScanRequest") Object request) {
+            return handle(singleton, client, DDB_ASYNC_SCAN, request, "Scan");
+        }
+    }
+
+    protected static Object handle(ThirdPartyMethodReplacementClass singleton, Object client, String id, Object request, String operationName) {
+        long start = System.currentTimeMillis();
+        boolean isAsync = client.getClass().getName().contains("Async");
```
jgaleotti: Why don't we habe a handleAsync() and handleSync() and we call each method replacement differently?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
```diff
@@ -0,0 +1,201 @@
...
+        if (request == null) return Collections.emptyList();
+
+        try {
+            Method getTableNameMethod = request.getClass().getMethod("tableName");
+            String tableName = (String) getTableNameMethod.invoke(request);
+            if (tableName != null) {
+                return Collections.singletonList(tableName);
+            }
+        } catch (NoSuchMethodException ignored) {
+            // no-op
```
jgaleotti: why is this ignored?

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
```diff
@@ -0,0 +1,201 @@
...
+            for (Object key : ((Map<?, ?>) requestItems).keySet()) {
+                if (key instanceof String) {
+                    tableNames.add((String) key);
+                }
+            }
+
+            Collections.sort(tableNames);
+            return tableNames;
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+            return Collections.emptyList();
```
jgaleotti: Why is is defaulted to empty?

docs/options.md
```diff
@@ -138,6 +138,7 @@ There are 3 types of options:
 |`impactAfterMutationFile`| __String__. Specify a path to save collected impact info after each mutation during search, only useful for debugging. *DEBUG option*. *Default value*: `impactSnapshot.csv`.|
 |`impactFile`| __String__. Specify a path to save derived genes. *DEBUG option*. *Default value*: `impact.csv`.|
 |`instrumentMR_BASE`| __Boolean__. Execute instrumentation for method replace with category BASE. Note: this applies only for languages in which instrumentation is applied at runtime, like Java/Kotlin on the JVM. *Default value*: `true`.|
+|`instrumentMR_DYNAMODB`| __Boolean__. Execute instrumentation for method replace with category DYNAMODB. Note: this applies only for languages in which instrumentation is applied at runtime, like Java/Kotlin on the JVM. *Default value*: `true`.|
```
jgaleotti: default value should be `false` until we finish the implementation.

pom.xml
```diff
@@ -604,6 +605,13 @@
                 <groupId>io.lettuce</groupId>
                 <artifactId>lettuce-core</artifactId>
                 <version>${io.lettuce.core.version}</version>
+            <!-- Need to exclude Netty from this package to avoid a clash of versions with AWS SDK -->
```
jgaleotti, line 608: Could you improve the comment why this exclusion is needed?

pom.xml
```diff
@@ -627,6 +635,19 @@
...
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>dynamodb</artifactId>
+                <version>${software.amazon.awssdk.dynamodb.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>netty-nio-client</artifactId>
+                <version>${software.amazon.awssdk.dynamodb.version}</version>
```
jgaleotti, line 648: Shouldn't the scope of this dependency be also `test`?

jgaleotti (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: was trying to optimize for code reuse as there will be 8 lines of repeated code if we split the methods. I'm ok changing if that fits better with project conventions. Please confirm you want this change and I'll do it.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Missing a really important comment. The following comment will replace that no-op: "// Ignored in the case we are calling for a batch request, it will be handled below in the next section" Let me know if that's clear enough.  Or you also want to split the logic for singleTable calls vs batch requests.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Seemed like the best option, as we don't want the interception to throw exceptions and retuning null requires handling it. We are just getting less metadata about the interception for this particular request if for some weird reason it fails (Only case I can think this failing would be a dependency problem with another version of the SDK breaking API compatibility, not expected from AWS).

pom.xml
aschenzle: How about "Need to exclude Netty from this package to avoid a conflict of versions with the Netty dependency introduced by AWS SDK, the AWS SDK version of Netty is newer and all e2e tests pass with it."?

docs/options.md
aschenzle: Done, changed it on EMConfig.kt and regenerated options.md

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
jgaleotti: please change it

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
jgaleotti: Why don't you encapsulate this and throw a RuntimeException?

pom.xml
jgaleotti: ok

pom.xml
aschenzle: done

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Done, will be pushed soon.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Comment updated

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Just committed a refactored version.

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
```diff
@@ -92,94 +105,129 @@ protected String getNameOfThirdPartyTargetClass() {
...
-            String tableName = (String) getTableNameMethod.invoke(request);
-            if (tableName != null) {
-                return Collections.singletonList(tableName);
-            }
+            Method getTableNameMethod = request.getClass().getMethod(METHOD_TABLE_NAME);
+            return (String) getTableNameMethod.invoke(request);
         } catch (NoSuchMethodException ignored) {
-            // Ignored in the case we are calling for a batch request, it will be handled below in the next section
+            // Ignore as BatchGetItem requests do not have tableName and are handled by extractBatchTableNames.
+            return null;
```
jgaleotti, line 219: shouldn't this throw an exception?

jgaleotti: it seems to be differently treated from extractBatchTableNames()

jgaleotti (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/DynamoDbClassReplacement.java
aschenzle: Check the comment on extractTableNames method which is the entrypoint for both GetItem (single table request) and BatchGetItem (possibly multiple Tables). For all requests we first try to extract a singleTableName (this method) if it fails (as BatchGetItem doesn't have a tablesName method) and returns null we know it's a Batch request and we call extractBatchTableNames (line 207) to get a List of tables. This is tested on testBatchGetItem on the test suite.

aschenzle: In fact this applies too all methods not only "gets"

jgaleotti: ack

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1990,6 +1990,11 @@ class EMConfig {
             " on the JVM.")
     var instrumentMR_MONGO = true
 
+    @Cfg("Execute instrumentation for method replace with category DYNAMODB." +
+            " Note: this applies only for languages in which instrumentation is applied at runtime, like Java/Kotlin" +
+            " on the JVM.")
+    var instrumentMR_DYNAMODB = false
```
arcuri82, line 1997: this should be marked with `@Experimental`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
jgaleotti: indeed.

aschenzle: Done, let's wait for the build to check it passes without issues.

aschenzle: All tests passed. https://github.com/aschenzle/EvoMaster/actions/runs/24733845308

aschenzle: @arcuri82 added you as collaborator just in case so you can see the build.

PR #1501, opened 2026-04-13
Java/JS regex support for escapes within charclasses, JS regex null escape, identity escapes & legacy octal escape

lmasroca:
Java/JS regex support for some escapes within character classes ([\s\S], etc.)
JS regex support for null escape (\0), identity escapes (\a, etc.), legacy octal escapes (\1, \012, \377, etc.) and fix for control letter escapes (different behavior regarding inside character class among other things)


core/src/main/antlr4/org/evomaster/core/parser/RegexEcma262.g4
```diff
@@ -86,40 +86,46 @@ bracketQuantifierRange
...
- //| IdentityEscape
+CharacterEscape
+ : SLASH ControlEscape
+ | SLASH HexEscapeSequence
+ | SLASH UnicodeEscapeSequence
+ | SLASH OctalEscapeSequence // legacy octal escapes are deprecated, but this also works for null escape (\u0000)
+ | SLASH IdentityEscape
+ ;
+
+//TODO backreferences
```
jgaleotti, line 106: explain in this comment a bit more what backreferences are

core/src/main/antlr4/org/evomaster/core/parser/RegexEcma262.g4
```diff
@@ -170,26 +176,29 @@ classAtomNoDash
...
 
 
-//TODO
-//ClassEscape
-// : CharacterClassEscape
-//// | DecimalEscape
-//// | 'b'
-// //| CharacterEscape
-// ;
+// TODO
```
jgaleotti: what is this TODO?

jgaleotti (changes requested):

core/src/main/antlr4/org/evomaster/core/parser/RegexEcma262.g4
lmasroca: This was for \b within charclass, in JS some escapes behave differently within charclasses. In the case of \b within charclass it is interpreted as backspace. Implemented that now and removed TODO. The \b and \B escapes outside charclass (word boundary assertions) are not implemented yet.

jgaleotti (approved):

arcuri82 (approved):

PR #1503, opened 2026-04-16
low-code implementation

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/enterprise/service/WFCReportWriter.kt
```diff
@@ -44,6 +48,53 @@ class WFCReportWriter {
...
+        val indexHtml = readResource("$prefix/index.html")
+        val reportJs = readResource("$prefix/assets/report.js")
+        val reportCss = readResource("$prefix/assets/report.css")
+        val iconSvg = readResource("$prefix/assets/icon.svg")
+
+        val reportJsonPath = Paths.get(config.outputFolder, "report.json").toAbsolutePath()
+        val reportJson = if (Files.exists(reportJsonPath)) {
+            reportJsonPath.toFile().readText(Charsets.UTF_8)
+        } else {
+            "{}"
```
arcuri82: shouldn't rather generate no file, and issue an error message here?

arcuri82 (changes requested):

arcuri82 (approved):

PR #1504, opened 2026-04-17
Skip class replacement on known classloaders to avoid noisy warning

aschenzle:
Avoid unnecessary warning to the user on every e2e test run. This warning also showed up when running generated tests to run on the SUT code base and made me waste a couple of hours as I was working on interception/method replacement. Overall I think this will save many hours for new developers.


arcuri82 (approved):

arcuri82 (changes requested):
hi @aschenzle . it looks fine, thx! but, you should no longer use a fork, but work directly in a branch of EM. reason is that CI is not run on PRs from forks... so I cannot see here if anything got broken...  please make a new PR from a branch, and wait for CI to be green before requesting my review

PR #1513, opened 2026-04-22
Introduce Build metrics step and upgrade CI Actions to Node 24 compatible versions.

aschenzle:
The original intention of this branch / PR was to introduce a new top level job in the build process that would display build metrics (for now just a count of warnings and errors). It will store the count on the environment for 30 days and each build of master will overwrite it, if the current build being executed increases the warnings or error counts the step will show as failed without blocking the build (yellow), if it's equal or less it will show all green. This is "broken window" mechanism to start giving visibility to something developers often oversee. The reviewer of a PR can choose to be strict and require changes to reduce the warning and error count or can let it pass if the merge is important.

As part of this I got carried away and upgraded CI actions to Node 24 compatible versions.

If all test pass please review and give me feedback. I still needs this to be merged to master once to make sure the mechanism works (retrieve the old metrics file and comparing to current build).

### DIFF
--- a/.github/actions/full-build-metrics/action.yml
+++ b/.github/actions/full-build-metrics/action.yml
@@ -94,7 +94,7 @@ runs:
         fi
 
     - name: Upload current metrics artifact
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: ${{ inputs.artifact-name }}
         path: build-metrics.env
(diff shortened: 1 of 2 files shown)


aschenzle:
@arcuri82 I just saw your other commits where you updated the Github actions versions too. Once merged I will rebase this branch onto those changes.

arcuri82 (approved):
@aschenzle thx. looks fine. i ll merge into master, and see how it works there. worst case, we can then disable it if we seen any major issue

PR #1519, opened 2026-04-23
support handleFlaky with tests

man-zhang:


dev-vektr-ai:
Thanks for the PR, @man-zhang. I took an initial look at [PR #1519](https://github.com/WebFuzzing/EvoMaster/pull/1519).

This patch appears to add support for running generated tests while handling flaky scenarios by threading a `handleFlakyWithTests` flag through the `runTestHandlingFlaky*` helpers and enabling that path in `FlakinessDetectBlackboxEMTest.kt`.

I do not see any immediate blocking issue from the current diff. A maintainer can follow up with a deeper review, but the change looks focused and the added test usage helps clarify the intended behavior.

arcuri82 (approved):

PR #1521, opened 2026-04-24
PR/DynamoDB operations and parsers

aschenzle:
Introduces DynamoDB operations and parsers. The design differ a little bit from Mongo and other DBs as DynamoDB has complex expression that need to be parsed in order of precedence. There are top level parsers instead of simple selectors and the logic is split per Dynamo API call and some helper classes. To avoid writing a full blown parser I'm using Antlr as a runtime dependency in controller, I shaded the dependency but please let me know if this could bring any problems or if I'm missing something.

Some random comments: 1) Netty dependency is no longer explicitly needed in controller as now AWSSDK is bringing it. 2)  There's a fix for hibernate-validator dependency which is unrelated but eliminates some warnings. 3) Added SimpleLogger to controller tests, eliminates a warning when running tests and actually logs (for example I could see the whole Docker log which wasn't visible before), this will probably increase the warning count when running tests, this is just surfacing them.


client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/operations/BetweenOperation.java
```diff
@@ -0,0 +1,36 @@
...
+    private final Object upperBound;
+
+    /**
+     * Creates a BETWEEN operation.
+     *
+     * @param fieldName attribute path
+     * @param lowerBound lower bound value
+     * @param upperBound upper bound value
+     */
+    public BetweenOperation(String fieldName, Object lowerBound, Object upperBound) {
```
jgaleotti, line 20: why lower and upper bounds are objects?

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/DynamoDbReflectionHelper.java
```diff
@@ -0,0 +1,46 @@
+package org.evomaster.client.java.controller.dynamodb;
+
+import java.lang.reflect.Method;
+
+/**
+ * Shared reflection helpers used by DynamoDB parser utilities.
+ */
+public final class DynamoDbReflectionHelper {
```
jgaleotti, line 8: Why is this class for? I am not understanding why reflection is needed for parsing DynamoDB queries

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/DynamoDbRequestParser.java
```diff
@@ -0,0 +1,71 @@
...
+        if (request == null || apiMethodName == null) {
+            return Collections.emptyMap();
+        }
+
+        DynamoDbApiMethodParser parser = parsersByApiMethod.get(apiMethodName);
+        if (parser == null) {
+            return Collections.emptyMap();
+        }
+
+        Map<String, QueryOperation> parsed = parser.parseRequest(request);
```
jgaleotti: The request is a dynamoDB object?

jgaleotti (changes requested):

jgaleotti:
Why don't you dump the dynamoDB query into a string and then parse it back from it?

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/operations/BetweenOperation.java
aschenzle: Because they can be strings (or binary! just realized this one, don't think I tested it). From DynamoDB doc: "BETWEEN : Greater than or equal to the first value, and less than or equal to the second value.

AttributeValueList must contain two AttributeValue elements of the same type, either String, Number, or Binary (not a set type). A target attribute matches if the target value is greater than, or equal to, the first element and less than, or equal to, the second element. If an item contains an AttributeValue element of a different type than the one provided in the request, the value does not match. For example, {"S":"6"} does not compare to {"N":"6"}. Also, {"N":"6"} does not compare to {"NS":["6", "2", "1"]}"

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/DynamoDbReflectionHelper.java
aschenzle: Added explanation in DynamoDbBaseApiMethodParser

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/operations/BetweenOperation.java
aschenzle: See comments added in latest commit

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/DynamoDbRequestParser.java
aschenzle: Yes, renamed all parameters like these to ddbRequest to make it more clear.

aschenzle:
> Why don't you dump the dynamoDB query into a string and then parse it back from it?

Check latest commit with comments added.

jgaleotti:
Please remember to re-ask for a review. Otherwise it might miss to review your changes.

jgaleotti (approved):

arcuri82:
@jgaleotti @aschenzle  sorry, due to exams and sickness, don't think i ll have any time to look at this till next week

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public Map<String, QueryOperation> parseRequest(Object ddbRequest) {
+        Object requestItemsObj = invokeNoArg(ddbRequest, METHOD_REQUEST_ITEMS);
+        if (!(requestItemsObj instanceof Map<?, ?>)) {
+            return Collections.emptyMap();
```
arcuri82, line 36: are there cases in which this is a normal behavior? 
or should rather throw an exception? or issue a warning? 
i mean, what are the cases in which METHOD_REQUEST_ITEMS does not return a map? 
or should explicitly as well check for `null`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
```diff
@@ -0,0 +1,76 @@
...
+            String tableName = entry.getKey() == null ? null : String.valueOf(entry.getKey());
+            if (tableName == null || tableName.trim().isEmpty()) {
+                continue;
+            }
+
+            Object keysAndAttributes = entry.getValue();
+            Object keysObj = invokeNoArg(keysAndAttributes, METHOD_KEYS);
+            if (!(keysObj instanceof Collection<?>)) {
+                continue;
+            }
```
arcuri82, line 53: similar to previous discussion. is this possilbe in a normal scenario? 
if yes, this is fine, but add a comment.
if not, should handle it (eg warn log it)

client-java/controller/pom.xml
```diff
@@ -140,6 +140,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- Needed for DynamoDb parsing -->
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
```
arcuri82, line 150: this can be problematic, as dependency used as well in core. 
you must here fix the version to 4.9.3.
also, add comment stating this dependency cannot be upgraded to 4.10 or later due to JDK 8 requirement in client-java

client-java/controller/pom.xml
```diff
@@ -204,10 +211,15 @@
...
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
-            <artifactId>netty-nio-client</artifactId>
+            <artifactId>dynamodb</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- Removes a warning and actually logs on tests executions -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
```
arcuri82, line 226: why are you configuring slf4j here in the tests of client-java? add an explanation comment here

client-java/controller/pom.xml
```diff
@@ -265,6 +276,23 @@
                 <artifactId>maven-compiler-plugin</artifactId>
             </plugin>
 
+            <!-- Needed for DynamoDB parsing -->
+            <plugin>
+                <groupId>org.antlr</groupId>
+                <artifactId>antlr4-maven-plugin</artifactId>
```
arcuri82, line 286: force here a version that is compatible with JDK 8

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
aschenzle: Hi @arcuri82 no, it's not expected (it's either a map or null at most), the design decision here is around allowing the program to continue with less information (given that DDB is a secondary target) vs throwing and catching exceptions upstream which could potentially (if not well managed) stop the whole test generation. Let me know if you strongly prefer Exceptions instead.

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
aschenzle: Not normal, will add a warning log.

client-java/controller/pom.xml
aschenzle: Done

client-java/controller/pom.xml
aschenzle: When running tests with Docker in particular, there was a warning about missing logging. When I added this I was able to see all Docker initialization logs printing correctly (and the warning gone). Is this a good enough explanation to add logging?

client-java/controller/pom.xml
aschenzle: @arcuri82 On a second thought about this, isn't the requirement that the generated code will run in Java8? Don't we actually compile (and generate ANTLR code) on Java 17/21?

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
arcuri82: not throwing exception is fine, but silently ignoring the problem is not. you can use `SimpleLogger` to log this issue

client-java/controller/pom.xml
arcuri82: you can add: `Some third-party library we use in our test scaffolding might need SLF4j`

client-java/controller/pom.xml
arcuri82: `core` and `client-java` have different constraints when it comes to JDK support

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
aschenzle: Added logging

jgaleotti (approved):

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public Map<String, QueryOperation> parseRequest(Object ddbRequest) {
+        Object requestItemsObj = invokeNoArg(ddbRequest, METHOD_REQUEST_ITEMS);
+        if (!(requestItemsObj instanceof Map<?, ?>)) {
+            return Collections.emptyMap();
```
arcuri82, line 36: you wrote `Added logging` but i see no logging before that `return Collections.emptyMap();`.
also the check seems wrong, as not preserving semantic, isn't it? eg, if `invokeNoArg` returns `null` then here you would wrongly return an `emptyMap`, isn't it?

client-java/controller/pom.xml
```diff
@@ -265,6 +276,23 @@
...
+                <configuration>
```
arcuri82, line 288: you marked this as resolved, but i don't see any `<version>` declaration here. how come?

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
```diff
@@ -0,0 +1,79 @@
...
+        Map<?, ?> requestItems = (Map<?, ?>) requestItemsObj;
+        for (Map.Entry<?, ?> entry : requestItems.entrySet()) {
+            String tableName = entry.getKey() == null ? null : String.valueOf(entry.getKey());
+            if (tableName == null || tableName.trim().isEmpty()) {
+                SimpleLogger.uniqueWarn("Failed to parse tablename for BatchGetItem");
+                continue;
+            }
+
+            Object keysAndAttributes = entry.getValue();
+            Object keysObj = invokeNoArg(keysAndAttributes, METHOD_KEYS);
```
arcuri82, line 49: here we should check if `keysObj ` is null, and don't issue a warning if that happens, as that could be normal behavior for that method.

arcuri82 (changes requested):

jgaleotti:
What is the status of this PR?

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
aschenzle: Now really added

client-java/controller/src/main/java/org/evomaster/client/java/controller/dynamodb/parsers/BatchGetItemApiMethodParser.java
aschenzle: somehow my previous comment got deleted but my understanding is that instanceof (right below) checks for nulls. Do you want a more explicit check?

arcuri82 (approved):

PR #1522, opened 2026-04-24
Hiding Javadoc unnecessary warnings by using quiet mode

aschenzle:
Without quiet mode, Javadoc is very verbose and prints a [WARNING] line for each HTML file it generates. This goes against the initiative of reducing overall warnings as it will generate a new warning for each class introduced. Note that missing Javadoc still generates and prints warnings (check this build), I'm going to pursue another change for that which I will discuss with our not-so-benevolent dictator :D

### DIFF
--- a/client-java/pom.xml
+++ b/client-java/pom.xml
@@ -66,8 +66,8 @@
     <build>
         <!--
             When we make a release, the client-java must be published on Maven Central.
-            This requires to create jars for sources and javadocs.
-            As the javadocs build might fail, we force its creation at each build,
+            This requires creating jars for sources and Javadocs.
+            As the Javadocs build might fail, we force its creation at each build
             so that any issue can be fixed as soon as it is introduced.
         -->
         <plugins>
@@ -91,6 +91,9 @@
                 <version>3.1.1</version>
                 <configuration>
                     <source>8</source>
+                    <!-- Reducing Javadoc noise, this only leaves the missing Javadoc warnings
+                        and avoids unnecessary warnings about output to HTML -->
+                    <quiet>true</quiet>
                 </configuration>
                 <executions>
                     <execution>
--- a/client-java/pom.xml
+++ b/client-java/pom.xml
@@ -91,9 +91,6 @@
                 <version>3.12.0</version>
                 <configuration>
                     <source>8</source>
-                    <!-- Reducing Javadoc noise, this only leaves the missing Javadoc warnings
-                        and avoids unnecessary warnings about output to HTML -->
-                    <quiet>true</quiet>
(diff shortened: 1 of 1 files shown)


jgaleotti:
Although I am not added as a reviewer, IMHO I think that removing this is kind of warning seems reasonable.

arcuri82:
in line of principle, I do not like to mute warning messages, even if, in all honesty, I don't think we would prioritize fixing those warnings any time soon compared to other more important tasks.

aschenzle:
Hi @arcuri82, I'm re-running the job as the build logs expired and I can't download it. Once it's done I will do a diff with a normal build to show you which warnings this hides (which according to my theory are only the warnings that look like this: "Warning:  Generating /home/runner/work/EvoMaster/EvoMaster/client-java/controller-api/target/apidocs/org/evomaster/client/java/controller/api/dto/class-use/ActionDto.html..." while leaving the ones like this: "Warning:  /home/runner/work/EvoMaster/EvoMaster/client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/ControllerConstants.java:15: warning: no comment
Warning:  public static final String TEST_RESULTS = "/testResults";"

aschenzle:
Hi @arcuri82 here are 2 builds with and without Javadoc quiet mode: Scanned through them and the result is as I expected.

AI summary: - The no-quiet build adds 1,355 Javadoc doclet progress lines, such as `Loading source files`, `Building index`, `Standard Doclet`, and `Generating .../apidocs/...`. These are warning-level log lines but are mostly verbosity noise, not new compiler or Javadoc diagnostics.
- Java/Javadoc diagnostic warning counts are unchanged overall: both logs have 604 Java/Javadoc warning diagnostics and 9 Javadoc warning summary lines.

Let me know if hiding this type of warnings is ok or there's a reason not to do it.

[kotlin-warnings-2-no-quiet-javadoc-warnings.txt](https://github.com/user-attachments/files/27252901/kotlin-warnings-2-no-quiet-javadoc-warnings.txt)
[reducing-javadoc-noise-javadoc-warnings.txt](https://github.com/user-attachments/files/27252903/reducing-javadoc-noise-javadoc-warnings.txt)

arcuri82:
@aschenzle  we are using a very old version of plugin. can you upgrade to latest `3.12.0` and see how logs look like?

aschenzle:
@arcuri82 You were right, latest version fixed the verbosity without the need to use quiet mode. Check by upgrading on another branch and comparing the outputs. Reverting quiet mode on this branch so the only thing to really merge is the version bump.

arcuri82 (approved):

PR #1524, opened 2026-04-25
Removes one unchecked warning and deletes deprecated PartialOracle

aschenzle:
For reviewer: 1) Check if the way of avoiding the unchecked warning for List<CharacterRange> is overkill. 2) Let me know if deleting deprecated and unused classes is ok. 3) Some Java changes doing exactly the opposite (suppressing the warnings) are included and I know how ironic that looks like. Feel free to let me know if some of these changes need to be removed.


client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
```diff
@@ -31,6 +31,7 @@ public static Object getValue(Object document, String fieldName) {
         }
     }
 
+    @SuppressWarnings("unchecked")
```
arcuri82: shouldn't add `@SuppressWarnings`. as a rule of thumb, adding annotations just to remove warning messages from build is not something i like

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
```diff
@@ -47,6 +48,7 @@ public static Boolean documentContainsField(Object document, String field) {
...
```
arcuri82: see previous comment

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
aschenzle: @arcuri82 revert or do this monstrosity which doesn't have warnings but creates a copy? "Set<?> keySet = (Set<?>) document.getClass().getMethod("keySet").invoke(document);
            return keySet.stream()
                    .map(String.class::cast)
                    .collect(Collectors.toCollection(LinkedHashSet::new));"

arcuri82: no, introducing "monstrosities" just to remove a warning is even worse. just revert

aschenzle: I've removed this method as it was a copy of documentKeys with only one call vs 8 on documentKeys. So one less warning anyway and less duplicated code.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java
aschenzle: Removed this one.

jgaleotti:
what is the status of this PR? Were all changes requested by @arcuri82 on Apr. 27th addressed?

aschenzle:
Dismissing old PR

PR #1525, opened 2026-04-26
 Json patch, initial pr, structures only

suarezrominajulieta:
## Summary

Introduces the foundational gene structure and builder needed to model JSON Patch

### Gene model (`core/src/main/kotlin/.../search/gene/patch/`)
- `JsonPatchOperationGene` — enum-backed gene representing a single RFC 6902 op (`add`, `remove`, `replace`, `move`, `copy`, `test`)
- `JsonPatchPathOnlyGene` — composite gene for ops that only require a `path` field (`remove`)
- `JsonPatchPathValueGene` — composite gene for ops that require `path` + `value` (`add`, `replace`, `test`)
- `JsonPatchFromPathGene` — composite gene for ops that require `path` + `from` (`move`, `copy`)
- `JsonPatchDocumentGene` — top-level gene representing a full JSON Patch document (array of operation objects); handles serialization to JSON and mutation logic

### Builder (`core/src/main/kotlin/.../problem/rest/builder/`)
- `JsonPatchGeneBuilder` — constructs a `JsonPatchDocumentGene` from an OpenAPI schema, mapping `PATCH` request body fields to the appropriate sub-gene variant

### Tests (`core/src/test/kotlin/...`)
- Unit tests for each gene class and the builder
- `GeneSamplerForTests` and `GeneNumberOfGenesTest` updated to include new gene types

## What's NOT included yet
- Integration with `RestActionBuilderV3` (wiring the builder into the main action pipeline)
- E2E test against a real PATCH endpoint

## Notes
This is a structural PR — all new classes compile and their unit tests pass, but the genes are not yet invoked during actual fuzzing runs.


jgaleotti:
Hi @suarezrominajulieta, first ask for my review and then, when I aprove the PR, I will request @arcuri82 's review.

suarezrominajulieta:
Hi @jgaleotti ! Okey, done :)

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/JsonPatchGeneBuilder.kt
```diff
@@ -0,0 +1,131 @@
...
+import org.evomaster.core.search.gene.wrapper.NullableGene
+import org.evomaster.core.search.gene.wrapper.OptionalGene
+
+/**
+ * Builder for JSON Patch document genes.
+ *
+ * Provides heuristics that extract valid JSON Pointer paths from a resource schema
+ * and group them by field type to construct type-safe path-value pairs.
+ */
+object JsonPatchGeneBuilder {
```
jgaleotti: Why it is saying that it provides "heuristics"? This class **deterministically** creates an JsonPathGene from a given schema

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,144 @@
...
+         * Builds the ArrayGene of patch operations from the resource schema.
+         *
+         * All six operation choices are always present in the ChoiceGene template so that
+         * mutation can switch between them. Only the active choice is included in the phenotype.
+         */
+        fun buildOperationsArray(
+            resourceSchema: Gene?
+        ): ArrayGene<ChoiceGene<JsonPatchOperationGene>> {
+
+            val allPaths = JsonPatchGeneBuilder.extractAllPaths(resourceSchema)
```
jgaleotti: what happens if resourceSchema is null?

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,144 @@
...
+            val pathList = if (allPaths.isEmpty()) listOf("/") else allPaths.map { it.path }
+            val pathEnum = EnumGene<String>("path", pathList)
+
+            val choices = mutableListOf<JsonPatchOperationGene>()
+
+            // Operations that only need a path
+            choices.add(JsonPatchPathOnlyGene("remove", pathEnum.copy() as EnumGene<String>))
```
jgaleotti: replace "remove" and other strings with constants

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,144 @@
...
+                JsonPatchFromPathGene(
+                    "copy",
+                    fromGene = pathEnum.copy() as EnumGene<String>,
+                    pathGene = pathEnum.copy() as EnumGene<String>
+                )
+            )
+
+            // Operations that need path + value; pairs are grouped by schema field type
+            val pathValueEntries = JsonPatchGeneBuilder.buildPathValueEntries(allPaths)
+            val effectiveEntries: List<PairGene<EnumGene<String>, Gene>> =
```
jgaleotti: what does effectiveEntries mean? Shouldn't this list be called pathValuePairGenes?

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchFromPathGene.kt
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * JSON Patch operation gene for operations that require "op", "from", and "path" fields.
+ * Used for: move, copy.
+ */
+class JsonPatchFromPathGene(
+    operationName: String,
+    val fromGene: EnumGene<String>,
+    val pathGene: EnumGene<String>,
+    geneName: String = "${operationName}Op"
+) : JsonPatchOperationGene(geneName, operationName, listOf(fromGene, pathGene)) {
```
jgaleotti: check that the operation name is move or copy. Why does it add the "Op" ?

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchFromPathGene.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        val from = fromGene.getValueAsPrintableString(previousGenes, mode, targetFormat, extraCheck)
+        val path = pathGene.getValueAsPrintableString(previousGenes, mode, targetFormat, extraCheck)
+        return "{\"op\":\"$operationName\",\"from\":$from,\"path\":$path}"
```
jgaleotti: shouldn't this respect the targetFormat? This could also be printed as a XML in some cases.

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
```diff
@@ -0,0 +1,31 @@
...
+    children: List<Gene>
+) : CompositeFixedGene(name, children) {
+
+    override fun checkForLocallyValidIgnoringChildren() = true
+
+    override fun randomize(randomness: Randomness, tryToForceNewValue: Boolean) {
+        children.filter { it.isMutable() }.forEach { it.randomize(randomness, tryToForceNewValue) }
+    }
+
+    override fun customShouldApplyShallowMutation(
```
jgaleotti, line 36: Add comment to explaon why is this always set to "false"

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchPathOnlyGene.kt
```diff
@@ -0,0 +1,40 @@
...
+    geneName: String = "${operationName}Op"
+) : JsonPatchOperationGene(geneName, operationName, listOf(pathGene)) {
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        val path = pathGene.getValueAsPrintableString(previousGenes, mode, targetFormat, extraCheck)
```
jgaleotti, line 30: Again, printing should be agnostic of JSON

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,144 @@
...
+ * The primary constructor builds the operations array from [resourceSchema].
+ * The private secondary constructor accepts a pre-built array, used internally by [copyContent].
+ */
+class JsonPatchDocumentGene private constructor(
+    name: String,
+    val resourceSchema: Gene?,
+    operationsArr: ArrayGene<ChoiceGene<JsonPatchOperationGene>>
+) : CompositeFixedGene(name, listOf(operationsArr)) {
+
+    constructor(name: String, resourceSchema: Gene? = null)
```
jgaleotti, line 30: move this logic to JsonPatchDocumentGeneBuilder

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/JsonPatchGeneBuilder.kt
jgaleotti: Do not include the schema extraction heuristics in this PR

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchPathOnlyGene.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.evomaster.core.search.gene.utils.GeneUtils
+
+/**
+ * JSON Patch operation gene for operations that only require an "op" and "path" field.
+ * Currently used for: remove.
+ */
+class JsonPatchPathOnlyGene(
+    operationName: String,
+    val pathGene: EnumGene<String>,
+    geneName: String = "${operationName}Op"
```
jgaleotti: why do we have operationName and geneName?

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchPathValueGene.kt
```diff
@@ -0,0 +1,55 @@
...
+    geneName: String = "${operationName}Op"
+) : JsonPatchOperationGene(geneName, operationName, listOf(pathValueChoice)) {
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        val activePair = pathValueChoice.activeGene()
```
jgaleotti: same as previous comments

core/src/test/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGeneTest.kt
```diff
@@ -0,0 +1,143 @@
...
+ * a ChoiceGene that selects among the six standard operations:
+ *   remove, move, copy, add, replace, test.
+ *
+ * When [resourceSchema] is provided, paths are extracted from the schema and value genes
+ * are typed to match each field's schema type. Otherwise a single "/" fallback path is used.
+ *
+ * The primary constructor builds the operations array from [resourceSchema].
+ * The private secondary constructor accepts a pre-built array, used internally by [copyContent].
+ */
+class JsonPatchDocumentGene private constructor(
```
jgaleotti: what is this private constructor?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchPathOnlyGene.kt
suarezrominajulieta: I'm not sure what I should implement. Could you explain again what we were expecting here?

I created the `operationName` property, which inherits from the abstract class, to verify and distinguish between a `pathfrom` that is a `move` and a `pathfrom` that is a `copy`, and to ensure that each is printed correctly.

And then there's the `geneName` or `name`, which should be the name of the gene, which can be anything, like "Gene1".

I'm wondering if the child classes should only have `operationName` and not `name`, or if we should keep both, but make them exactly the same as those defined by the parent class in the inheritance.

jgaleotti: I understood why we have both. Please keep geneName (name) and operationName.

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
suarezrominajulieta: Can/should we let it do nothing for this pr?

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * All six RFC 6902 operation choices (remove, move, copy, add, replace, test) are always
+     * present in the ChoiceGene template so that mutation can switch freely between them.
+     * [resourceSchema] is reserved for future schema-based path extraction and is ignored for now.
+     */
+    fun buildOperationsArray(
+        // TODO: resourceSchema is ignored in this PR — path extraction will be wired in a follow-up
+        resourceSchema: Gene? = null,
+        paths: List<String> = DEFAULT_PATHS
+    ): ArrayGene<ChoiceGene<JsonPatchOperationGene>> {
```
jgaleotti: Consider a resourceSchema, even though still it is not being extracted from the SUT

core/src/main/kotlin/org/evomaster/core/search/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,85 @@
...
+                pathGene  = pathEnum.copy() as EnumGene<String>
+            )
+        )
+
+        // TODO: replace StringGene with a schema-aware gene derived from resourceSchema
+        //  (e.g. IntegerGene, BooleanGene, ObjectGene) once path extraction is wired in
+        val entryTemplate = PairGene<EnumGene<String>, Gene>(
+            "entry_0",
+            pathEnum.copy() as EnumGene<String>,
+            StringGene("value")
```
jgaleotti: Consider the value type in the schema

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.evomaster.core.search.service.mutator.genemutation.SubsetGeneMutationSelectionStrategy
+
+/**
+ * Represents a JSON Patch document (RFC 6902) as a gene.
+ *
+ * The document is an array of patch operations. Each operation is represented by
+ * a ChoiceGene that selects among the six standard operations:
+ *   remove, move, copy, add, replace, test.
+ *
+ * [resourceSchema] is stored for future schema-aware path extraction but is not yet analysed.
```
jgaleotti: update this javadoc

core/src/test/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGeneTest.kt
```diff
@@ -0,0 +1,303 @@
...
+            ))
+        )
+        val result = op.getValueAsPrintableString()
+        assertEquals("{\"op\":\"add\",\"path\":\"/name\",\"value\":\"Alice\"}", result)
+    }
+
+    // --- XML serialization ---
+
+    @Test
+    fun testRemoveOperationXmlFormat() {
```
jgaleotti, line 171: Add a test case checking JSON Format for this

core/src/test/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGeneTest.kt
```diff
@@ -0,0 +1,303 @@
...
+        val op = JsonPatchPathOnlyGene(JsonPatchOperationGene.OP_REMOVE, JsonPatchOperationGene.OP_REMOVE, EnumGene("path", listOf("/x")))
+        val result = op.getValueAsPrintableString(mode = GeneUtils.EscapeMode.XML)
+        assertEquals("<operation><op>remove</op><path>/x</path></operation>", result)
+    }
+
+    @Test
+    fun testMoveOperationXmlFormat() {
```
jgaleotti, line 189: Add a test case checking JSON Format

suarezrominajulieta:
Hi @jgaleotti ! i added the following as previously talked:

1.  the resourceSchema logic when given, or random if not and it's tests (and value entries with random string or integer genes also).
3.  and the whole json document tests for json and xml printing.

 
I'd appreciate it if you could review it now, thanks!

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        val inner = operationsArray.getValueAsPrintableString(previousGenes, mode, targetFormat, extraCheck)
+        return if (mode == GeneUtils.EscapeMode.XML) "<patch>$inner</patch>" else inner
```
jgaleotti, line 58: Add comment regarding lack of XML convention for JSON Patch documents written in XML

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,95 @@
...
+ *   remove, move, copy, add, replace, test.
+ *
+ * [resourceSchema] is stored for future schema-aware path extraction but is not yet analysed.
+ * Until that step is wired in, paths come from [JsonPatchDocumentGeneBuilder.DEFAULT_PATHS].
+ *
+ * The private constructor accepts a pre-built operations array (used by [copyContent]).
+ */
+class JsonPatchDocumentGene private constructor(
+    name: String,
+    val resourceSchema: Gene?,
```
jgaleotti: it seems to me that the resourceSchema does not need to be stored

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchFromPathGene.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String = formatOperation(
+        mode,
+        OpField("from", fromGene.getValueAsRawString()),
```
jgaleotti: shouldn't this be fromGene.getValueAsPrintableString()

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchPathOnlyGene.kt
```diff
@@ -0,0 +1,43 @@
...
+    val pathGene: EnumGene<String>
+) : JsonPatchOperationGene(name, operationName, listOf(pathGene)) {
+
+    init {
+        require(operationName == JsonPatchOperationGene.OP_REMOVE) {
+            "JsonPatchPathOnlyGene only supports 'remove', got: $operationName"
+        }
+    }
+
+    override fun getValueAsPrintableString(
```
jgaleotti, line 24: getValueAsPrintableString should be defined in the superclass JsonPatchOperationGene

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
```diff
@@ -0,0 +1,58 @@
...
+        const val OP_REPLACE = "replace"
+        const val OP_MOVE    = "move"
+        const val OP_COPY    = "copy"
+        const val OP_TEST    = "test"
+    }
+
+    /** A single field in a JSON Patch operation, with its serialized value.
+     *  [quoted] controls whether the value is wrapped in quotes in JSON output
+     *  (true for string fields like path/from; false for typed fields like value). */
+    protected data class OpField(val name: String, val value: String, val quoted: Boolean = true)
```
jgaleotti: I think this class is not actually needed

jgaleotti (changes requested):

suarezrominajulieta:
Hi, @jgaleotti  i made the changes we last talked about: 

1. Add a comment stating that we are not implementing a convention, because there isn't one, by using `patch` as the root's nametag.
2. We can remove the resource schema: receive it, build the template with it, and discard it.
3. The parent operation should only have `getValueAsPrintableString`, and the children should not:

I had to keep getValueAsRawString with EnumGene because getValueAsPrintableString in enums, prints with "", i.e: "/example", and it would print an xml like this <path>"/x"</path> instead of <path>/x</path>.

ty!

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,165 @@
...
+     * All six RFC 6902 operation choices (remove, move, copy, add, replace, test) are always
+     * present in the ChoiceGene template.
+     *
+     * - If [resourceSchema] yields fields: paths and typed value genes come from the schema.
+     * - Otherwise: paths are generated randomly via [randomness] (a fresh [Randomness] is used
+     *   when none is supplied), and value choices are [StringGene] + [IntegerGene].
+     */
+    fun buildOperationsArray(
+        resourceSchema: Gene? = null,
+        randomness: Randomness? = null
```
jgaleotti: I don't remember discussing the creation of random paths when no resourceSchema is provided. I am not very confortable having a "nondetermintisc" gene creation.

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,165 @@
...
+        } else {
+            buildFromPaths(generateRandomPaths(randomness ?: Randomness()))
+        }
+    }
+
+    // ---------------------------------------------------------------------------
+    // private helpers
+    // ---------------------------------------------------------------------------
+
+    private fun generateRandomPaths(randomness: Randomness): List<String> =
```
jgaleotti: I am not sure about this renadom path creation. I think this builder should be completely deterministic.

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,165 @@
...
+                }
+
+        return assemble(pathEnum, pathValueEntries)
+    }
+
+    /** No-schema case: both [StringGene] and [IntegerGene] offered as value choices. */
+    private fun buildFromPaths(
+        paths: List<String>
+    ): ArrayGene<ChoiceGene<JsonPatchOperationGene>> {
+        val pathEnum = EnumGene<String>("path", paths)
```
jgaleotti: use constants instead of strings such as "path", "entry_string", etc.

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
```diff
@@ -0,0 +1,87 @@
...
+        const val OP_TEST    = "test"
+    }
+
+    override fun getValueAsPrintableString(
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        // path/from fields use getValueAsRawString() instead of getValueAsPrintableString() because
```
jgaleotti: I am still not sure about using getValueAsRawString(). As an example, what happens if I have an ObjectGene with an EnumGene?

core/src/test/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilderTest.kt
```diff
@@ -0,0 +1,337 @@
...
+            JsonPatchDocumentGeneBuilder.buildOperationsArray(randomness = rand).template)
+    }
+
+    // -------------------------------------------------------------------------
+    // buildOperationsArray — operation order and types
+    // -------------------------------------------------------------------------
+
+    private fun children(randomness: Randomness = rand): List<JsonPatchOperationGene> =
+        JsonPatchDocumentGeneBuilder.buildOperationsArray(randomness = randomness)
+            .template.getViewOfChildren().map { it as JsonPatchOperationGene }
```
jgaleotti, line 56: I do not like builder using randomness.

core/src/test/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilderTest.kt
```diff
@@ -0,0 +1,337 @@
...
+
+    @Test
+    fun testEmptySchemaFallsBackToRandomPaths() {
+        val schema = ObjectGene("body", emptyList())
+        val array = JsonPatchDocumentGeneBuilder.buildOperationsArray(resourceSchema = schema, randomness = rand)
+        val removeOp = array.template.getViewOfChildren()[0] as JsonPatchPathOnlyGene
+        val paths = (removeOp.pathGene as EnumGene<*>).values.map { it.toString() }
+        assertTrue(paths.isNotEmpty() && paths.all { it.startsWith("/") })
+    }
+}
```
jgaleotti: Add test cases for the scenario where there are multiple paths with the same value type

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
suarezrominajulieta: ok, but what should we do in case there's no schema provided, how do we build the enum genes for the paths?

suarezrominajulieta: should we make it non nulleable, and throw an exception if not provided?

jgaleotti: Let's try having a schema by default as you used to have (the one with /a, /b, etc).
What do you think?

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
```diff
@@ -0,0 +1,87 @@
...
+        // EnumGene<String>.getValueAsPrintableString() always wraps the value in quotes regardless of
+        // mode or targetFormat. Using it would produce double-quoting in JSON ("path":""/x"") and
+        // wrong quotes inside XML tags (<path>"/x"</path>). getValueAsRawString() returns the
+        // bare string (e.g. /x), letting us control quoting per format ourselves.
+        val fields = when (this) {
+            is JsonPatchPathOnlyGene -> {
```
jgaleotti: I think it is possible to have this implementation without referring explicitly to each subclass using the ``children'' fields

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
```diff
@@ -0,0 +1,87 @@
...
+        previousGenes: List<Gene>,
+        mode: GeneUtils.EscapeMode?,
+        targetFormat: OutputFormat?,
+        extraCheck: Boolean
+    ): String {
+        // path/from fields use getValueAsRawString() instead of getValueAsPrintableString() because
+        // EnumGene<String>.getValueAsPrintableString() always wraps the value in quotes regardless of
+        // mode or targetFormat. Using it would produce double-quoting in JSON ("path":""/x"") and
+        // wrong quotes inside XML tags (<path>"/x"</path>). getValueAsRawString() returns the
+        // bare string (e.g. /x), letting us control quoting per format ourselves.
```
jgaleotti: This is strange since I am not seeing that this is actually happening when having enumgenes in an objectgene

    @Test
    fun testObjectGeneWithEnumGene() {
        val name = "com.example.MyDto"
        val schema = """
            "$name": {
                "type": "object",
                "properties": {
                    "path": {
                        "type": "string",
                        "enum": ["/some", "/some/path", "/some/other/path"]
                    }
                },
                "required": ["path"]
            }
        """.trimIndent()

        val gene = RestActionBuilderV3.createGeneForDTO(name, schema, name, RestActionBuilderV3.Options())

        assertTrue(gene is ObjectGene)
        val objectGene = gene as ObjectGene
        val field = objectGene.fields.find { it.name == "path" }
        assertNotNull(field)
        assertTrue(ParamUtil.getValueGene(field!!) is EnumGene<*>)
        val enumGene = ParamUtil.getValueGene(field) as EnumGene<String>
        assertEquals(3, enumGene.values.size)
        assertTrue(enumGene.values.containsAll(listOf("/some", "/some/path", "/some/other/path")))

        val json = objectGene.getValueAsPrintableString(listOf(), GeneUtils.EscapeMode.JSON, OutputFormat.JAVA_JUNIT_5)
        assertTrue(json == """{"path":"/some"}""" || json == """{"path":"/some/path"}""" || json == """{"path":"/some/other/path"}""")
    }

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchOperationGene.kt
suarezrominajulieta: you are rigth, this prints okey for json, adding the "" to the string printed (not double-quoting, that's wrong in my comment/possible implementation) => assertTrue(enumGene.values.containsAll(listOf("/some", "/some/path", "/some/other/path"))) ok.

but for xml, enumGene.getValueAsPrintableString prints the same as json, with the simple quoting, resulting in, for ex.:` <operation><op>remove</op><path>"/some/path"</path></operation>`, which is not xml correct.


```
val xmlPrintable = enumGene.getValueAsPrintableString(mode = GeneUtils.EscapeMode.XML)
assertEquals("\"/age\"", xmlPrintable)
```

that's why i used raw, so for json i add the simple quoting and in xml, they don't appear. but maybe we should fix the mode printing in enumgene?

core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
```diff
@@ -225,7 +225,7 @@ class EnumGene<T : Comparable<T>>(
     override fun getValueAsPrintableString(previousGenes: List<Gene>, mode: GeneUtils.EscapeMode?, targetFormat: OutputFormat?, extraCheck: Boolean): String {
 
         val res = values[index]
-        if (res is String && !treatAsNotString) {
+        if (res is String && !treatAsNotString && mode != GeneUtils.EscapeMode.XML) {
```
jgaleotti, line 231: This will collide with master once the fix is merged, right?

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
suarezrominajulieta: hi, it's the same code as the other pr. it should not, right? if the enumGene fix pr passes first, this should be automatically ignored. and if this pr passses first, i shall erase the other pr. right?

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
+package org.evomaster.core.search.gene.builder
```
arcuri82: having a package `builder` just for this class is bit too much. should move into `jsonpatch` package. maybe in future can be refactored if we introduce more builder types

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+    const val MIN_SIZE = 1
+    const val DEFAULT_MAX_SIZE = 10
+
+    const val FIELD_PATH = "path"
+    const val FIELD_FROM = "from"
+    const val FIELD_VALUE = "value"
+
+    const val ENTRY_STRING = "entry_string"
+    private const val ENTRY_INT = "entry_int"
+    private const val ENTRY_BOOL = "entry_bool"
```
arcuri82: why is `ENTRY_BOOL` private, but `ENTRY_STRING` is public?

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+ * Builds a JSON Patch document gene (RFC 6902).
+ *
+ * When [resourceSchema] is provided, paths and value gene types are derived from the schema.
+ * Otherwise, a fixed deterministic default template is used: paths /a, /b, /c, /d with 
+ * String, Integer, and Boolean value types.
+ */
+object JsonPatchDocumentGeneBuilder {
+
+    const val MIN_SIZE = 1
+    const val DEFAULT_MAX_SIZE = 10
```
arcuri82, line 29: if those are public constants, should add some JavaDoc on each of them

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+    const val FIELD_VALUE = "value"
+
+    const val ENTRY_STRING = "entry_string"
+    private const val ENTRY_INT = "entry_int"
+    private const val ENTRY_BOOL = "entry_bool"
+
+    val DEFAULT_PATHS = listOf("/a", "/b", "/c", "/d")
+
+    /** A single patchable field extracted from a resource schema. */
+    internal data class SchemaField(val path: String, val gene: Gene)
```
arcuri82, line 59: add info on what `path` represent, and what is its format

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Walks [schema] recursively and returns one [SchemaField] per reachable leaf field,
+     * using JSON Pointer notation for paths (e.g. "/name", "/address/street").
```
arcuri82: is this RFC6901? clarify. also look at `JsonUtils` for difference between JSON Pointer and JSON Path

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * - [ObjectGene]: descends into each fixed field.
+     * - [OptionalGene]: unwraps and descends using the same path prefix.
```
arcuri82, line 66: what about `NullableGene` and `ChoiceGene`?

arcuri82: might leave a comment for now, to address (with test cases) at later stage/PR

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
```diff
@@ -0,0 +1,83 @@
...
+class JsonPatchDocumentGene private constructor(
+    name: String,
+    operationsArr: ArrayGene<ChoiceGene<JsonPatchOperationGene>>
+) : CompositeFixedGene(name, listOf(operationsArr)) {
+
+    constructor(name: String, resourceSchema: Gene? = null)
+            : this(name, JsonPatchDocumentGeneBuilder.buildOperationsArray(resourceSchema))
+
+    companion object {
+        val MIN_SIZE get() = JsonPatchDocumentGeneBuilder.MIN_SIZE
```
arcuri82: is it necessary to specify `get()` here? wouldn't the follow work? (or would it not compile?)

`val MIN_SIZE  = JsonPatchDocumentGeneBuilder.MIN_SIZE`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
suarezrominajulieta: Because it was imported by GeneSamplerTest, but I fixed it to make it consistent.

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
suarezrominajulieta: done

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
suarezrominajulieta: done

core/src/main/kotlin/org/evomaster/core/search/gene/jsonpatch/JsonPatchDocumentGene.kt
suarezrominajulieta: no, it wasn't, i fix it.

core/src/main/kotlin/org/evomaster/core/search/gene/builder/JsonPatchDocumentGeneBuilder.kt
suarezrominajulieta: ok, done

arcuri82 (approved):

PR #1530, opened 2026-04-29
Initial CQL Session Replacement

gonzalotguerrero:
# Description
Added minimal support and config for the tracking of CQL queries through CqlSession. At the moment, only one version of the execute method is supported (when the query is a string). A future iteration should provide support for other variants of this method.


core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1990,6 +1990,11 @@ class EMConfig {
             " on the JVM.")
     var instrumentMR_MONGO = true
 
+    @Cfg("Execute instrumentation for method replace with category CASSANDRA." +
+            " Note: this applies only for languages in which instrumentation is applied at runtime, like Java/Kotlin" +
+            " on the JVM.")
+    var instrumentMR_CASSANDRA = false
```
jgaleotti, line 1997: Add the @Experimental annotation to Cassandra-related configuration keys

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
gonzalotguerrero: Done, thanks!

jgaleotti: @experimental goes before any other annotation in the method

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
gonzalotguerrero: My bad, thanks!!

jgaleotti (approved):

arcuri82:
@gonzalotguerrero @jgaleotti  thx! it looked fine, i merged into a different internal branch to trigger CI #1540

PR #1531, opened 2026-05-01
Fixing metrics action so it turns Red when increasing warnings/errors

aschenzle:
Build will continue as usual, this is just adding a visual indicator to draw attention.


aschenzle:
Closing as this fails the build instead of just a visual hint as expected. Will reopen a new one once I have the solution.

PR #1532, opened 2026-05-03
http misleading create put

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -433,4 +436,58 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+        val putOperations = RestIndividualSelectorUtils.getAllActionDefinitions(actionDefinitions, HttpVerb.PUT)
+
+        putOperations.forEach { putOp ->
+
+            if (hasPhaseTimedOut()) return
+
+            // template: an existing PUT 201 on this path (its body is known valid)
+            val putTemplate = RestIndividualSelectorUtils.findIndividuals(
+                individualsInSolution, HttpVerb.PUT, putOp.path, status = 201
+            ).flatMap { ei ->
```
arcuri82: we just need first, and iterating over all actions in all individuals might be bit expensive. this might be one of the cases where using lazy evaluation of stream might make sense, ie, use `asSequence()`. (Google `kotlin asSequence` if you are not familiar between eager and lazy evaluation of streams). most of time eager evaluation is fine,  apart from maybe cases like this one

arcuri82 (changes requested):

arcuri82 (approved):

PR #1536, opened 2026-05-04
Fix IndexOutOfBoundsException when actionResults is shorter than additionalInfoList

agusaldasoro:
## Summary
  - Add guard in `expandIndividual` to return early when `actionResults.size < additionalInfoList.size`, preventing a crash when fewer actions were executed than declared
  - Guard the `getlocation5xx` call in `handleResponseTargets` with a bounds check on `additionalInfoList`, fixing a crash on 500-status responses when `actionResults` is longer than
  `additionalInfoList`
  - Add regression tests in `AbstractRestFitnessTest` covering both crash scenarios

  ## Root cause
  `actionResults` can be shorter than `additionalInfoList` when not all HTTP actions are executed (e.g., SUT returns an error early). Both `expandIndividual` and `handleResponseTargets` iterated over
   `additionalInfoList.indices` or `actionResults.indices` without bounding the other list, causing `IndexOutOfBoundsException`.


core/src/main/kotlin/org/evomaster/core/problem/rest/service/fitness/AbstractRestFitness.kt
```diff
@@ -211,6 +211,11 @@ abstract class AbstractRestFitness : HttpWsFitness<RestIndividual>() {
             return
         }
 
+        if (actionResults.size < additionalInfoList.size) {
```
jgaleotti, line 221: could you add a comment to explain why actionResults can be smaller than the  additionalInfoList ?

jgaleotti (changes requested):

jgaleotti (approved):

arcuri82 (approved):

PR #1547, opened 2026-05-18
Java regex groups and their back references

lmasroca:
**IMPORTANT**: This PR adds a new gene `BackReferenceRxGene`.

Java regex added support for:
- Named capturing groups `(?<name>)`
- Capturing groups `()`
- Non-capturing groups `(?:)`
- Numbered back references `\1`,`\2`, etc.
- Named back references `\k<name>`

Numbered back references may be used to reference both named capturing groups (as each of them also get numbered) as well as plain capturing groups. Named back references may only be used to reference named capturing groups. To accomplish this, a new gene was added: `BackReferenceRxGene`.


core/src/test/kotlin/org/evomaster/core/search/gene/GeneSamplerForTests.kt
```diff
@@ -321,6 +321,7 @@ object GeneSamplerForTests {
                 .filter { !it.isAbstract }
                 .filter { it.java != CycleObjectGene::class.java && it.java !== LimitObjectGene::class.java }
                 .filter { it.java != ArrayGene::class.java && it.java != SqlMultidimensionalArrayGene::class.java }
+                .filter{ it.java != SqlPrimaryKeyGene::class.java }
```
lmasroca, line 324: Added this as `GeneRandomizedTest` failed because an `SqlPrimaryKeyGene` appeared in an `ArrayGene`. This gene is not globally valid unless it is mounted (at least one of its ancestors must be an `SqlAction`). There may be other genes that present this behavior, or other places this kind of genes might need to be excluded.

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
```diff
@@ -137,10 +150,9 @@ class GeneRegexJavaVisitor : RegexJavaBaseVisitor<VisitResult>(){
             }
 
             val resTerm = ctx.term()[i].accept(this)
-            val gene = resTerm.genes.firstOrNull()
 
-            if(gene != null) {
-                res.genes.add(gene)
+            if(resTerm.genes.isNotEmpty()) {
+                res.genes.addAll(resTerm.genes)
```
jgaleotti, line 166: It used to add only one gene, what happens if resTerm.size()>1

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
```diff
@@ -482,6 +528,47 @@ class GeneRegexJavaVisitor : RegexJavaBaseVisitor<VisitResult>(){
...
+            val allDigits = txt.drop(1)
+            val maxDigits = captureGroups.size.toString().length
+
+            // In Java, multi-digit back references interprets trailing digits literally, see more:
+            // https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#groupname:~:text=the%20parser%20will-,drop%20digits,-until%20the%20number
+            val backRefDigitCount = when {
+                maxDigits > allDigits.length -> allDigits.length
+                allDigits.take(maxDigits).toInt() <= captureGroups.size -> maxDigits
+                maxDigits > 1 -> maxDigits - 1
+                else -> throw IllegalStateException(
```
lmasroca, line 568: In Java, using a back ref that does not match an existing group (even taking into account digit dropping) does not result in an error. Instead, it results in a regex that cannot be matched by any string.
i.e.: regex "\1" has no matching strings.

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
lmasroca: Before these changes `resTerm.genes` could only have 0-1 genes. These are the parser rules for `term` which are handled in `visitTerm` (from where `resTerm` comes from):

`assertion` always results in 0 genes,
`atom` always results in 1 gene,
`atom quantifier` always results in 1 gene (quantifier wraps atom result).

With these changes now `atom` may have multiple genes, as `atomEscape` can be a `BackReference`, which might need to interpret some characters literally. As it gets tokenized as a single token we end up with a single visit which results in multiple genes. Therefore we must also consider this in `visitTerm`, as `visitAtom` now might have resulted in multiple genes.

So with these changes both `atom` and `atom quantifier` can now result in one or more genes instead of exactly one gene.

We could separate this case with an if so as to guarantee we are only changing the behavior for this particular case. Something like this:
```kotlin
            if (ctx.term()[i].atom()?.atomEscape()?.BackReference() != null){
                // if atom is a BackReference we addAll genes from result as there may be more than one if digits are dropped
                res.genes.addAll(resTerm.genes)
                // then the previous behavior:
            } else if (gene != null) {
                res.genes.add(gene)
            } else {
                 //...
            }
```

jgaleotti: ok. Please do this change and add more comments to explain why the conditions changed.

jgaleotti (changes requested):

jgaleotti (approved):

arcuri82 (approved):

PR #1550, opened 2026-05-19
Redis actions extension. Hash commands.

aszyrej:
Redis hash commands HGET and HGETALL are now included in the data generation process, generating RedisHsetActions.

Since RedisDbAction has now multiple concrete subclasses, the impact collection system was failing to match actions by class name across mutations. Added getImpactGroupKey() to Action — overridden in RedisDbAction to return a shared key for all its subtypes — and updated ImpactUtils and EvaluatedIndividual to use it instead of the raw class name.


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/redis/RedisHandler.java
```diff
@@ -106,18 +104,54 @@ public List<RedisCommandEvaluation> getEvaluatedRedisCommands() {
...
+        RedisCommand.RedisCommandType type = redisCommand.getType();
         if (distance > 0 &&
-            redisCommand.getType().equals(RedisCommand.RedisCommandType.GET)) {
+                type.equals(GET) || type.equals(KEYS) || type.equals(HGET) || type.equals(HGETALL)) {
             //For this first iteration we'll only work on GET commands.
             failedCommands.add(createFailedCommand(redisCommand));
         }
     }
 
     private RedisFailedCommand createFailedCommand(RedisCommand redisCommand) {
```
jgaleotti, line 115: I am a bit worried about this term. Is this actually a Failed Command or is it a actually a command that returns no results?

core/src/main/kotlin/org/evomaster/core/output/RedisWriter.kt
```diff
@@ -83,4 +89,10 @@ object RedisWriter {
 
         insertionVars.add(insertionVar to insertionVarResult)
     }
+
+    private fun escape(value: String, format: OutputFormat): String {
```
jgaleotti, line 103: Is there StringUtils for this escaping?

core/src/main/kotlin/org/evomaster/core/redis/RedisHsetAction.kt
```diff
@@ -0,0 +1,31 @@
+package org.evomaster.core.redis
+
+import org.evomaster.core.search.action.Action
+import org.evomaster.core.search.gene.Gene
+import org.evomaster.core.search.gene.string.StringGene
+
+/**
+ * Represents an HSET action, generated from a failed HGET or HGETALL command.
```
jgaleotti: what does failed mean? Again, I think "failed" might be confusing

core/src/main/kotlin/org/evomaster/core/redis/RedisSetAction.kt
```diff
@@ -0,0 +1,26 @@
+package org.evomaster.core.redis
+
+import org.evomaster.core.search.action.Action
+import org.evomaster.core.search.gene.Gene
+import org.evomaster.core.search.gene.string.StringGene
+
+/**
+ * Represents a SET action, generated from a failed GET command.
```
jgaleotti, line 8: failed?

jgaleotti (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/redis/RedisHandler.java
aszyrej: those are commands that returned no data. I associate this to the failedWhere for SQL. Is it not the same logic? As in "failed to return data"... Otherwise I'm not sure what other name we could use for this class.

core/src/main/kotlin/org/evomaster/core/output/RedisWriter.kt
aszyrej: I changed the deprecated library to the new one, but I'm still using StringEscapeUtils since it's similar for SQL and Mongo. A refactor for this should apply for all writers.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/redis/RedisHandler.java
jgaleotti: Let's keep the current name  because it requires a broader refactor of names. I think failedWhere is not either good as it seems as the command could not be executed, not that it was successfully executed and returned no rows.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/redis/RedisHandler.java
```diff
@@ -106,18 +104,54 @@ public List<RedisCommandEvaluation> getEvaluatedRedisCommands() {
...
+                    throw new IllegalArgumentException("Command HGET has invalid arguments.");
+                }
+                return new RedisFailedCommand(
+                        type.getLabel().toUpperCase(),
+                        args.get(0),
+                        null,
+                        args.get(1));
+            }
+            default:
+                throw new RuntimeException("Invalid command registering failed redis commands.");
```
jgaleotti: Add the type to the message. More generally, all exception messages should include context information to favour debugging.

core/src/main/kotlin/org/evomaster/core/redis/RedisSetAction.kt
jgaleotti: Clarify in all comments that by failing we mean "returning no data"

core/src/main/kotlin/org/evomaster/core/redis/RedisHsetAction.kt
jgaleotti: Clarify in comments that by failed it meant commands returning no data when executed.

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/output/RedisWriter.kt
```diff
@@ -40,15 +41,22 @@ object RedisWriter {
...
-                val escapedValue = StringEscapeUtils.escapeJava(rawValue).let { escaped ->
-                    // In Kotlin generated tests, dollar signs in string literals must be escaped
-                    if (format.isKotlin()) escaped.replace("$", "\\$") else escaped
+                val action = evaluated.redisAction
+
+                val dslCall = when (action) {
+                    is RedisSetAction -> {
+                        val key = escape(action.keyGene.getValueAsRawString(), format)
+                        val value = escape(action.valueGene.getValueAsRawString(), format)
+                        ".set(\"$key\", \"$value\")"
```
jgaleotti: getValueAsRawString()? why not using getValueAsPrintableString()?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/output/RedisWriter.kt
aszyrej: Using printable now.

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
```diff
@@ -18,18 +24,64 @@ object RedisInsertBuilder {
...
-            .map { cmd ->
-                // Only GET commands with a StringGene as value is supported at the moment.
-                // More complex types will be included in the future.
-                val valueGene = StringGene("value").also {
-                    it.randomize(randomness, false)
+            .filter { it.key == null || it.key !in existingKeys }
+            .mapNotNull { cmd ->
+                when (cmd.command) {
+                    "GET" -> {
+                        val keyGene = StringGene("key", cmd.key)
```
arcuri82: hi! bit confused here. is `keyGene` supposed to be a constant? or something that be modified during the search? In former case, shouldn't use a gene for it, but directly a `val` stirng in the action. In latter case, the value `cmd.key` will be lost as soon as the action is initialized

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
```diff
@@ -18,18 +24,64 @@ object RedisInsertBuilder {
...
+                        val valueGene = StringGene("value").also {
+                            it.randomize(randomness, false)
```
arcuri82: those calls to `it.randomize` might be redundant, as might be called again when intializing the action

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
```diff
@@ -18,18 +24,64 @@ object RedisInsertBuilder {
...
+                        // field is unknown — using a placeholder so the key exists in Redis
+                        RedisHsetAction(
+                            keyGene = keyGene,
+                            field = "field",
+                            valueGene = valueGene
+                        )
+                    }
+                    "KEYS" -> {
+                        val keyGene = StringGene("key").also {
+                            it.addSpecializations("key",
```
arcuri82: this is a bit confusing. specialzations is for Taint Analysis handling. if here you already know it is going to be a regex, why not using a `RegexGene` directly? Or is just a matter of typing in the action? note that specialization can be deactivated during the search

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
```diff
@@ -18,18 +24,64 @@ object RedisInsertBuilder {
...
+                                listOf(StringSpecializationInfo(StringSpecialization.REGEX_WHOLE, cmd.pattern)),
+                                randomness,
+                                updateGlobalInfo = false,  // should this be false?
```
arcuri82: yes. otherwise it crash, as you are adding a specialization to a gene that is not mount yet inside an Individual

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
```diff
@@ -18,18 +24,64 @@ object RedisInsertBuilder {
...
+                        }
+                        val valueGene = StringGene("value").also {
+                            it.randomize(randomness, false)
+                        }
+                        RedisSetAction(
+                            keyGene = keyGene,
+                            valueGene = valueGene
+                        )
+                    }
+                    else -> {
```
arcuri82: add a `LoggingUtils.uniqueWarn` here, as well as a `assert(false)`, to make it crash if it happens in any test case

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
aszyrej: Modified keys to be a val string since we're considering them as constant in the current scope.

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
aszyrej: I've refactored the InsertBuilder, and removed the randomize.

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
aszyrej: That was the initial approach but end up as a StringGene with specialization. I've generated another action for this, using RegexGene for the new key based on the pattern.

core/src/main/kotlin/org/evomaster/core/redis/RedisInsertBuilder.kt
aszyrej: Added!

arcuri82 (approved):

PR #1552, opened 2026-05-20
Testify/playwright

mkugelstad:
This PR introduce Playwright support in EvoMaster’s JavaScript test generator. The changes include:

- Implemented Playwright request generation.

- Fixed header serialization, request body handling, and test initialization issues in generated Playwright tests.

- Removed duplicate Playwright generator entries to ensure a single, consistent code path.

- Added Playwright dependencies to BlackBox test modules and updated BlackBoxUtils.kt with correct base/relative paths for Playwright execution.

- Multiple merges from master to keep the branch aligned with upstream changes.

All modifications should be isolated to Playwright‑specific code paths to avoid regressions in other generators.


core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
```diff
@@ -14,12 +14,15 @@ object BlackBoxUtils {
     private const val GENERATED_FOLDER_NAME = "generated"
 
     const val baseLocationForJavaScript = "$JS_BASE_PATH/$GENERATED_FOLDER_NAME"
+    const val baseLocationForPlaywright = "$JS_BASE_PATH/$GENERATED_FOLDER_NAME/playwright"
```
arcuri82: why the need for a new folder location instead of using same `baseLocationForJavaScript`? also, if really need new folder, should had written like:

`const val baseLocationForPlaywright = "$baseLocationForJavaScript/playwright"`

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
```diff
@@ -14,12 +14,15 @@ object BlackBoxUtils {
...
     const val baseLocationForPython = "$PY_BASE_PATH/$GENERATED_FOLDER_NAME"
     const val baseLocationForJava = "$MAVEN_BASE_PATH/src/test/java"
     const val baseLocationForKotlin = "$MAVEN_BASE_PATH/src/test/kotlin"
 
     fun relativePath(folderName: String) = "$GENERATED_FOLDER_NAME/$folderName"
 
+    fun relativePathPlaywright(folderName: String) = "$GENERATED_FOLDER_NAME/playwright/$folderName"
```
arcuri82: see previous comment. this function might not be needed

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
```diff
@@ -120,6 +122,20 @@ object BlackBoxUtils {
...
+    fun runPlaywrightTests(folderRelativePath: String) {
+        runNpmInstall()
+
+        val path = if(folderRelativePath.endsWith("/")){
+            folderRelativePath
+        } else {
+            "$folderRelativePath/"
+        }
+
+        val npx = if (isWindows()) "npx.cmd" else "npx"
```
arcuri82: we have `fun npm()` and `fun mvn()`. for consistency should introduce `fun npx()`

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
```diff
@@ -120,6 +122,20 @@ object BlackBoxUtils {
         runTestsCommand(command, JS_BASE_PATH, "NPM")
     }
 
+    fun runPlaywrightTests(folderRelativePath: String) {
```
arcuri82: why do you need this function and not using `runNpmTests()`?  aren't tests generated for `JS_JEST_PLAYWRIGHT` valid jest tests? why do they require to run `npx playwright` instead of using `npm test`? 
add comments to clarify and explain the reason

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
```diff
@@ -120,6 +122,20 @@ object BlackBoxUtils {
...
+        runNpmInstall()
+
+        val path = if(folderRelativePath.endsWith("/")){
+            folderRelativePath
+        } else {
+            "$folderRelativePath/"
+        }
```
arcuri82, line 122: code here seems duplicated from `runNpmTests()`. if indeed this function is needed (see previous comment) than should refactor (eg a single function with possible more parameters) to avoid code duplication

core-tests/e2e-tests/spring/spring-graphql-bb/src/test/kotlin/org/evomaster/e2etests/spring/graphql/bb/SpringTestBase.kt
```diff
@@ -90,6 +90,7 @@ abstract class SpringTestBase : GraphQLTestBase() {
         lambda: Consumer<MutableList<String>>
     ){
         val baseLocation = when {
+            outputFormat.isPlaywright() -> BlackBoxUtils.baseLocationForPlaywright
```
arcuri82: should avoid switches that are order dependent. for example, currently all formats that satisfy `isPlaywright()` also satisfy `isJavaScript()`. 
if indeed playwright here needs to be handle differently (see previous comments) should rather have something like:

```
outputFormat.isJavaScript() && outputFormat.isPlaywright() -> BlackBoxUtils.baseLocationForPlaywright
outputFormat.isJavaScript() && !outputFormat.isPlaywright()-> BlackBoxUtils.baseLocationForJavaScript
```

note that in future we might have palywright tests that are not related to JS

core-tests/e2e-tests/spring/spring-graphql-bb/src/test/kotlin/org/evomaster/e2etests/spring/graphql/bb/SpringTestBase.kt
```diff
@@ -100,6 +101,7 @@ abstract class SpringTestBase : GraphQLTestBase() {
     fun runGeneratedTests(outputFormat: OutputFormat, outputFolderName: String){
 
         when{
+            outputFormat.isPlaywright() -> BlackBoxUtils.runPlaywrightTests(BlackBoxUtils.relativePathPlaywright(outputFolderName))
```
arcuri82: see previous comment

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
```diff
@@ -160,6 +159,7 @@ abstract class SpringTestBase : RestTestBase() {
         lambda: Consumer<MutableList<String>>
     ){
         val baseLocation = when {
+            outputFormat.isPlaywright() -> BlackBoxUtils.baseLocationForPlaywright
```
arcuri82: see previous comments on use of `.isPlaywright()`

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
```diff
@@ -172,6 +172,7 @@ abstract class SpringTestBase : RestTestBase() {
     fun runGeneratedTests(outputFormat: OutputFormat, outputFolderName: String){
 
         when{
+            outputFormat.isPlaywright() -> BlackBoxUtils.runPlaywrightTests(BlackBoxUtils.relativePathPlaywright(outputFolderName))
```
arcuri82: see previous comments on use of `.isPlaywright()`

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
```diff
@@ -35,6 +35,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
+            <scope>provided</scope>
```
arcuri82: why this?

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
```diff
@@ -133,6 +134,16 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.14.1</version>
```
arcuri82: why this? also, it does not seem related to this PR's goals. if such change was needed for any reason, should had gone in its own separated PR.

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
```diff
@@ -133,6 +134,16 @@
...
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                            <version>1.18.30</version>
```
arcuri82: why hardcoding here a version of lombok that might diverge from the one declared for the dependency?

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -50,12 +50,15 @@ object CookieWriter {
             when {
                 format.isJava() -> lines.add("final Map<String,String> ${cookiesName(k)} = ")
                 format.isKotlin() -> lines.add("val ${cookiesName(k)} : Map<String,String> = ")
+                format.isPlaywright() -> lines.add("let ${cookiesName(k)};")
```
arcuri82: see previous comments on use of `.isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -50,12 +50,15 @@ object CookieWriter {
...
                 format.isJavaScript() -> lines.add("const ${cookiesName(k)} = ")
             }
 
             if (!format.isPython()) {
                 // TODO: should we use DTO for cookie related requests?
-                testCaseWriter.startRequest(lines)
+                if (!format.isPlaywright()) {
```
arcuri82: see previous comments on use of `.isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -76,12 +79,17 @@ object CookieWriter {
...
             }
 
-            if(format.isJavaScript()){
+            if(format.isJavaScript() && !format.isPlaywright()){
                 lines.add(".then((res) => res.headers['set-cookie'][0].split(';')[0])")
                 lines.add(".catch((err) => (err.status >= 300 && err.status <= 399) ? err.response.headers['set-cookie'][0].split(';')[0] : null)")
                 lines.appendSemicolon()
             }
 
+            if (format.isPlaywright()) {
```
arcuri82: should be nested inside a `if(format.isJavaScript()){`

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -110,7 +118,12 @@ object CookieWriter {
...
 
-        if(format.isJavaScript()) {
+        if (format.isPlaywright()) {
+            callEndpoint(lines, k, format, baseUrlOfSut)
+            return
+        }
+
+        if(format.isJavaScript() && !format.isPlaywright()) {
             callEndpoint(lines, k, format, baseUrlOfSut)
         }
```
arcuri82: avoid duplication, and possibly inverted boolean checks. for example, this might be better if refactored into:

```
if(format.isJavaScript() ) {
      callEndpoint(lines, k, format, baseUrlOfSut)
      if (format.isPlaywright()) {
             return
      }
}

```

with an explanation of why the need to return from that block

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -150,14 +167,17 @@ object CookieWriter {
         for(header in k.headers) {
             when {
                 format.isJavaOrKotlin() -> lines.add(".header(\"${header.name}\", \"${header.value}\")")
+                format.isPlaywright() -> {
+                    // handled in callEndpoint for Playwright
+                }
                 format.isJavaScript() -> lines.add(".set(\"${header.name}\", \"${header.value}\")")
```
arcuri82: recall previous comments on `isPlaywright() `

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -52,11 +52,16 @@ object TokenWriter {
             when {
                 format.isJava() -> lines.add("final String ${tokenName(k)} = ")
                 format.isKotlin() -> lines.add("val ${tokenName(k)} : String = ")
+                format.isPlaywright() -> {
+                    lines.add("let ${tokenName(k)};")
+                }
```
arcuri82: be consistent with existing code. if all other options are inlined with no `{}`, should do the same

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -52,11 +52,16 @@ object TokenWriter {
...
                 format.isJavaScript() -> lines.add("let ${tokenName(k)} = ")
             }
 
             when{
                 format.isJavaOrKotlin() -> lines.append("given()")
+                format.isPlaywright() -> {
+                }
```
arcuri82: if a block is empty, always add a comment to specify why so

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -343,9 +344,16 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
     private fun handleAssertionsOnField(value: Any?, flakyValue: Any?, lines: Lines, fieldPath: String, responseVariableName: String?) {
 
         if (value == null) {
+            val field = when {
+                format.isPlaywright() -> "(await $responseVariableName.json())"
```
arcuri82: recall previous comments on `isPlaywright() `

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -395,10 +403,18 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
 
             if (isSuitableToPrint(toPrint)) {
                 if (format.isJavaScript() || format.isPython()) {
+                    val field = when {
+                        format.isPlaywright() -> "(await $responseVariableName.json())"
```
arcuri82: recall previous comments on `isPlaywright() `

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -578,8 +597,15 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
                 val path = if (fieldPath.isEmpty()) "" else "$fieldPath."
                 ".body(\"${path}size()\", equalTo($expectedSize))"
             }
-            format.isJavaScript() ->
-                "expect($responseVariableName.body$fieldPath.length).toBe($expectedSize);"
+            format.isJavaScript() -> {
+                if (format.isPlaywright()) {
+                    val field = if (fieldPath.isEmpty()) "" else if (fieldPath.startsWith("[")) fieldPath else if (fieldPath.startsWith(".")) fieldPath else ".$fieldPath"
```
arcuri82: the check here seem similar (same?) to what done on line 352. avoid duplication. refactor into a private method

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -201,7 +201,7 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
         when {
             format.isJava() -> lines.add("String $name = ")
             format.isKotlin() -> lines.add("val $name : String? = ")
-            format.isJavaScript() -> lines.add("const $name = ")
+            format.isJavaScript() || format.isPlaywright() -> lines.add("const $name = ")
```
arcuri82: `||` here is problematic. in future we can have Playwright in other contexts. eg controlling browser for web testing from a Java JUnit test case

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -413,7 +422,11 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
                         lines.add("assertTrue(isValidURIorEmpty($location));")
                     }
                     format.isJavaScript() -> {
-                        lines.add("const $location = $resVarName.header['location'];")
+                        if (format.isPlaywright()) {
+                            lines.add("const $location = $resVarName.headers()['location'];")
+                        } else {
+                            lines.add("const $location = $resVarName.header['location'];")
```
arcuri82: avoid duplication. This could be written as:

```
lines.add("const $location = $resVarName")

if (format.isPlaywright()) {
      lines.append(".headers()['location'];")
} else {
      lines.append(".header['location'];")
}
```

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -445,7 +458,7 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
                 val extract = extractValueFromJsonResponse(resVarName, idPointer)
 
                 when{
-                    format.isJavaScript() -> lines.add("const ")
+                    format.isJavaScript() || format.isPlaywright() -> lines.add("const ")
                     format.isJava() -> lines.add("String ")
```
arcuri82: `||` is problematic

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -138,7 +139,7 @@ abstract class TestCaseWriter {
             lines.add("}")
         }
 
-        if (format.isJavaScript()) {
+        if (format.isJavaScript()) { // add to code block - ok for playwright?
```
arcuri82: is the comment still valid?

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -150,14 +167,17 @@ object CookieWriter {
...
                 format.isPython() -> {
                     lines.add("headers[\"${header.name}\"] = \"${header.value}\"")
                 }
             }
         }
 
-        if (format.isJavaScript()){
+        if (format.isJavaScript() && !format.isPlaywright()){
```
arcuri82: are you making sure of disabling redirect in Playwright? if so, could leave brief comment here stating where it is done

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -548,7 +552,8 @@ class TestSuiteWriter {
             if (controllerName != null) {
                 lines.add("const $controllerName = require(\"${config.jsControllerPath}\");")
             }
-            if (config.testTimeout > 0) {
+
+            if (config.testTimeout > 0 && !format.isPlaywright()) {
                 lines.add("jest.setTimeout(${config.testTimeout * 1000});")
```
arcuri82, line 560: aren't JS Playwright tests using Jest?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -740,7 +745,11 @@ class TestSuiteWriter {
                 lines.add("const $controller = new $controllerName();")
                 lines.add("let $baseUrlOfSut;")
             } else {
-                lines.add("const $baseUrlOfSut = \"${BlackBoxUtils.targetUrl(config, sampler)}\";")
+                if (config.outputFormat.isPlaywright()) {
+                    lines.add("const $baseUrlOfSut = \"${BlackBoxUtils.targetUrl(config, sampler)}\";")
+                } else {
+                    lines.add("const $baseUrlOfSut = \"${BlackBoxUtils.targetUrl(config, sampler)}\";")
```
arcuri82: aren't these 2 branches doing the same?

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
```diff
@@ -28,6 +29,12 @@ enum class OutputFormat {
 
     fun isJavaScript() = this.name.startsWith("js_", true)
 
+    /**
+     *  Return true if the output format is Playwright.
+     *  Playwright is currently only supported for JavaScript (or TypeScript).
+     */
+    fun isPlaywright() = this.name.endsWith("_playwright", true) // Testing new playwright imp
```
arcuri82: is comment `// Testing new playwright` still relevant?

arcuri82 (changes requested):
thx for the PR! see my comments

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
mkugelstad: To isolate the playwright tests and avoid overwriting other tests. But if that's not required, I'll remove it. Or use your suggestion.

arcuri82: @mkugelstad  thx. that shouldn't be necessary, as anyway we do a `FileUtils.deleteDirectory(folder.toFile())` in `SpringTestBase` before each test case execution

mkugelstad: Removing baseLocationForPlaywright in new commit

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
mkugelstad: Removing relativePathPlaywright in new commit ref previous comment

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
mkugelstad: npx playwright test is the recommended command (ref Playwright documentation), e.g. when using this command there is no need for a script in package.json. However, I've refactored runNpmTests to include Plawyright, and included a playwright script in package.json. Tests locally look promising, so will commit and verify in CI.

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
mkugelstad: See previous comment

core-tests/e2e-tests/e2e-tests-utils/src/test/kotlin/org/evomaster/e2etests/utils/BlackBoxUtils.kt
mkugelstad: This is removed in new commit due to refactoring of runNpmTests.

core-tests/e2e-tests/spring/spring-graphql-bb/src/test/kotlin/org/evomaster/e2etests/spring/graphql/bb/SpringTestBase.kt
mkugelstad: Since baseLocationForPlaywright is removed, separate handling for Playwright is removed in new commit.

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
mkugelstad: Since baseLocationForPlaywright is removed, separate handling for Playwright is removed in new commit.

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
mkugelstad: New commit includes outputFormat.isPlaywright() still, but without runPlaywrightTests and relativePathPlaywright as these have been removed ref previous comments. Testing locally indicates slightly different handling is required for Playwright, i.e. isPlaywright = true'

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
mkugelstad: Removed in new commit

core-tests/e2e-tests/spring/spring-graphql-bb/src/test/kotlin/org/evomaster/e2etests/spring/graphql/bb/SpringTestBase.kt
mkugelstad: New commit includes outputFormat.isPlaywright() still, but without runPlaywrightTests and relativePathPlaywright as these have been removed ref previous comments. Testing locally indicates slightly different handling is required for Playwright, i.e. isPlaywright = true'

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
mkugelstad: I believe this is something that was automatically added (by Maven?) when I initially set up the project. My bad for including it in commit, removed in new commit.

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
mkugelstad: See previous comment re lombok

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
mkugelstad: See previous comment re lombok

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: Which previous comment are you refewrring to here?

arcuri82: hi @mkugelstad . i mean the `should avoid switches that are order dependent`.  Playwright is a tool/library, not a programming language. in future we might have Playwright tests in Java for example

mkugelstad: I see, thx. That makes sense. In new commit, Playwright is refactored to be nested inside Javascript instead, hope that's better.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: In new commit, Playwright is refactored to be nested inside Javascript.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: After refactoring it seems like this is not required for Playwright after all, hence new commit does not include this exemption.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: New commit includes suggested refactoring. If not returned, there will be two request calls generated for the same endpoint.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: In new commit, Playwright is refactored to be nested inside Javascript.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: Further investigation shows that Playwright methods such as request.get, request.fetch etc. follow redirects by default.Hence in new commit, callEndpoint has been updated with maxRedirects: 0. In addition, maxRedirects: 0 has also been added to the Playwright request options in HttpWsTestCaseWriter.kt and GraphQLTestCaseWriter.kt

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: In new commit {} is removed. Also gone through other changed code to ensure it's consistent.

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: Comment added in new commit

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: Playwright is nested inside Javascript branch in new commit

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: Playwright is nested inside Javascript in new commit

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: Private method created and called where relevant in new commit

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
mkugelstad: No it's not, removed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: In new commit, a new helper method is added in OutputFormat, and called here and in handleLocationHeader

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: See previous comment re new helper method

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: New commit includes this suggestion

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: Playwright has it's own test runner, Playwright Test. This is the recommended test runner, but it's possible to use a package called jest-playwright if required.

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: Thanks for spotting this, clearly an error. Removed in new commit.

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
```diff
@@ -133,6 +133,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+
```
arcuri82: remove this added empty line

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -45,6 +47,7 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
         insertionVars: MutableList<Pair<String, String>>,
         testName: String
     ) {
+        this.baseUrlOfSut = baseUrlOfSut
```
arcuri82: cannot have something like this. ie, passing an input to a function and temporarely save it to a local class variable. this variable should be passed around as a method input

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -103,6 +104,7 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
             when {
                 format.isKotlin() -> lines.append("val $resVarName: ValidatableResponse = ")
                 format.isJava() -> lines.append("ValidatableResponse $resVarName = ")
+                format.isPlaywright() -> lines.append("const $resVarName = ")
                 format.isJavaScript() -> lines.append("const $resVarName = ")
```
arcuri82, line 130: aren't those 2 doing the same? so, `format.isJavaScript() || format.isPlaywright()`?

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -217,14 +217,23 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
...
-        } else {
-            if (verb == "trace" && format.isJavaOrKotlin()) {
-                //currently, RestAssured does not have a trace() method
-                lines.add(".request(io.restassured.http.Method.TRACE, ")
-            } else {
-                lines.add(".$verb(")
+        when {
+            format.isPlaywright() -> {
+                val verbToUse = call.verb.name.lowercase()
+                if (verbToUse.uppercase() == "OPTIONS") {
```
arcuri82: you first put it `lowercase()`, and then in a check you put it in `uppercase()`? why not just checking agains `options` to avoid the `uppercase()`?

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -117,7 +117,8 @@ abstract class TestCaseWriter {
         when {
             format.isJava() -> lines.add("public void ${test.name}() throws Exception {")
             format.isKotlin() -> lines.add("fun ${test.name}()  {")
-            format.isJavaScript() -> lines.add("test(\"${test.name}\", async () => {")
+            format.isJavaScript() && !format.isPlaywright()-> lines.add("test(\"${test.name}\", async () => {")
+            format.isJavaScript() && format.isPlaywright() -> lines.add("test(\"${test.name}\", async ({ request }) => {")
```
arcuri82: is this `request` variable then referred inside the block? if so, should put it in a `const val` somewhere in a companion object to avoid duplication of string values for that variable name

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
arcuri82: add comment to specify that Playwright tests do not use Jest

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -852,6 +859,10 @@ class TestSuiteWriter {
                     addStatement("RestAssured.urlEncodingEnabled = false", lines)
                 }
 
+                if (format.isJavaScript() && !format.isPlaywright()) {
+                    addStatement("process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'", lines)
```
arcuri82: what is this? and why it is done in this PR for something that is not for Playwright? if needed, should be in a separate PR with an explanation

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
```diff
@@ -18,6 +18,7 @@ enum class OutputFormat {
     KOTLIN_JUNIT_4,
     KOTLIN_JUNIT_5,
     JS_JEST,
+    JS_JEST_PLAYWRIGHT,
```
arcuri82: if Playwright test output you are writing is not using Jest, then this variable name is misleading. should be rather called `JS_PLAYWRIGHT`, isn't it?

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
```diff
@@ -41,4 +44,6 @@ enum class OutputFormat {
 
     fun isPython() = this.name.startsWith("python_", true)
 
+    fun isJsBased() = isJavaScript() || isPlaywright()
```
arcuri82: this is confusing. remove this `isJsBased`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: New commit includes comment

core-tests/e2e-tests/spring/spring-rest-rsa/pom.xml
mkugelstad: Removed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: From lowercase to uppercase is replaced by a complete list of Playwright direct helpers for lowercase instead. Fallback solution of request.fetch for OPTIONS, TRACE etc. intact. May not be the best solution, but I have used input from various AI agents to ensure it should be a robust solution, and all BB tests pass.

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
mkugelstad: Yes, this is from initial setup. Fixed in new commit.

core/src/main/kotlin/org/evomaster/core/output/OutputFormat.kt
mkugelstad: isJsBased is removed in new commit. Implemented a local private helper in RestTestCaseWriter instead. Another alternative is to refactor to a util.

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
mkugelstad: New commit include a private helper for request that can be extended to include additional Playwright fixtures later (e.g., page).

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: Further testing shows that this is not required after all, so removed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: I am unable to run e.g. BBHttpsLocalhostEMTest for JS_JEST locally all of a sudden without this. I will do some more testing locally and in CI before concluding.

mkugelstad: I see disabling TLS verification for JS is now added in header function, I suspect I did not have this master version when executing the tests previously. BBHttpsLocalhostEMTest runs OK, both locally and in CI, without my proposed change.

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
```diff
@@ -172,6 +171,7 @@ abstract class SpringTestBase : RestTestBase() {
     fun runGeneratedTests(outputFormat: OutputFormat, outputFolderName: String){
 
         when{
+            outputFormat.isPlaywright() -> BlackBoxUtils.runNpmTests(BlackBoxUtils.relativePath(outputFolderName), true)
             outputFormat.isJavaScript() -> BlackBoxUtils.runNpmTests(BlackBoxUtils.relativePath(outputFolderName))
```
arcuri82: this is going to be problematic in the future, when we are going to have Palywright outside of JavaScript (eg, in Kotlin to control a browser for web GUI testing). this could rewritten as:
```
outputFormat.isJavaScript() -> BlackBoxUtils.runNpmTests(BlackBoxUtils.relativePath(outputFolderName), outputFormat.isPlaywright())
```

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
```diff
@@ -31,6 +31,68 @@ object AuthWriter {
         placeHolderResolver: PlaceHolderResolver?
     ) {
 
+        // Compute placeholder replacements for PLaywright (used for dynamic user data)
```
arcuri82: `PLaywright` -> `Playwright`

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
```diff
@@ -31,6 +31,68 @@ object AuthWriter {
...
+        if (format.isPlaywright()) {
```
arcuri82, line 35: quite a long block, 50-60 lines, could be refactored into a private function

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
```diff
@@ -31,6 +31,68 @@ object AuthWriter {
...
+
+                // body/payload
+                val contentType = k.contentType
+                if (contentType != null) {
+                    when (contentType) {
+                        ContentType.X_WWW_FORM_URLENCODED -> {
+                            if (replacements == null) {
+                                lines.add("data: \"${k.payload}\",")
+                            } else {
+                                lines.add("data: \"${k.payload}\"")
```
arcuri82: a lot of the code here seems duplicate of what already in this class? if so, common code should be refactored in reusable functions, where possible. we need to avoid code duplication whenever possible

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -77,9 +82,28 @@ object CookieWriter {
...
             }
 
-            if(format.isJavaScript()){
-                lines.add(".then((res) => res.headers['set-cookie'][0].split(';')[0])")
-                lines.add(".catch((err) => (err.status >= 300 && err.status <= 399) ? err.response.headers['set-cookie'][0].split(';')[0] : null)")
+            if (format.isJavaScript()) {
+
+                // SuperAgent/Playwright cookie extraction
+                if (format.isPlaywright()) {
+                    lines.add(".then((res) => {")
```
arcuri82, line 89: you can use `lines.indented{}` to get indentation in the generated tests. right now all those line would have same indentation, which is likely not so readable when generated in test outputs

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
```diff
@@ -77,9 +82,28 @@ object CookieWriter {
...
+                    lines.add("return setCookie.split('\\n')[0].split(';')[0];")
+                    lines.add("}")
+                    lines.add("return null;")
+                    lines.add("})")
+                } else {
+                    lines.add(".then((res) => res.headers['set-cookie'][0].split(';')[0])")
+                    lines.add(".catch((err) => (err.status >= 300 && err.status <= 399) ? err.response.headers['set-cookie'][0].split(';')[0] : null)")
+                }
+
+                if (format.isPlaywright()) {
```
arcuri82: this `if` statements look weird. you have an opening `if (format.isJavaScript()) {`, where inside there are 2 `if (format.isPlaywright())`. shouldn't just simplify with a single 
```
if (format.isJavaScript() && format.isPlaywright())
```
?
also, could create a function in format like:
```
fun isJavaScriptPlaywright() =  format.isJavaScript() && format.isPlaywright()
```

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -54,11 +54,13 @@ object TokenWriter {
             when {
                 format.isJava() -> lines.add("final String ${tokenName(k)} = ")
                 format.isKotlin() -> lines.add("val ${tokenName(k)} : String = ")
+                format.isPlaywright() -> lines.add("let ${tokenName(k)};")
```
arcuri82: see previous comments. maybe best to introduce an explicit `isJavaScriptPlaywright`. however, the `when` here would break if one day we change the default for JS. better to be explicit:
```
format.isJavaScript() && format.isPlaywright() -> lines.add("let ${tokenName(k)};")
format.isJavaScript() && !format.isPlaywright()-> lines.add("let ${tokenName(k)} = ")
```

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -54,11 +54,13 @@ object TokenWriter {
...
                 format.isJavaScript() -> lines.add("let ${tokenName(k)} = ")
             }
 
             when{
                 format.isJavaOrKotlin() -> lines.append("given()")
+                format.isPlaywright() -> testCaseWriter.startRequest(lines) //handled in HttpWsTestCaseWriter with startRequest(lines) and await request
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()` in `when` clauses

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -87,7 +89,9 @@ object TokenWriter {
 
             when(token.extractFrom){
                 TokenHandling.ExtractFrom.BODY -> {
-                    if (format.isJavaScript()) {
+                    if (format.isPlaywright()) {
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
```diff
@@ -102,7 +106,9 @@ object TokenWriter {
                 }
                 TokenHandling.ExtractFrom.HEADER -> {
                     val header = token.extractSelector
-                    if (format.isJavaScript()) {
+                    if (format.isPlaywright()) {
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -350,12 +350,38 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
         return text.replace("\$", "\\\$")
     }
 
+    private fun formatFieldPath(fieldPath: String): String {
```
arcuri82, line 365: what is a `fieldPath` in this context? add some `/** */` JavaDoc to this function

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
```diff
@@ -350,12 +350,38 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
...
+                    }
+                }
+                    else -> ""
+                }
+            val fieldWithDot = if (fieldPath.isEmpty() || fieldPath.startsWith("[")) fieldPath else if (fieldPath.startsWith(".")) fieldPath else ".$fieldPath"
             val instruction = when {
                 format.isJavaOrKotlin() -> ".body(\"${fieldPath}\", nullValue())"
-                format.isJavaScript() -> "expect($responseVariableName.body$fieldPath).toBe(null);"
+                format.isPlaywright() -> "expect(($field)$fieldWithDot).toBe(null);"
+                format.isJavaScript() -> "expect($field$fieldPath).toBe(null);" // ($field$)fieldPath
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/service/GraphQLTestCaseWriter.kt
```diff
@@ -52,6 +52,9 @@ class GraphQLTestCaseWriter : HttpWsTestCaseWriter() {
 
         when {
             format.isJavaOrKotlin() -> lines.add(".contentType(\"application/json\")")
+            format.isPlaywright() -> {
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -73,14 +73,15 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
...
+    protected fun expectsRejection(res: ActionResult) = format.isJavaScript() && !format.isPlaywright() && hasTimeoutFault(res)
 
     protected fun expectsAssertRaises(res: ActionResult) = format.isPython() && hasTimeoutFault(res)
 
     fun startRequest(lines: Lines){
         when {
             format.isJavaOrKotlin() -> lines.append("given()")
-            format.isJavaScript() -> lines.append("await superagent")
+            format.isPlaywright() -> lines.append("await request")
+            format.isJavaScript() && !format.isPlaywright() -> lines.append("await superagent")
```
arcuri82: see previous comments on `isJavaScript` and `isPlaywright()`

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -123,7 +124,7 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
             when {
                 format.isKotlin() -> lines.append("val $resVarName: ValidatableResponse = ")
                 format.isJava() -> lines.append("ValidatableResponse $resVarName = ")
-                format.isJavaScript() -> lines.append("const $resVarName = ")
+                format.isJavaScript() || format.isPlaywright() -> lines.append("const $resVarName = ")
```
arcuri82: isn't `isPlaywright()` redundant here? also, it will brake the day we use Playwright for web testing in other languages

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -251,7 +258,8 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
 
         val set = when {
             format.isJavaOrKotlin() -> "header"
-            format.isJavaScript() -> "set"
+            format.isJavaScript() && !format.isPlaywright()-> "set"
+            format.isPlaywright() -> "" // headers are handled in a map in the options object
```
arcuri82: ` format.isJavaScript() && format.isPlaywright() -> `

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -538,6 +586,27 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
 
             format.isJavaScript() -> {
                 lines.indent(2)
+                if (format.isPlaywright()) {
+                    handleVerbEndpoint(baseUrlOfSut, call, lines)
+                    lines.replaceInCurrent(Regex("\\)$"), "")
```
arcuri82, line 595: is this `"\\)$"` still needed after the recent refactoring for the handling of dynamic users? if so, add comment to explain why doing this here

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -576,7 +646,8 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
     fun sendBodyCommand(): String {
         return when {
             format.isJavaOrKotlin() -> "body"
-            format.isJavaScript() -> "send"
+            format.isJavaScript() && !format.isPlaywright() -> "send"
+            format.isPlaywright() -> "data"
```
arcuri82: `format.isJavaScript() &&`

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -881,6 +1027,8 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
             } else {
                 when {
                     format.isJavaOrKotlin() -> ".header(\"$name\", isEmptyOrNullString())"
+                    format.isPlaywright() ->
```
arcuri82: check and update all the `when` clauses in this PR, to make explicit check for both JS and Playwright. i ll not repeat this comment in all remaining `when`

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -35,6 +36,9 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
         private val log = LoggerFactory.getLogger(RestTestCaseWriter::class.java)
     }
 
+    // Local extension/helper to avoid depending on OutputFormat.isJsBased from elsewhere
+    private fun OutputFormat.isJsBasedLocal(): Boolean = this.isJavaScript() || this.isPlaywright()
```
arcuri82: based on previous comments, i do not understand this check. recall that `isPlaywright` in future might refer as well as Playwright used in different languages besides JS

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
```diff
@@ -201,7 +205,7 @@ class RestTestCaseWriter : HttpWsTestCaseWriter {
         when {
             format.isJava() -> lines.add("String $name = ")
             format.isKotlin() -> lines.add("val $name : String? = ")
-            format.isJavaScript() -> lines.add("const $name = ")
+            format.isJsBasedLocal() -> lines.add("const $name = ")
```
arcuri82: this could be simply `isJavaScript()`

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
arcuri82: unless GitHub is displaying me outdated code, the suggestion is not included.  you don't have the `lines.add("const $location = $resVarName")` before the if statement to avoid duplicating the const allocation code

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
arcuri82: this can just be `isJavaScript()`. the format `JS_PLAYWRIGHT` is matching for `isJavaScript()`

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
```diff
@@ -81,6 +81,13 @@ abstract class TestCaseWriter {
 
     protected abstract fun addTestCommentBlock(lines: Lines, test: TestCase)
 
+    /**
+     * Compute the list of Playwright fixtures to destructure in the test signature.
```
arcuri82: if this is a list, shouldn't be `: List<String> = listOf("request")` ?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -541,9 +541,16 @@ class TestSuiteWriter {
 
         if (format.isJavaScript()) {
             lines.add("process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = '0';")
+            if (format.isPlaywright()) {
+                // Guard: avoid executing Playwright tests when Jest is running in the same folder
+                lines.add("const isJest = typeof process !== 'undefined' && process.env && process.env.JEST_WORKER_ID;")
```
arcuri82: why is this necessary?  can you add some motivation as comments? why would we mix Jest and Playwright together if they are not compatible?

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
```diff
@@ -1110,6 +1139,14 @@ class TestSuiteWriter {
 
 
     private fun footer(lines: Lines) {
+        // Close Playwright guard if it was opened; add a Jest placeholder test in the else branch
+        if (config.outputFormat.isPlaywright()) {
+            lines.add("} else {")
+            lines.indented {
+                lines.add("test('placeholder - Playwright suite skipped under Jest', () => { expect(true).toBe(true); });")
```
arcuri82: explain this. i m bit a confused :)

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: Not sure what happened here but the suggestion is included in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: Removed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/RestTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Yes it's unnecessary, as it's also true for Playwright. Removed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: Refactored in new commit

core/src/main/kotlin/org/evomaster/core/output/service/GraphQLTestCaseWriter.kt
mkugelstad: Refactored in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Refactored in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
mkugelstad: Further tests show it's still needed. It turns await request.<verb>(url) into await request.<verb>(url, { ... })
• handleVerbEndpoint ends the call with ')'
• The final ')' is removed so we can append the options object (method, headers, body, etc.)
• Dynamic users only change what goes inside { ... }, not this call shape
Comment added in new commit.

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: If separate folders for Playwright tests (in \spring-rest-bb\javascript) is not preferred, then I believe having a guard in header (and footer) is required to avoid executing Playwright when running Jest. Currently, with no separate Playwright folders, and the regex for Jest in package.json, Jest will also get a match on Playwright. I believe this can result in failure, as Jest will attempt to import @playwright/test or generate zero tests.

arcuri82: hi @mkugelstad  we shouldn't make changes in `core` module that are needed only for some limitations in our E2E tests. should consider alternatives that just need to make changes in `spring-rest-bb`. for example, the regex in `package.json` could be changed, and/or tests could be created with different prefix (`--outputFilePrefix`) or suffix  (`--outputFileSuffix`)

mkugelstad: In new commit this is now fixed with outputFilePrefix in SpringTestBase.kt and with testPathIgnorePatterns in package.json. File name now contains _pw_ for Playwright tests, e.g. EvoMaster_pw_faults_Test.js. Ran all tests OK locally so seems to work. I hope this is a sufficient solution.

core/src/main/kotlin/org/evomaster/core/output/service/TestSuiteWriter.kt
mkugelstad: Removed in new commit, see comment for header

core/src/main/kotlin/org/evomaster/core/output/service/TestCaseWriter.kt
mkugelstad: It's not a list as the current usage is only for 'request' fixture, this is also reflected in convertToCompilableTestCode. The comment was to illustrate that it can be extended to include multiple fixtures when/if needed in the future, e.g. 'page', 'context' etc. In new commit I've updated comment to reflect a single fixture, please let me know if it's preferred to refactor to List<String> straight away.

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/SpringTestBase.kt
mkugelstad: Rewritten as suggested in new commit.

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
mkugelstad: Refactored into a private function in new commit

core/src/main/kotlin/org/evomaster/core/output/auth/AuthWriter.kt
mkugelstad: New commit includes refactoring of the new private Playwright function, three helpers and shared utilities.

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: Using lines.indented{} in new commit

core/src/main/kotlin/org/evomaster/core/output/auth/CookieWriter.kt
mkugelstad: In new commit, playwright if statements have been consolidated into one single if statement

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: New commit includes the explicit suggestion.

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/auth/TokenWriter.kt
mkugelstad: Fixed in new commit

core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
mkugelstad: JavaDoc included in new commit.

arcuri82 (approved):

PR #1553, opened 2026-05-21
asyncapi: per-field facets for the headers schema (M11-PR6)

LautaroPetaccio:
## Summary

- Extracts per-field facets (required / enum / const / pattern / min / max / length / format) from a message's `headers:` schema and emits matching JUnit assertions on the captured reply headers map.
- `ReplyEnvelope` now carries the decoded `Map<String, String>` headers; `kafkaAwaitReplyEnvelope` populates it from every record header (UTF-8).
- Six new test helpers mirror the payload-side vocabulary (`replyHeaderHas`, `replyHeaderText`, `replyHeaderNumber`, `replyHeaderTextLength`, `replyHeaderFormatMatches`, `replyHeaderPatternMatches`); format checks share a private `formatMatches` helper.

## Motivation

We already lift the AsyncAPI 3.0 `headers:` schema into a separate gene tree so the EA can mutate it independently of the payload, and the schema-derived target families (`ENUM_VALUE_USED`, `FIELD_BOUNDARY`, …) fire on the *input* side. But on the **observation** side only the payload carried per-field facet oracles — declared `required` / `enum` / `format` on response headers (auth tokens, tenant ids, tracing keys) were just gone after parse time. This PR closes that asymmetry without changing the wire-level transport.

## Test plan

- [x] `mvn -pl core test -Dtest='AsyncAPI*'` — 64 tests pass (62 existing + 2 new in `AsyncAPIHeaderFieldAssertionsBuilderTest`)
- [x] `mvn -pl core test -Dtest='*Writer*Test,*OutputFormat*'` — 64 tests pass (writer surface unchanged for non-AsyncAPI formats; per-field reply assertions still emit identically)
- [ ] Sweep on the validation SUTs to confirm any schema with a declared `headers:` block (e.g. `multitenant-orders`) lifts header facets into generated assertions


LautaroPetaccio:
Mis-routed — should target the fork, not upstream. Re-opening on LautaroPetaccio/EvoMaster.

PR #1554, opened 2026-05-21
http oracle invalid location

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -684,6 +684,53 @@ object HttpSemanticsOracle {
         return false
     }
 
+    /**
+     * Checks the invalid-location oracle:
+     *
+     *   POST|PUT /X  -> 2xx with Location header L
```
arcuri82: not just for POST/PUT, but any verb where response has a location header

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -684,6 +684,53 @@ object HttpSemanticsOracle {
...
+     *   GET      L   -> 404   (BUG: location does not point to an existing resource)
+     *
+     * Sequence checked: the last two main actions of the individual.
+     *  - second-to-last (creator) is POST or PUT, has a 2xx status, and its result has a
+     *    non-blank Location header.
+     *  - last is a GET bound to the creator's saved Location via [RestCallAction.usePreviousLocationId].
```
arcuri82: that part of the code is deemed to refactoring... as naming of some methods has evolved since their original implementation :(  eg, that one also includes linking without Location header

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -684,6 +684,53 @@ object HttpSemanticsOracle {
...
+    ): Boolean {
+
+        if (individual.size() < 2) return false
+
+        val actions = individual.seeMainExecutableActions()
+        val creator = actions[actions.size - 2]
+        val follow  = actions[actions.size - 1]
+
+        // creator must be a potential resource-creating verb (POST/PUT)
+        if (!creator.isPotentialActionForCreation()) return false
```
arcuri82: any verb is fine, although i guess 99% of time it would be a POST :)

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -684,6 +684,53 @@ object HttpSemanticsOracle {
...
+
+        // follow-up must be a GET, and it must actually be chained to the creator's Location
+        if (follow.verb != HttpVerb.GET) return false
+        val expectedLocId = try { creator.creationLocationId() } catch (e: Exception) { return false }
```
arcuri82: add a TODO stating this will need to be refactored once we fix that function.
at a certain point i need to start a major refactoring of the action bindings... and i can then update the code here by checking the TODOs

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -684,6 +684,53 @@ object HttpSemanticsOracle {
...
+        // same auth so a 404 cannot be confused with an authorization problem
+        if (creator.auth.isDifferentFrom(follow.auth)) return false
+
+        val resCreator = actionResults.find { it.sourceLocalId == creator.getLocalId() } as RestCallResult?
+            ?: return false
+        val resFollow  = actionResults.find { it.sourceLocalId == follow.getLocalId() } as RestCallResult?
+            ?: return false
+
+        // creator must have succeeded and returned a Location header
+        if (!StatusGroup.G_2xx.isInGroup(resCreator.getStatusCode())) return false
```
arcuri82: why succeed? if it fails, and still it returns a location header, we want to check the location header being correct :)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,75 @@ class HttpSemanticsService : TimeBoxedPhase{
             prepareEvaluateAndSave(ind)
         }
     }
+
+
+    /**
+     * HTTP_INVALID_LOCATION oracle: a POST/PUT that returns 2xx with a Location header
```
arcuri82: any verb, not just POST/PUT

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,75 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+     *   GET      L   -> oracle target: must NOT be 404
+     *
+     */
+    private fun invalidLocation() {
+
+        val creatorVerbs = listOf(HttpVerb.POST, HttpVerb.PUT)
+
+        for (verb in creatorVerbs) {
+
+            val creatorOps = RestIndividualSelectorUtils.getAllActionDefinitions(actionDefinitions, verb)
```
arcuri82: we should rather search in the archive, for each endpoint, the tests for which in response there is a Location header, eg, `HttpWsCallResult.getLocation()`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,75 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+
+            creatorOps.forEach { creatorOp ->
+
+                if (hasPhaseTimedOut()) return
+
+                // GET endpoint to follow up on: same path (e.g. PUT /x/{id} -> GET /x/{id})
+                // or the closest descendant (e.g. POST /x -> GET /x/{id})
+                val getDef = actionDefinitions.find { it.verb == HttpVerb.GET && it.path == creatorOp.path }
```
arcuri82: the GET to do should be based on the actual Location header. if the location header is faulty, with mistakes in it, then it would not match any endpoint in the schema. those are the actual cases in which we want to verify this oracle. we might need to create a special action (or special configuration) to handle this kind of custom GET request

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,75 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+                }.minByOrNull { it.individual.size() } ?: return@forEach
+
+                val ind = RestIndividualBuilder.sliceAllCallsInIndividualAfterAction(
+                    candidate, verb, creatorOp.path, statusGroup = StatusGroup.G_2xx
+                )
+
+                val creator = ind.seeMainExecutableActions().last()
+
+                // Build the follow-up GET and force Location chaining.
+                val getAction = getDef.copy() as RestCallAction
```
arcuri82: the GET should be as based in the Location header, which might use a relative URL or not, and/or have query parameters. most likely we need new action subtype to handle this. might also most likely impact how we output test cases

core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/main/kotlin/com/foo/rest/examples/spring/openapi/v3/httporacle/invalidlocation/base/HttpInvalidLocationApplication.kt
```diff
@@ -0,0 +1,52 @@
...
+    ): ResponseEntity<Any> {
+
+        val isNew = !data.containsKey(id)
+        data[id] = "$id"
+
+        // bug: Location header points to a different id than the one
+        // actually stored, so a follow-up GET on it will return 404
+
+        val status = if (isNew) 201 else 200
+        return ResponseEntity.status(status)
```
arcuri82, line 51: also add test case for when the relative path is invalid, eg, `/somePathThatDoesNotExist`, and also full URL, eg, `http://localhost:$port/api/resources/${id+1000}`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,85 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+     * Sequence built:
+     *   [...]
+     *   ANY /X  -> response with Location header L
+     *   GET  L  -> oracle target: must NOT be 404
+     *
+     */
+    private data class LocationCandidate(
+        val individual: EvaluatedIndividual<RestIndividual>,
+        val sourceIndex: Int,
+        val source: RestCallAction
```
arcuri82: if you have `individual` and `sourceIndex`, isn't `source` redundant?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,85 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+    /**
+     * Every action in [ei] whose response carried a non-blank Location header,
+     * paired with its index in [RestIndividual.seeMainExecutableActions].
+     */
+    private fun locationCandidatesIn(
+        ei: EvaluatedIndividual<RestIndividual>
+    ): Sequence<LocationCandidate> {
+        val mainActions = ei.individual.seeMainExecutableActions()
+        return ei.evaluatedMainActions().asSequence().mapNotNull { ea ->
+            val a = ea.action as? RestCallAction ?: return@mapNotNull null
```
arcuri82: if you need the `idx`, why are you using a stream instead of a for loop on the `indices` of `ei.evaluatedMainActions()`?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -552,4 +556,85 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+                // TODO: RestCallAction.creationLocationId() currently restricts location-id generation
+                //  to POST/PUT and throws otherwise, so this branch silently no-ops on other verbs.
+                //  After that restriction is refactored to allow any verb whose response carried a
+                //  Location header, this catch can be dropped and the oracle will fire for all verbs.
+                creator.saveAndLinkLocationTo(getAction)
+            } catch (e: IllegalArgumentException) {
+                continue
+            }
+
+            ind.addMainActionInEmptyEnterpriseGroup(-1, getAction)
```
arcuri82, line 641: what is the `-1` here for?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
omursahin: I added this comment: `// add getAction as a last operation`

arcuri82 (approved):

PR #1555, opened 2026-05-23
asyncapi: MassTransit envelope support for AMQP transport (M12-PR1)

LautaroPetaccio:
## Summary

- Adds an `EnvelopeStrategy` (`NONE` | `MASSTRANSIT`) to `AmqpBrokerClient`, auto-detected from the AsyncAPI document's `defaultContentType` field via the new `MessageBrokerClient.configureContentType(String?)` hook (default no-op; Kafka and WebSocket ignore it).
- When `defaultContentType: application/vnd.masstransit+json`, publishes flip to fanout-exchange routing (channel-as-exchange, routing-key `""`), wrap the inner payload in `{messageId, messageType: ["urn:message:<channel>"], message: <inner>}`, set the right Content-Type header, and the receive side declares the matching fanout exchange + binds a server-named queue + unwraps before returning the inner payload to the fitness layer. `NONE` mode is unchanged from M11-PR9.
- Also ships the orphaned M11-PR9 round-trip integration test (`AmqpBrokerClientRoundTripTest`) and its `org.testcontainers:rabbitmq` test dependency — these were authored alongside that PR but never made it into master. Five existing M11-PR9 test methods + the two new MassTransit ones in one file.

## Why now (M12 motivation)

The §29 schema-only sweep characterized what EvoMaster can explore from an AsyncAPI document alone (without any SUT in the loop) — useful as a schema-exploration ceiling, but not behaviorally meaningful. The §30 pivot brings SUTs into the search loop so the engine's already-implemented reply-observation machinery (`REPLY_RECEIVED`, `REPLY_SCHEMA_VALID`, `REPLY_VARIANT`, `OUTPUT_RECEIVED`, ... in `AbstractAsyncAPIFitness.handleSubscribeReply`/`handleSubscribeOutput`) can fire on real SUT responses instead of always timing out.

The first Path-A SUT is the bookworm trio (foxminchan/BookWorm — .NET 9 + Aspire + MassTransit-on-RabbitMQ). Without this PR every bookworm publish silently fails to dispatch in the .NET consumers, because MassTransit's deserializer rejects everything that isn't its envelope shape. With this PR, the §30 sweep can drive bookworm end-to-end.

## Test plan

- [x] `mvn -pl core test -Dtest='AmqpBrokerClientEnvelopeTest'` — 14/14 Level-1 unit tests pass (envelope wrap/unwrap, content-type detection, fallbacks, round trip). Runs without Docker.
- [x] `mvn -pl core test -Dtest='AmqpBrokerClientRoundTripTest'` — 7/7 Level-2 integration tests pass (5 existing M11-PR9 round-trip + 2 new MassTransit) against Testcontainers RabbitMQ.
- [x] `mvn -pl core test -Dtest='AsyncAPI*Test,*BrokerClient*Test'` — 112/112 broader AsyncAPI suite still passes (no regressions).
- [ ] Level-3 manual verification: deferred to the bookworm deploy task (#66). When bookworm comes up, a single test publish against its `BookWorm.Contracts:UserCheckedOutIntegrationEvent` exchange should result in observable consumer dispatch / reply on the corresponding response channel.

## Design notes (questions resolved during M12-PR1 design)

- **Document-level detection only.** Per-message `contentType` overrides are an M12 follow-up if a corpus member ever mixes envelopes within one schema. All seven bookworm schemas declare `defaultContentType` at document level; per-message override never actually occurs in the corpus.
- **Schema-layer field already existed.** `AsyncAPISchema.defaultContentType` was parsed by `AsyncAPIAccess` but no downstream code read it. No schema-layer plumbing was added — the field is now consumed by the broker layer via the new interface hook.
- **Both wrap and unwrap in the broker layer.** Keeps the fitness layer transport-neutral; `MessageBrokerClient`'s default no-op for `configureContentType` means Kafka/WebSocket don't need to know MassTransit exists.
- **No `--bbAmqpEnvelope` CLI flag added.** Document-level auto-detection covers the corpus; an explicit override is easy to add later (`var envelopeStrategy: EnvelopeStrategy` on the broker is already public for that reason — also useful for unit tests).
- **Defensive fallbacks.** Wrap of non-JSON inner payload writes the bytes as a string field; unwrap of non-envelope bytes returns them unchanged. The engine sees `DELIVERY_OK` rather than crashing mid-action when a SUT misbehaves.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LautaroPetaccio:
Wrong target — recreating against the fork (LautaroPetaccio/EvoMaster), per thesis workflow.

PR #1556, opened 2026-05-25
Fix enumGene's getValueAsPrintableString in xml mode

suarezrominajulieta:
We add this condition so, in xml mode, it prints without simple quoting.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
@@ -225,7 +225,7 @@ class EnumGene<T : Comparable<T>>(
     override fun getValueAsPrintableString(previousGenes: List<Gene>, mode: GeneUtils.EscapeMode?, targetFormat: OutputFormat?, extraCheck: Boolean): String {
 
         val res = values[index]
-        if (res is String && !treatAsNotString) {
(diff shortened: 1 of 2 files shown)


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
```diff
@@ -225,7 +225,7 @@ class EnumGene<T : Comparable<T>>(
     override fun getValueAsPrintableString(previousGenes: List<Gene>, mode: GeneUtils.EscapeMode?, targetFormat: OutputFormat?, extraCheck: Boolean): String {
 
         val res = values[index]
-        if (res is String && !treatAsNotString) {
+        if (res is String && !treatAsNotString && mode != GeneUtils.EscapeMode.XML) {
```
arcuri82, line 231: add a comment to specify why XML needs to be treated specially.

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/collection/EnumGene.kt
suarezrominajulieta: Hi @arcuri82, I added the comment, thanks!

arcuri82 (approved):

PR #1564, opened 2026-06-01
Java regex char class intersection, union (nesting) and subtraction.

lmasroca:
Java regex added support for character class:
- Intersection: `[a-z&&[a-o]&&[aeiou]]`,
- Subtraction: `[a-z0-9&&[^aeiou135-9]]`,
- Union: `[aeiou[A-Z]\d[b-d]z]`.


core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -189,15 +189,23 @@ patternCharacter
  | BRACE_close
  | BRACKET_close
  | COLON
+ | '&&'
```
jgaleotti: could we use double_ampersand instead of '&&' ?

jgaleotti (changes requested):

jgaleotti (approved):

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
```diff
@@ -189,15 +189,23 @@ patternCharacter
  | BRACE_close
  | BRACKET_close
  | COLON
+ | DOUBLE_AMPERSAND
```
arcuri82: is this specific only to Java? does JavaScript have it or not? if not, add a comment here. if yes, needs to be implemented :)

arcuri82 (changes requested):

core/src/main/antlr4/org/evomaster/core/parser/RegexJava.g4
lmasroca: By default JS does not support character class intersection. It is only supported if the "v" flag is turned on (which cannot be embedded in the regex). Currently we do not support any flags in JS, and some of them change the behavior of the regex engine dramatically.

Added a comment regarding this.

arcuri82 (approved):

PR #1565, opened 2026-06-04
Json Patch e2e

suarezrominajulieta:
We now add the logic to restAction to handle json patch request, and to read the schema from get, put or post endpoints in the same place from patch request, if there's any.

also, we add the black box test.


core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -118,6 +119,7 @@ class DtoWriter(
             gene is ObjectGene -> calculateDtoFromObject(gene, actionName)
             gene is ArrayGene<*> -> calculateDtoFromArray(gene, actionName)
             gene is FixedMapGene<*, *> -> calculateDtoFromFixedMapGene(gene, actionName)
+            gene is JsonPatchDocumentGene -> return
```
jgaleotti, line 126: Add TODO on JsonPatchDocumentGene written as DTO when the test case is written

core/src/main/kotlin/org/evomaster/core/problem/rest/builder/RestActionBuilderV3.kt
```diff
@@ -742,11 +739,28 @@ object RestActionBuilderV3 {
...
-        val deref = obj.schema.`$ref`?.let { ref -> SchemaUtils.getReferenceSchema(schemaHolder, currentSchema, ref, messages) } ?: obj.schema
-        val name = deref?.xml?.name ?: deref?.`$ref`?.substringAfterLast("/") ?: "body"
+        val isJsonPatch = verb == HttpVerb.PATCH && bodies.keys.any { it.contains("json-patch") }
 
-        var gene = getGene(name, obj.schema, schemaHolder,currentSchema, referenceClassDef = null, options = options, messages = messages, examples = examples)
+        val name: String
+        var gene: Gene
+        if (isJsonPatch) {
+            name = "body"
+            val patchResourceSchema = JsonPatchSchemaResolver.resolveResourceSchema(
```
jgaleotti, line 766: Add comment on what is the intended behaviour of this.

core/src/test/kotlin/org/evomaster/core/output/dto/DtoWriterTest.kt
```diff
@@ -95,6 +98,23 @@ class DtoWriterTest {
         assertEquals(dtoWriter.getCollectedDtos().size, 0)
     }
 
+    @Test
```
jgaleotti: this test case should be removed

jgaleotti (changes requested):

core-tests/e2e-tests/spring/spring-rest-bb/src/main/kotlin/com/foo/rest/examples/bb/jsonpatch/BBJsonPatchApplication.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration
+import org.evomaster.e2etests.utils.CoveredTargets
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.*
+
+@SpringBootApplication(exclude = [SecurityAutoConfiguration::class])
+@RestController
+@RequestMapping("/pets")
+open class BBJsonPatchApplication {
```
jgaleotti, line 15: Why blackbox and whitebox, I think we need only one of them for testing this functionality.

jgaleotti (changes requested):

jgaleotti (approved):

core-tests/e2e-tests/spring/spring-rest-bb/src/main/kotlin/com/foo/rest/examples/bb/jsonpatch/BBJsonPatchApplication.kt
```diff
@@ -0,0 +1,142 @@
...
+    @PatchMapping("/{id}", consumes = ["application/json-patch+json"])
+    fun patchPet(@PathVariable id: Long, @RequestBody body: String): ResponseEntity<String> {
+        if (parsePatchDocument(body) == null)
+            return ResponseEntity.badRequest().body("Patch document must be a JSON array")
+
+        CoveredTargets.cover("PATCHED")
+        return ResponseEntity.ok("patched")
+    }
+
+    @ApiResponses(value = [
```
arcuri82: why all these `@ApiResponses` documentation? if it is not used directly, then it is just noise for the tests (eg, make this class harder to read). Remove them unless they are necessary

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
```diff
@@ -118,6 +119,11 @@ class DtoWriter(
             gene is ObjectGene -> calculateDtoFromObject(gene, actionName)
             gene is ArrayGene<*> -> calculateDtoFromArray(gene, actionName)
             gene is FixedMapGene<*, *> -> calculateDtoFromFixedMapGene(gene, actionName)
+            // TODO: a JsonPatchDocumentGene is currently skipped from DTO collection. Once we decide
+            //  how a JSON Patch document should be rendered when a test case is written (it is not a
+            //  regular object/array DTO but an RFC 6902 array of operations), this should build and
+            //  emit the corresponding DTO instead of returning.
+            gene is JsonPatchDocumentGene -> return
```
arcuri82, line 126: these will break generated tests. somewhere, there is the option of not applying DTO for an action. you could check if action is using any JsonPatchDocumentGene, and uses string instead of DTO in those cases

core/src/test/kotlin/org/evomaster/core/problem/rest/builder/JsonPatchSchemaResolverTest.kt
```diff
@@ -0,0 +1,270 @@
...
+        schema.main.schemaParsed.paths[path]!!.patch,
+        schema,
+        schema.main,
+        messages
+    )
+
+    @Test
+    fun testResolveFromGetResponse() {
+        val schema = parse(minimalSpec("""
+            "/pets/{id}": {
```
arcuri82: this is fine. but, in future, it is best to have schema files under `resources` and read them in the tests. the point is that, once  in a file, it is easier to edit and validate them with an IDE.

arcuri82 (changes requested):

core-tests/e2e-tests/spring/spring-rest-bb/src/main/kotlin/com/foo/rest/examples/bb/jsonpatch/BBJsonPatchApplication.kt
suarezrominajulieta: done, i removed them 👍

core/src/test/kotlin/org/evomaster/core/problem/rest/builder/JsonPatchSchemaResolverTest.kt
suarezrominajulieta: done. i moved them to a file under resources 👍

core/src/main/kotlin/org/evomaster/core/output/dto/DtoWriter.kt
suarezrominajulieta: ok, i applied that check in HttpWsTestCaseWriter.kt, lines 131-133, is that alright?

core/src/main/kotlin/org/evomaster/core/output/service/HttpWsTestCaseWriter.kt
```diff
@@ -127,6 +128,9 @@ abstract class HttpWsTestCaseWriter : ApiTestCaseWriter() {
         val bodyParam = call.parameters.find { p -> p is BodyParam } as BodyParam?
         if (bodyParam != null && bodyParam.isJson() && payloadIsValidJson(bodyParam)) {
             val primaryGene = bodyParam.primaryGene()
+            if (primaryGene.getWrappedGene(JsonPatchDocumentGene::class.java) != null) {
```
arcuri82, line 131: fine for now. not in this PR, but make new branch PR in which you create the following E2E for an API that:
- has a POST to create a resource
- has a PATCH to manipulate such created resource (404 if no resource)
and then make sure using DTO (there is a EMConfig for it), so that, in same generates test, we have a DTO used for POST following a call on PATCH without DTO (ie Patch must use a string representation, at least for now). any case, test should not fail to compile

arcuri82 (approved):

PR #1573, opened 2026-06-06
Adding LLM Naming strategy

Pgarrett:
Test cases can now be named using an LLM, EM will have to be set up using LLM config. Naming strategy involves prompting the LLM stating it should behave as a software engineer specializing in test naming. The prompt will include:
- Programming language being used (in order to follow naming conventions)
- Max amount of characters the new name can have
- Already generated test case names (to avoid repetition)
- The test case source code

Upon LLM response, a simple validation takes place in which we verify the response does not contain any illegal character, whitespaces and if it honours the max amount of characters the test name should have. If any of those validations fails, the LLM is re-prompted to provide a new name which follows the guidelines.

No disambiguation phase is added since that's already handled when prompting the LLM for the test case name. 

Test case source code is generated at this stage only to provide the LLM with context. Since generating source code is just concatenating strings, we can accept the cost of it being called here and later on for the real test case writing.


core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+    private val outputFormat: OutputFormat,
+    private val llmService: LlmService,
+    maxTestCaseNameLength: Int,
+    private val testCaseWriter: TestCaseWriter
+) : NumberedTestCaseNamingStrategy(solution) {
+
+    private val log: Logger = LoggerFactory.getLogger(TestSuiteWriter::class.java)
+    private val generatedNames = mutableSetOf<String>()
+
+    private val baseUrlOfSut = "baseUrlOfSut"
```
arcuri82: this constant is hardcoded and repeated view places throughout codebase. shoudl be refactored in a single place

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+        generatedNames.add(newName)
+        return newName
+    }
+
+    // LLM is sometimes returning names as "\n\ntheNewName" so we need to fix that.
+    // Also, some tests were ending as "200__" so we remove any trailing underscores.
+    private fun sanitizeName(testName: String): String {
+        var safeName = TestWriterUtils.safeVariableName(testName.trim().replace("\n", ""))
+        while (safeName.endsWith("_")) {
+            safeName = safeName.substringBeforeLast("_")
```
arcuri82: removing `\n` is fine, as otherwise name is useless. but removing `_` is somethign that we are manually fixing in the choice the LLM made. i m unsure about this. maybe we should comment it out

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return safeName
+    }
+
+    private fun getNewName(test: TestCase): String {
+        val testLines = getTestSourceCode(test)
+        val targetLanguage = getTargetLanguage()
+        return llmService.chat("""
```
arcuri82: all prompts should be move to the the `Prompts` class

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+
+            - The suffix must follow the naming conventions of $targetLanguage (e.g. snake_case for Python, camelCase for Java).
+            - The suffix must be derived directly from the code: what unit is being exercised, under what conditions, and what outcome is asserted.
+            - The suffix should follow the pattern `<method/feature>_<condition>_<expectedOutcome>` or a close variant natural to $targetLanguage test frameworks. Pick whichever fits the test best.
+            - Use only information present in the test body — do not invent context.
+            - Be specific: prefer `createUser_duplicateEmail_throwsConflict` over `createUser_fails`.
+            - Do not include words like "test", "check", "verify", or "ensure" in the suffix.
+            - The suffix must not exceed $remainingNameChars characters.
+            - The suffix must not match any of the already assigned names listed below.
+            - Output only the suffix, nothing else.
```
arcuri82: this first part that is always the same should be put in a SystemMessage, to help caching

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+            ## Test case
+
+            $testLines
+
+        """.trimIndent())
+    }
+
+    // Just in case the LLM did not follow the directive of just giving the new name as output.
+    private fun promptReIterateName(): String {
+        return llmService.chat("Your previous response contained more than just the suffix. Output only the suffix, nothing else. No explanation, no punctuation, no extra text.")
```
arcuri82: text in `Prompts`

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
```diff
@@ -0,0 +1,138 @@
...
+            outputFormat.isKotlin() -> "Kotlin"
+            outputFormat.isJavaScript() -> "JavaScript"
+            outputFormat.isPython() -> "Python"
+            else -> throw IllegalStateException("Unrecognized language: $outputFormat")
+        }
+    }
+
+    private fun getTestSourceCode(test: TestCase): Lines {
+        return try {
+            if (outputFormat.isCsharp())
```
arcuri82: C# is deprecated. no need to handle it in new code

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
Pgarrett: Actually @arcuri82, the trailing `_` is being added by the `TestWriterUtils.safeVariableName`, that's why I removed it

core/src/main/kotlin/org/evomaster/core/llm/Prompts.kt
```diff
@@ -63,6 +99,15 @@ object Prompts {
         return Pair(VALUE_BASED_ON_NAME_FAILURE, error)
     }
 
+    fun getPromptForTestCaseName(targetLanguage: String, remainingNameChars: Int, generatedNames: MutableSet<String>, testLines: String): Pair<String,String>{
```
arcuri82, line 111: add some JavaDocs for this function. for exaple, what is `remainingNameChars`?

core/src/main/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategy.kt
arcuri82: should avoid hardcoded constants. in that case, the `"_"` in that function should be refactored, eg, as a secondary parameter `safeReplacement` with default input value `"_"`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1575, opened 2026-06-07
basic mock scaffolding for LLM use

arcuri82:


arcuri82:
hi @Pgarrett  i might merge this later tonight or more likely tomorrow. this will give you a way to mock responses in LLMs

core/src/main/kotlin/org/evomaster/core/llm/mock/MockChatModel.kt
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Main method that needs to be mocked.
+     * All other methods seem to call this one.
+     */
+    override fun chat(chatRequest: ChatRequest, options: ChatRequestOptions) : ChatResponse {
+
+        val text = chatRequest.messages().joinToString { it.toString() }
+
+        val res = responses.find { mock -> mock.matcher(text) }?.responseProducer(text)
+            ?: "Hei! You forgot to setup the mock for me"
```
Pgarrett, line 62: Shouldn't it just throw an exception? I understand the test will eventually fail due to the assertion, but to make it more transparent

core/src/main/kotlin/org/evomaster/core/llm/LlmSupport.kt
```diff
@@ -76,6 +85,8 @@ object LlmSupport {
          */
 
         return when (provider) {
+            LlmProvider.MOCK -> MockChatModel()
```
Pgarrett, line 88: Should we have an `LlmMockService` as well? Or how can we inject the `EMConfig` in the [LlmService](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/llm/service/LlmService.kt#L21-L22) in order to use it in unit tests?

core/src/main/kotlin/org/evomaster/core/llm/mock/MockChatModel.kt
arcuri82: hmmm, i see your point. but not sure. i mean, we might have tests where using LLM but don't care of responses, so would not want to specify mocks manually for those. but i guess we could also have a boole parameter to decide what to do in case of no match (eg response as now or throw exception), to make the mock more configurable

core/src/main/kotlin/org/evomaster/core/llm/LlmSupport.kt
arcuri82:
```
        val injector: Injector = LifecycleInjector.builder()
                .withModules(* arrayOf<Module>(BaseModule()))
                .build().createInjector()
                
                injector.getInstance(LlmService::class.java)
 ```               
 
should avoid mocking when can use the real thing

PR #1577, opened 2026-06-08
Java regex unsatisfiable alternatives support

lmasroca:
Added support for satisfiable java regex with unsatisfiable pieces. This allows for things like `[a&&b]` which are unsatisfiable to appear inside regexes which have another satisfiable alternative, for example: `a|[a&&b]|c`, this would be able to sample the same values as `a|c` instead of throwing.


core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
```diff
@@ -47,7 +47,7 @@ class GeneRegexJavaVisitor : RegexJavaBaseVisitor<VisitResult>(){
      * Same as [captureGroups] but for named backreferences, which can be accessed
      * with their name or number.
      */
-    private val namedCaptureGroups = mutableMapOf<String, DisjunctionListRxGene>()
+    private val namedCaptureGroups = mutableMapOf<String, DisjunctionListRxGene?>()
```
jgaleotti, line 58: What does a null DisjunctionListRxGene represent?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
lmasroca: Since `DisjunctionListRxGene` extends `CompositeFixedGene` it cannot have an empty `disjunction` value. Because of that we cannot represent an unsatisfiable regex with `DisjunctionListRxGene`.

As the captured groups (named or numbered) may be unsatisfiable, in which case no `DisjunctionListRxGene` is created, so we use a null value to represent that case. Then if the group is referenced we know the back reference is unsatisfiable. `BackReferenceRxGene`s may be created with null `DisjunctionListRxGene` to represent this case. This is one of the possible ways to generate unsatisfiable regex, the other ways being character classes and invalid back references (`\1` when no groups are present).

This way we can accept things like `([a&&b])*c\1?` by sampling `c`, as unsatisfiable regex may be recovered with quantifies that accept 0 template repetitions.

jgaleotti: could you please add a comment explaining why the value could be null in this map?

jgaleotti (changes requested):

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -30,14 +30,16 @@ class CharacterRangeRxGene(
         private val log = LoggerFactory.getLogger(CharacterRangeRxGene::class.java)
     }
 
-    var value : Char = validRanges[0].start
+    var value : Char = if (isUnsatisfiable()) '\u0000' else validRanges[0].start
```
arcuri82, line 36: explain, in a comment, the use of `'\u0000` here

arcuri82 (changes requested):

arcuri82 (approved):

PR #1581, opened 2026-06-10
fix feature-service bug

yu-han-yang:


yu-han-yang:
@man-zhang

whytomato:
## Problem

`SqlNameContext` could not correctly process aliases assigned to parenthesized subqueries, particularly when the subquery contained a set operation such as `UNION ALL`.

For example:

```sql
SELECT derived.id
FROM (
    SELECT id FROM constraint_requires
    UNION ALL
    SELECT id FROM constraint_excludes
) derived
WHERE derived.id = 2
```

The previous implementation called ParenthesedSelect#getPlainSelect(). However, a UNION query is represented by a SetOperationList, so getPlainSelect() can return null, resulting in a NullPointerException.

The alias was also read from the inner FROM item, although derived belongs to the enclosing ParenthesedSelect.

## Fix
The alias is now read directly from the ParenthesedSelect.

Parenthesized subqueries are represented as UNNAMED_TABLE, and their aliases are mapped to that value. This avoids incorrectly assuming that every parenthesized SELECT contains a PlainSelect and allows qualified columns such as derived.id to be resolved correctly.

PR #1584, opened 2026-06-15
Redis actions for SMEMBERS and SINTER

aszyrej:
New actions for set-type commands. 
New e2e tests for APIs with no endpoints for saving data.


client-java/controller/src/main/java/org/evomaster/client/java/controller/redis/dsl/RedisDsl.java
```diff
@@ -71,6 +71,23 @@ public RedisStatementDsl hset(String key, String field, String value) {
...
+    public RedisStatementDsl sadd(String key, String member) {
+        checkDsl();
+        if (key == null || key.isEmpty()) {
+            throw new IllegalArgumentException("Unspecified key");
+        }
+        if (member == null || member.isEmpty()) {
+            throw new IllegalArgumentException("Unspecified member");
+        }
+        RedisInsertionDto dto = new RedisInsertionDto();
+        dto.command = "SADD";
```
jgaleotti: replace "SADD" with constant

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/problem/enterprise/service/EnterpriseFitness.kt
```diff
@@ -228,9 +228,18 @@ abstract class EnterpriseFitness<T> : FitnessFunction<T>() where T : Individual
...
-            redisResults[index].setInsertExecutionResult(b)
+        if (results?.executionResults != null) {
+            var dtoIndex = 0
+            allRedisActions.forEachIndexed { actionIndex, action ->
+                // Now actions such as SaddFromSinter have multiple insertion dtos related.
+                val count = action.insertionsCount()
+                val success = (dtoIndex until dtoIndex + count).all {
+                    results.executionResults[it]
+                }
+                if (!success) println("FAILED insertion $actionIndex: ${action.getName()}")
```
arcuri82: do not use `println`. rather use a logger. might also want to consider if should also add a `assert(false)` here, eg, crash in tests but not in production if it happens

core/src/main/kotlin/org/evomaster/core/redis/RedisDbAction.kt
```diff
@@ -15,5 +15,7 @@ sealed class RedisDbAction : EnvironmentAction(listOf()) {
      */
     open fun getTargetKey(): String? = null
 
+    open fun insertionsCount(): Int = 1
```
arcuri82, line 22: add JavaDoc explaining what this method is for

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/enterprise/service/EnterpriseFitness.kt
aszyrej: added log.warn and assert(false)

core/src/main/kotlin/org/evomaster/core/redis/RedisDbAction.kt
aszyrej: Added.

arcuri82 (approved):

PR #1586, opened 2026-06-16
Z3-based SQL data generation: memoization, explicit outcomes, statistics & fixes

agusaldasoro:
## Summary

Adds memoization and explicit outcome handling to EvoMaster's Z3-based SQL data generation (`generateSqlDataWithZ3`), plus optional statistics/correctness instrumentation and several encoding fixes.

- **Memoization:** caches solver results by `(sqlQuery, numberOfRows)` in a bounded LRU, skipping redundant Docker/Z3 executions for repeated queries. Only deterministic outcomes (SAT/UNSAT) are cached; UNKNOWN/ERROR are not (they may be transient). The cache stores the `Z3Result` (not the generated `SqlAction`s), so each individual still gets fresh, independently-mutable genes.
- **Explicit solver outcomes:** introduces `Z3Result` (`SAT` / `UNSAT` / `UNKNOWN` / `ERROR`) and `Z3Solution`, replacing `Optional<Map<String, SMTLibValue>>`. `Z3DockerExecutor` classifies Z3's stdout before parsing, so `unknown`/`unsat` are no longer misparsed as an empty SAT model.
- **Configurable Z3 timeout** (`sqlZ3TimeoutMs`, default 5000ms): passes a soft `-t:` timeout to Z3; on expiry Z3 returns `unknown` (mapped to `UNKNOWN`) instead of running unbounded.
- **Configurable rows** (`sqlZ3NumberOfRows`, default 1): number of rows generated per table.
- **Optional statistics** (`collectSqlZ3Stats`): per-run counters written to `statistics.csv` only when enabled — query volume & cache reuse, solver outcomes, and cost (Z3 time, SMT-LIB gen time/size).
- **Optional correctness measurement** (`measureSqlZ3Correctness`): after a SAT result, generated rows are re-evaluated with `SqlHeuristicsCalculator` to measure their distance to the original SQL predicate; mismatches are logged and recorded.
- **Encoding fixes:**
  - Composite primary keys now enforce **tuple** uniqueness (OR of per-column distinctness) instead of per-column distinctness, which over-constrained the encoding.
  - `SmtLibGenerator` supports `DELETE` and `UPDATE`, not just `SELECT`.
  - Integer columns map to `LongGene` instead of truncating via `.toInt()`.
  - Improved SQL parsing in `JSqlVisitor` (LOWER/UPPER, timestamp literals) and added `SmtLibGeneratorTest` coverage.
- **Memory fix:** the solver holds `Statistics` via a `WeakReference`, so its `@PreDestroy` registration (a GC root through Governator's predestroy-monitor thread) no longer pins `Statistics → Archive → individuals`, which was OOMing test suites that build many injectors.

All new options are `@Experimental` and default-off (or default 1 / 5000), so default runs are unaffected.

## Configuration

| Flag | Default | Purpose |
|---|---|---|
| `generateSqlDataWithZ3` | false | Enable Z3-based SQL data generation |
| `sqlZ3NumberOfRows` | 1 | Rows generated per table |
| `sqlZ3TimeoutMs` | 5000 | Soft per-query Z3 timeout in ms (0 = off) |
| `collectSqlZ3Stats` | false | Emit solver statistics to CSV |
| `measureSqlZ3Correctness` | false | Post-verify generated rows against the SQL predicate |

## Statistics (`collectSqlZ3Stats`)

- **Query volume / cache:** `sqlZ3QueriesSeen`, `sqlZ3UniqueQueries`, `sqlZ3DuplicateQueries`, `sqlZ3CacheHits`, `sqlZ3CacheMisses` — invariant: `sqlZ3QueriesSeen = sqlZ3CacheHits + sqlZ3CacheMisses`.
- **Outcomes:** `sqlZ3Sat`, `sqlZ3Unsat`, `sqlZ3Unknown`, `sqlZ3Errors`, `sqlZ3ParseFailures` — sum equals `sqlZ3CacheMisses`.
- **Cost:** `sqlZ3TotalMs`, `sqlZ3SmtlibGenTotalMs`, `sqlZ3AvgSmtlibSizeBytes`.

## Correctness (`measureSqlZ3Correctness`)

`sqlZ3CorrectnessChecks`, `sqlZ3CorrectnessZeroDistance`, `sqlZ3CorrectnessNonZero`, `sqlZ3CorrectnessAvgDist`, `sqlZ3CorrectnessEvalFailures`.

## Known limitations (documented in code as `KNOWN LIMITATION` / `FUTURE WORK`)

- JOIN `ON` uses diagonal row pairing (row `i` ↔ row `i`) and only the first `ON` conjunct — sufficient at `k=1`, not full INNER JOIN semantics.
- Composite foreign keys are matched per-column, not as a tuple.
- NULL conditions (`IS NULL` / `IS NOT NULL` / `= NULL`) are dropped; all columns are treated as `NOT NULL`.
- `NUMERIC` maps to `Int` (truncating decimals), inconsistent with `DECIMAL → Real`.
- Three type-name vocabularies interpret `ColumnDto.type` independently and could drift on variant spellings.


jgaleotti:
Is this ready for review? It seems to be failing...

agusaldasoro:
> Is this ready for review? It seems to be failing...


@jgaleotti The changes are ready and working, but there seems to be an out of memory issue that I can not solve in the Core and CoreIT tests

agusaldasoro:
@jgaleotti Hey! It's ready for review now, thanks!

core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlVisitor.java
```diff
@@ -45,7 +49,14 @@ public void visit(NullValue nullValue) {
 
     @Override
     public void visit(Function function) {
-        // TODO This translation should be implemented
+        String name = function.getName().toUpperCase();
+        if ((name.equals("LOWER") || name.equals("UPPER"))
```
jgaleotti: replace "LOWER" and "UPPER" with constants

core-extra/solver/src/main/java/org/evomaster/solver/Z3Result.java
```diff
@@ -0,0 +1,46 @@
+package org.evomaster.solver;
+
+import org.evomaster.solver.smtlib.value.SMTLibValue;
+
+import java.util.Map;
+
+/**
+ * Represents the outcome of a Z3 solver invocation.
+ * Distinguishes between three possible states: SAT (satisfiable with a model),
+ * UNSAT (unsatisfiable), and ERROR (solver or parsing failure).
```
jgaleotti: what about timeout? is that an outcome?

jgaleotti: also, can Z3 return UNKNOWN?

core-extra/solver/src/main/java/org/evomaster/solver/Z3Result.java
```diff
@@ -0,0 +1,46 @@
...
+        ERROR
+    }
+
+    public final Status status;
+    /** Non-null only when status == SAT. */
+    public final Map<String, SMTLibValue> model;
+    /** Non-null only when status == ERROR. */
+    public final String errorMessage;
+
+    private Z3Result(Status status, Map<String, SMTLibValue> model, String errorMessage) {
```
jgaleotti: create a abstract class Z3Solution and rename model to solution. (Using `model` can be a bit misleading nowadays due to LLMs...)

core-extra/solver/src/main/java/org/evomaster/solver/Z3Result.java
```diff
@@ -0,0 +1,46 @@
...
+    public enum Status {
+        /** Z3 found a satisfying assignment. The model is available. */
+        SAT,
+        /** The problem is unsatisfiable. No model exists. */
+        UNSAT,
+        /** A solver, I/O, or parsing error occurred. */
+        ERROR
+    }
+
+    public final Status status;
```
jgaleotti: why not getters and private fields?

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -86,6 +86,25 @@ class Statistics : SearchListener {
     private var sqlHeuristicEvaluationFailureCount = 0;
     private val sqlRowsAverageCalculator = IncrementalAverage()
 
+    // DSE (Dynamic Symbolic Execution) statistics
+    private var dseTotalQueriesProcessed = 0
```
jgaleotti: To be honest, I need we need a more suitable term sonce there is no "dse/dynamic symbolic execution" involved in this work :) 
What do you think of generateSqlDataWithDSE -> generateSqlDataWithZ3 ? Regarding the "dse..." it could be prefixed to "sqlZ3..." ?

core/src/test/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolverTest.kt
```diff
@@ -66,15 +66,15 @@ class SMTLibZ3DbConstraintSolverTest {
                 "ID" -> {
                     assertTrue(gene is SqlPrimaryKeyGene)
                     val child = gene.getViewOfChildren().first()
-                    assertEquals(4, (child as IntegerGene).value)
+                    assertEquals(4L, (child as LongGene).value)
                 }
                 "NAME" -> {
                     assertTrue(gene is StringGene)
                     assertEquals("agus", (gene as StringGene).value)
```
jgaleotti: "Alice" and "Bob" please :)

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
agusaldasoro: Done!

core-extra/solver/src/main/java/org/evomaster/solver/Z3Result.java
agusaldasoro: Done!

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -347,8 +347,8 @@ abstract class ApiWsStructureMutator : StructureMutator() {
             return handleSearch(ind, sampler, mutatedGenes, fw)
         }
         
-        if (config.generateSqlDataWithDSE) {
-            return handleDSE(ind, sampler, failedWhereQueries)
+        if (config.generateSqlDataWithZ3) {
+            return handleZ3(ind, sampler, failedWhereQueries)
```
jgaleotti: generateSqlDataWithZ3(...)

core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -347,8 +347,8 @@ abstract class ApiWsStructureMutator : StructureMutator() {
             return handleSearch(ind, sampler, mutatedGenes, fw)
```
jgaleotti: generateSqlDataWithSearch(...)

core-tests/e2e-tests/spring/spring-rest-h2-z3solver/src/test/java/org/evomaster/e2etests/spring/h2/z3solver/Z3SolverEMTest.java
```diff
@@ -31,7 +31,7 @@ public void testRunEM() throws Throwable {
                     args.add("true");
```
jgaleotti: replace args.add with setOption(args,"generateSqlDataWithSearch","false")

core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlVisitor.java
```diff
@@ -599,7 +615,19 @@ public void visit(TimeKeyExpression timeKeyExpression) {
 
     @Override
     public void visit(DateTimeLiteralExpression dateTimeLiteralExpression) {
-        // TODO This translation should be implemented
+        if (dateTimeLiteralExpression.getType() == DateTimeLiteralExpression.DateTime.TIMESTAMP) {
+            String value = dateTimeLiteralExpression.getValue();
+            if (value.startsWith("'") && value.endsWith("'")) {
```
jgaleotti: replace "'" with constant

core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlVisitor.java
```diff
@@ -599,7 +615,19 @@ public void visit(TimeKeyExpression timeKeyExpression) {
...
+                value = value.substring(1, value.length() - 1);
+            }
+            // Treat the timestamp string as UTC to match the UTC-based decoder in
+            // SMTLibZ3DbConstraintSolver (LocalDateTime.ofInstant(..., UTC)).
+            long epochSeconds = java.time.LocalDateTime.parse(value,
+                    DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss"))
```
jgaleotti: replace "yyyy-MM-dd HH:mm:ss" with constant

core-extra/solver/src/main/java/org/evomaster/solver/Z3DockerExecutor.java
```diff
@@ -54,28 +54,61 @@ public Z3DockerExecutor(String resourcesFolder) {
...
+                return Z3Result.error("Z3 produced no output for file: " + fileName
+                        + " stderr: " + result.getStderr());
+            }
 
-            // Check if the solver returned any output
-            if (stdout == null || stdout.isEmpty()) {
-                String stderr = result.getStderr();
-                throw new RuntimeException("No result after solving file " + stderr);
+            String trimmed = stdout.trim();
+            if (trimmed.startsWith("unsat")) {
```
jgaleotti: replace "unsat" with constant

core-extra/solver/src/main/java/org/evomaster/solver/Z3DockerExecutor.java
```diff
@@ -54,28 +54,61 @@ public Z3DockerExecutor(String resourcesFolder) {
...
+                return Z3Result.unsat();
             }
+            // "unknown" means Z3 could not decide (e.g. incomplete theory or timeout).
+            // It must be handled explicitly: otherwise it would fall through and be parsed
+            // as an empty model, silently masquerading as SAT.
+            if (trimmed.startsWith("unknown")) {
```
jgaleotti: replace "unknown" with constant

core-extra/solver/src/main/java/org/evomaster/solver/Z3DockerExecutor.java
```diff
@@ -54,28 +54,61 @@ public Z3DockerExecutor(String resourcesFolder) {
...
+                return Z3Result.unknown();
+            }
+            if (!trimmed.startsWith("sat")) {
```
jgaleotti: replace "sat" with constant

core-extra/solver/src/main/java/org/evomaster/solver/Z3DockerExecutor.java
```diff
@@ -54,28 +54,61 @@ public Z3DockerExecutor(String resourcesFolder) {
...
+                return Z3Result.error("Unexpected Z3 output for file " + fileName + ": " + stdout);
+            }
+
+            Map<String, SMTLibValue> assignments = SMTResultParser.parseZ3Response(stdout);
```
jgaleotti: The parser should create a solution instead of a assignment map

core-extra/solver/src/main/java/org/evomaster/solver/Z3Solution.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * This is intentionally an abstract type (rather than a raw {@link Map}) so that
+ * alternative solution representations can be introduced without changing callers.
+ */
+public abstract class Z3Solution {
+
+    /**
+     * @return the assignments of this solution, keyed by variable/constant name.
+     */
+    public abstract Map<String, SMTLibValue> getAssignments();
```
jgaleotti: why is this abstract?

jgaleotti (changes requested):

core-extra/solver/src/main/java/org/evomaster/solver/smtlib/SMTResultParser.java
```diff
@@ -43,6 +50,9 @@ public static Map<String, SMTLibValue> parseZ3Response(String z3Response) {
         String[] lines = z3Response.split("\n");
 
         for (String line : lines) {
+            // Defensive: Z3DockerExecutor already classifies 'unsat'/'unknown' before invoking this
+            // parser, so in practice only 'sat' models reach here. This guard is kept as a safety net
+            // in case the parser is ever called directly with an unsat response.
             if (line.startsWith("unsat")) {
```
jgaleotti: replace "unsat" with constant value

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -368,21 +409,25 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
-                    LoggingUtil.getInfoLogger().warn("Could not translate WHERE clause to SMT-LIB, skipping: ${where}. Reason: ${e.message}")
+        if (where != null && defaultTable != null) {
+            try {
+                val condition = parser.parse(where.toString(), toDBType(schema.databaseType))
+                for (i in 1..numberOfRows) {
+                    val constraint = parseQueryCondition(tableAliases, defaultTable, condition, i)
+                    smt.addNode(constraint)
                 }
+            } catch (e: RuntimeException) {
+                LoggingUtil.getInfoLogger().warn("Could not translate WHERE clause to SMT-LIB, skipping: ${where}. Reason: ${e.message}")
```
jgaleotti, line 519: is this failure considered in the sqlZ3 stats?

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -86,18 +142,113 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
+                     * that the generated rows satisfy insertion preconditions such as FK constraints
+                     * or NOT NULL columns that Z3 SQL generation currently leaves unconstrained.
+                     *
+                     * Note: SqlHeuristicsCalculator is SELECT-oriented. For DELETE/UPDATE the distance
+                     * computation may fail and be reported as an evaluation failure (sqlDistanceEvaluationFailure)
+                     * rather than a real distance; such failures are counted separately and are excluded
+                     * from the average, so they do not distort the correctness metric.
+                     */
+                    val distResult = computeCorrectnessDistance(sqlQuery, schemaDto, sqlActions)
+                    if (distResult.sqlDistanceEvaluationFailure) {
```
jgaleotti: is this considered in the stats?

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -86,18 +142,113 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
+                        LoggingUtil.getInfoLogger().warn("SQL-Z3: correctness evaluation failure for query '$sqlQuery'")
+                    } else if (distResult.sqlDistance != 0.0) {
```
jgaleotti: is this case considered in the stats?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
agusaldasoro: Yes. Both branches are recorded by the `reportSqlZ3CorrectnessDistance(distResult.sqlDistance, distResult.sqlDistanceEvaluationFailure)` call right below this block.

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
agusaldasoro: Good catch! It wasn't (the catch is inside `generateSMT`, so the predicate is silently dropped and Z3 still returns SAT).

Now fixed: it increments `skippedQueryConstraints` (here + the JOIN ON catch) and is reported as a new `sqlZ3PartialTranslations` stat column.

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -381,6 +482,33 @@ class Statistics : SearchListener {
...
+                add(Pair("sqlZ3Errors", "$sqlZ3ErrorCount"))
+                add(Pair("sqlZ3ParseFailures", "$sqlZ3ParseFailureCount"))
+                add(Pair("sqlZ3PartialTranslations", "$sqlZ3PartialTranslationCount"))
+                add(Pair("sqlZ3TotalMs", "$sqlZ3TimeMs"))
+                add(Pair("sqlZ3SmtlibGenTotalMs", "$sqlZ3SmtlibGenTimeMs"))
+                add(Pair("sqlZ3AvgSmtlibSizeBytes", "%.1f".format(sqlZ3SmtlibSizeBytes.mean)))
+            }
+
+            // correctness distance stats (only emitted when measureSqlZ3Correctness=true)
+            if (config.measureSqlZ3Correctness) {
```
arcuri82: why 2 distinct configuration options? as we do this only for debugging, could just have 1

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -402,15 +450,24 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
                 for (join in joins) {
                     val onExpressions = join.onExpressions
                     if (onExpressions.isNotEmpty()) {
+                        // KNOWN LIMITATION: only the first ON expression is used; a composite ON
```
arcuri82: should add a `TODO` in the comments here

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -402,15 +450,24 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
+                        // (e.g. "a = b AND c = d") drops all but the first conjunct.
                         val onExpression = onExpressions.elementAt(0)
                         try {
                             val condition = parser.parse(onExpression.toString(), toDBType(schema.databaseType))
                             val tableFromQuery = TablesNamesFinder().getTables(sqlQuery as Statement).first()
+                            // KNOWN LIMITATION: the ON condition is translated with the SAME row index on
```
arcuri82: should add a `TODO` in the comments here

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
 
     companion object {
```
arcuri82: companion objects should be top of file, and not after line 600

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
 
-        // Maps database column types to SMT-LIB types
+        // Bounds for TIMESTAMP columns, encoded as epoch seconds (SMT Int).
```
arcuri82: use JavaDoc `/** */` for fields and functions

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
+        private const val SMT_STRING = "String"
+
+        // SQL type names that need special interpretation beyond their SMT sort: BOOLEAN is encoded as an
+        // SMT String and TIMESTAMP as an SMT Int, so gene reconstruction must consult the original type.
+        // Shared with the comparison sites in this class and referenced by SMTLibZ3DbConstraintSolver.
+        const val BOOLEAN_TYPE = "BOOLEAN"
+        const val TIMESTAMP_TYPE = "TIMESTAMP"
+
+        // The string values a BOOLEAN column may take (BOOLEAN is encoded as an SMT String).
+        private val BOOLEAN_LITERALS = listOf("true", "True", "TRUE", "false", "False", "FALSE")
```
arcuri82: is `tRue` not valid? if not, could just check `true` and `false` without considering case

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
+
+        // Maps database column types to SMT-LIB types.
+        // FUTURE WORK: this is one of three independent type vocabularies interpreting ColumnDto.type
+        // (the others are SMTLibZ3DbConstraintSolver.getColumnDataType and .hasColumnType). They can
+        // silently disagree when a backend reports a variant spelling; consolidating them into a single
+        // source of truth is future work (see the note on SMTLibZ3DbConstraintSolver.hasColumnType).
         private val TYPE_MAP = mapOf(
```
arcuri82: JavaDoc

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
+        // SMT String and TIMESTAMP as an SMT Int, so gene reconstruction must consult the original type.
+        // Shared with the comparison sites in this class and referenced by SMTLibZ3DbConstraintSolver.
+        const val BOOLEAN_TYPE = "BOOLEAN"
+        const val TIMESTAMP_TYPE = "TIMESTAMP"
+
+        // The string values a BOOLEAN column may take (BOOLEAN is encoded as an SMT String).
+        private val BOOLEAN_LITERALS = listOf("true", "True", "TRUE", "false", "False", "FALSE")
+
+        // Maps database column types to SMT-LIB types.
+        // FUTURE WORK: this is one of three independent type vocabularies interpreting ColumnDto.type
```
arcuri82: add `TODO`

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
```diff
@@ -539,37 +608,62 @@ class SmtLibGenerator(private val schema: DbInfoDto, private val numberOfRows: I
...
+            "BIGINT" to SMT_INT,
+            "BIT" to SMT_INT,
+            "INTEGER" to SMT_INT,
+            "INT" to SMT_INT,
+            "INT2" to SMT_INT,
+            "INT4" to SMT_INT,
+            "INT8" to SMT_INT,
+            "TINYINT" to SMT_INT,
+            "SMALLINT" to SMT_INT,
+            // KNOWN LIMITATION: NUMERIC is mapped to Int, so any fractional part is truncated. This is
```
arcuri82: `TODO`

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -54,13 +67,54 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
 
+    // Memoization cache: (sqlQuery, numberOfRows) -> Z3Result (SAT or UNSAT only; errors are not cached)
+    // Schema is assumed stable within a single run, so only query + row count form the key.
+    // Null until Z3 SQL generation is enabled in postConstruct — avoids allocating the map in runs where Z3 SQL generation is off.
+    //
+    // THREAD-SAFETY: this is an access-ordered LinkedHashMap (LRU), whose get() mutates internal order,
+    // so it is NOT thread-safe. It relies on solve() being called from a single thread, which holds today
+    // because the MIO search loop (and thus fitness evaluation / structure mutation) is single-threaded.
+    // If fitness evaluation is ever parallelized, this cache must be synchronized or replaced with a
+    // concurrent LRU, otherwise concurrent access would corrupt it (lost entries / ConcurrentModificationException).
```
arcuri82: concurrency in EM is coming hopefully this year... so yes, use a concurrent map already

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -54,13 +67,54 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
 
+    /*
+        Held WEAKLY on purpose. This solver has @PreDestroy, so each instance is
+        retained by Governator's predestroy-monitor thread (a GC root) for the whole
+        lifetime of the JVM. A strong reference to Statistics would therefore pin
+        Statistics -> Archive -> every individual, leaking across every injector ever
+        created. That is harmless in production (a single injector, process exits) but
+        OOMs test suites that build thousands of injectors (RestIndividualTestBase,
+        SamplerVerifierTest). A weak reference lets that graph be collected once the
+        owning injector is otherwise unreachable, while still resolving fine during an
```
arcuri82, line 94: did you actually experience problems there?

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -86,18 +142,120 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
      *
      * @param sqlQuery The SQL query to solve.
-     * @return A list of SQL actions that can be executed to satisfy the query.
+     * @return A list of SQL actions that can be executed to satisfy the query,
+     *         or an empty list if the problem is UNSAT, unparseable, or an error occurred.
      */
     override fun solve(schemaDto: DbInfoDto, sqlQuery: String, numberOfRows: Int): List<SqlAction> {
-        // TODO: Use memoized, if it's the same schema and query, return the same result and don't do any calculation
+        val collectStats = ::config.isInitialized && config.collectSqlZ3Stats
+        val stats: Statistics? = if (collectStats) statisticsRef?.get() else null
```
arcuri82, line 145: i don't understand the problem here. even if we do not use statics, why not just always inejecting the service? or is it issue with starting/stopping executor for each injector in each E2E test?

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -226,13 +385,17 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
+     * Row constants are named "${smtName}${i}" (e.g. "users1", "users2"). The trailing digits are
+     * stripped so this works for any number of rows (e.g. "users10" -> "users"), not just single-digit
+     * indices. Note: this still assumes table names themselves do not end in a digit.
      *
      * @param key The key containing the table name and index.
      * @return The extracted table name.
      */
     private fun getTableName(key: String): String {
-        return key.substring(0, key.length - 1) // Remove last character
+        return key.replace(Regex("\\d+$"), "")
```
arcuri82: so, if the table name ends with a number (eg "Inventory2026"), that would be removed?

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -226,13 +385,17 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
     }
 
     /**
-     * Extracts the table name from the key by removing the last character (index).
+     * Extracts the table name from a row-constant key by removing the trailing row index.
+     *
+     * Row constants are named "${smtName}${i}" (e.g. "users1", "users2"). The trailing digits are
```
arcuri82: did you choose this? if so, wouldn't better to have a separator? eg, `${smtName}__${i}`

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -325,6 +488,13 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
     /**
      * Maps column types to ColumnDataType.
      *
+     * FUTURE WORK: this recognizes only a small subset of SQL type spellings and falls back to
```
arcuri82: put `TODO`. note that `TODO` and `FIXME` comments are handled specially in IDEs

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -1955,9 +1961,42 @@ class EMConfig {
...
+    @DependsOnTrueFor("generateSqlDataWithZ3")
+    var collectSqlZ3Stats = false
+
+    @Experimental
+    @Cfg("Measure the correctness of Z3-generated SQL inserts by computing the heuristic " +
+            "distance between the original failing WHERE query and the generated INSERT data. " +
+            "Distance=0 means the insert satisfies the WHERE; distance>0 means it does not. " +
+            "Only meaningful when generateSqlDataWithZ3=true.")
+    @DependsOnTrueFor("generateSqlDataWithZ3")
+    var measureSqlZ3Correctness = false
```
arcuri82: no need of 2 separated options to collect statistics

core/pom.xml
```diff
@@ -33,6 +33,10 @@
             <groupId>org.evomaster</groupId>
             <artifactId>evomaster-client-java-controller-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.evomaster</groupId>
+            <artifactId>evomaster-client-java-sql</artifactId>
```
arcuri82: you should not add this kind of dependency between modules before discussing it, as it has several architectural consequences. remove it. different though if it is imported as `test`

core/pom.xml
```diff
@@ -473,6 +477,21 @@
...
+                On CI runners where algorithm tests complete faster than usual, less GC activity
+                accumulates before gene tests run, leaving the heap more fragmented and full.
+                5120m (vs the global 4096m) provides the headroom to survive runner variability.
+                GitHub Actions ubuntu-latest has 7 GB RAM: 5 GB heap + 1 GB Maven + 1 GB OS = 7 GB.
+            -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>@{argLine} -ea -Xms1024m -Xmx5120m -Xss4m -Dfile.encoding=UTF-8 -Djdk.attach.allowAttachSelf=true -Duser.language=en -Duser.country=GB ${addOpens}</argLine>
```
arcuri82: you shouldn't copy&paste configurations like this, as it ends up with code duplication, which must be avoided, as the 2 will inevitably diverge throught time. why not increasing -Xms directly on the root pom avoiding duplication? otherwise, that value itself should be in a param, which is the overriden in properties declaration in this pom file

arcuri82 (changes requested):
@agusaldasoro  hi, this was a large PR, touching different aspects. please try to make smaller, more focused PRs in future :)  eg, changing of surefire setting could had been done independently on a different PR

core/src/main/kotlin/org/evomaster/core/solver/SmtLibGenerator.kt
agusaldasoro: Good point. This is a generation constraint (Z3 picks the value, it's not matching existing input), and the result is read back through `toBoolean()` which is already case-insensitive, so the extra casings were redundant. Simplified to just `"true"/"false"`, which is all Z3 needs to generate.

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
agusaldasoro: Done, wrapped the LRU in `Collections.synchronizedMap`.

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
agusaldasoro: Fixed it by giving row constants an explicit separator (`smtName__index`, via `SmtLibGenerator.rowConstantName`), so getTableName is now just `key.substringBeforeLast("__")`. Trailing digits in the table name are preserved (inventory2026__1 → inventory2026), and it's still unambiguous if the name itself contains `__`.

core/src/main/kotlin/org/evomaster/core/solver/SMTLibZ3DbConstraintSolver.kt
agusaldasoro: Yes, this was a real problem — I hit Java OOM in the test suites (Core/CoreIT), including tests that don't use the Z3 solver at all. I chased it for weeks in [#1609](https://github.com/WebFuzzing/EvoMaster/pull/1609); the weak reference was the only
  thing that reliably fixed it.

  The root cause isn't the injection itself — injecting Statistics normally is fine. The issue is that this solver has a @PreDestroy, so Governator's predestroy-monitor thread holds a reference to
  each solver instance for the whole JVM lifetime (it's a GC root). A strong reference to Statistics then pins Statistics → Archive → all individuals, and because that chain is rooted by a live
  thread, it survives even after the injector is otherwise unreachable. In a test run that builds thousands of injectors, every one of those object graphs is retained → OOM, regardless of whether Z3
  was used.

  Making the reference weak breaks that retention: Statistics is still strongly held by SearchTimeController during an active search (so it resolves fine at runtime), but once the owning injector
  goes away, the graph becomes collectable.

  Happy to hear alternatives if you see a cleaner fix — e.g. not registering @PreDestroy on a long-lived monitored bean — but this was the only approach that held up across the whole suite.

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
agusaldasoro: These correctness stats were something I added to debug the code for myself — the idea was to run them locally only (not on the cluster). They were genuinely useful: they helped me find a few bugs
  by measuring how far the generated rows were from actually satisfying the predicate. But they're not needed for the time-measurement runs we do on the cluster, and they add overhead that could
  interfere with the real objective there — which is exactly why they were a separate flag from collectSqlZ3Stats.

  That said, to keep this PR focused I've moved the whole correctness feature to [a separate branch (dse/sql-z3-correctness)](https://github.com/WebFuzzing/EvoMaster/pull/1637) so we can discuss it later. It's not needed for now, so this PR no longer
  has that second option.

arcuri82 (approved):

PR #1589, opened 2026-06-16
json patch external api test

suarezrominajulieta:
In this pr we fix the TO-DO on json patch dtoWritter, and add a slice of the api  https://github.com/cassiomolin/http-patch-spring, for e2e testing.


suarezrominajulieta:
hi @jgaleotti. I had a question about this following pr, where the idea was to implement the DTO passage, and the point 5 that Andrea requested:

4) Create E2E tests for it

5) Find an open-source project using JSON PATCH (ideally in Java or Kotlin, alternatively Python or JS). On GitHub, searching for “"application/json-patch+json" language:java” yields few results… looking at those, a viable example could be: https://github.com/cassiomolin/http-patch-spring
Another one could be: https://github.com/cjmalloy/jasper

Looking at the two examples, the one at https://github.com/cassiomolin/http-patch-spring is shorter and more concise: I went with that one, but what I wanted to verify is whether I should copy part of that API and create an E2E application with it, or if I should add it to the EMB.

I leave this question here for our meeting, ty!

PR #1605, opened 2026-06-23
N flaky handle

man-zhang:


core/src/main/kotlin/org/evomaster/core/problem/httpws/HttpWsCallResult.kt
```diff
@@ -205,38 +211,316 @@ abstract class HttpWsCallResult : EnterpriseActionResult {
...
-            setFlakyStatusCode(pStatusCode)
+    /**
+     * field value differences between the original result and re-execution results.
+     */
+    data class FieldVariation(
+        val field: ResponseField,
+        val valuesByExecIndex: Map<Int, String?>
+    )
+
+    private val flakyObservations: MutableList<FlakyObservation> = mutableListOf()
```
arcuri82: for readability, fields should up in the file, not mixed in the middle of function declarations. eg, this should be moved just after the `companion object` declarations

core/src/main/kotlin/org/evomaster/core/search/service/FlakinessDetector.kt
```diff
@@ -37,7 +36,12 @@ class FlakinessDetector<T: Individual> : TimeBoxedPhase {
     private lateinit var epc: ExecutionPhaseController
 
     override fun applyPhase() {
-        reexecuteToDetectFlakiness()
+        if (config.enableStaticFlakyInference) {
```
arcuri82, line 43: shouldn't `handleFlakiness` be considered it here?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/service/FlakinessDetector.kt
man-zhang: `handleFlakiness` is considered by where this `applyPhase` is invoked. once the function is invoked, `handleFlakiness` must be true. this check is added now.

arcuri82 (approved):

PR #1606, opened 2026-06-23
Kotlin test writer dollar sign escape bug

lmasroca:
Kotlin requires `$` symbols to be escaped when writing strings. Bug fix for double escape of dollar sign symbols, as `SqlWriter` escaped the already escaped dollar signs, resulting in bare dollar signs appearing on generated Kotlin code (causing compilation errors).

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/SqlWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/SqlWriter.kt
@@ -110,6 +110,14 @@ object SqlWriter {
                 //TODO already escaped???
                 return x.replace("\"","\\\"")
             }
+            if (format.isKotlin()) {
+                // Applying escapeJava on top would corrupt Kotlin escapes, e.g. \$ -> \\$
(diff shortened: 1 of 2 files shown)


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/output/SqlWriter.kt
```diff
@@ -119,6 +123,10 @@ object SqlWriter {
         }
     }
 
+    private fun escapeKotlin(value: String): String {
+        return StringEscapeUtils.escapeJava(value).replace("\\$", "\$")
```
arcuri82, line 133: add some comment to explain why this

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/output/SqlWriter.kt
lmasroca: Hi, just added a brief comment explaining the escape logic here.

arcuri82 (approved):

PR #1610, opened 2026-06-24
Set LLM naming strategy prompts to return a json answer

Pgarrett:
Since we're setting the LLM to respond in JSON, we need to make it clear in the prompt. We also added a fallback name just in case the name returned by the LLM is an invalid JSON or returns any invalid name according to the criteria.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/llm/Prompts.kt
+++ b/core/src/main/kotlin/org/evomaster/core/llm/Prompts.kt
@@ -56,7 +56,7 @@ object Prompts {
     const val NEW_TEST_CASE_NAME = """
         You are an expert software engineer specializing in test naming.
 
-        Given the following test case written in [targetLanguage], produce a descriptive suffix to append to its existing name.
+        Given the following test case written in [targetLanguage], produce a descriptive suffix to append to its existing name. Return a JSON string.
 
         ## Rules
 
@@ -87,7 +87,7 @@ object Prompts {
         [testLines]
     """
 
-    const val RE_ITERATE_TEST_CASE_NAME = "Your previous response contained more than just the suffix. Output only the suffix, nothing else. No explanation, no punctuation, no extra text, do not exceed max chars."
+    const val RE_ITERATE_TEST_CASE_NAME = "Your previous response contained more than just the suffix. Output only the suffix, nothing else. No explanation, no punctuation, no extra text, do not exceed max chars. Return a JSON string."
 
     fun getPromptForNameDescription(name: String, description: String?): Pair<String,String> {
         var user = "Your input is\n [name]:$name"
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/org/evomaster/core/llm/Prompts.kt
```diff
@@ -87,7 +87,7 @@ object Prompts {
         [testLines]
     """
 
-    const val RE_ITERATE_TEST_CASE_NAME = "Your previous response contained more than just the suffix. Output only the suffix, nothing else. No explanation, no punctuation, no extra text, do not exceed max chars."
+    const val RE_ITERATE_TEST_CASE_NAME = "Your previous response contained more than just the suffix. Output only the suffix, nothing else. No explanation, no punctuation, no extra text, do not exceed max chars. Return a JSON string."
```
arcuri82, line 90: hi. this might cover 99.9% of cases. but, still, 1 out of 1000 you could get back something that is not valid JSON. so, need to somehow enforce not crashing when we get an invalid response (and think about what default alternative should be used in those cases)

arcuri82 (changes requested):

core/src/test/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategyTest.kt
```diff
@@ -32,7 +32,7 @@ import kotlin.jvm.java
 class LlmServiceTestCaseNamingStrategyTest {
 
     companion object {
-        const val MAX_NAME_LENGTH = 80
+        const val MAX_NAME_LENGTH = 40
```
arcuri82, line 37: how/why is this different from `maxTestCaseNameLength = 120` ?

arcuri82 (changes requested):

core/src/test/kotlin/org/evomaster/core/output/naming/LlmServiceTestCaseNamingStrategyTest.kt
Pgarrett: @arcuri82 just to force a condition in which the LLM does not respond with what we asked since the mock will return a longer name, thus forcing the re-prompt and later the fallback name

arcuri82: @Pgarrett  ah, this is in a test? not production code... isn't it?

arcuri82: @Pgarrett  maybe you can just add what you wrote as a comment there?

Pgarrett: Yep, it's a test, haha

Pgarrett: Done

arcuri82 (approved):

PR #1614, opened 2026-06-29
Java regex COMMENTS (x) flag

lmasroca:
Added support for Java regex `COMMENTS` (`x`) flag, this flag strips white space and comments (e.g.: `#this is a comment\n`) from the regex. This is done by pre-processing the regex string. Comments are started by the `#` character and end on line terminators.


core/src/main/kotlin/org/evomaster/core/parser/RegexHandler.kt
```diff
@@ -48,6 +51,108 @@ object RegexHandler {
         return gene
     }
 
+    /**
+     * This function handles comments and whitespace for Java regex, striping them when the "x" flag is on.
+     */
+    private fun preprocessCommentsForJavaRegex(regex: String, externalRegexFlags: RegexFlags): String {
```
jgaleotti, line 68: This seems as a very low level parsing code. Could you explain a bit more in the documentation why this is required (instead of using ANTLR)

jgaleotti (changes requested):

jgaleotti (approved):

arcuri82 (approved):

PR #1627, opened 2026-07-03
invalid merge patch

omursahin:
```
   /**
    * Calls:
    * 1 - (201) POST:/api/mergepatch/buggy
    * 2 - (200) GET:/api/mergepatch/buggy/{id}
    * 3 - (200) PATCH:/api/mergepatch/buggy/{id}
    * 4 - (200) GET:/api/mergepatch/buggy/{id}
    * Found 1 potential fault of type-code 922
    */
    @Test @Timeout(60)
    fun test_10_getOnBuggiInvalidMergePatch()  {
        
        val dto_MergePatchResource_1 = MergePatchResource()
        dto_MergePatchResource_1.setValue(526)
        
        val res_0: ValidatableResponse = given().accept("*/*")
                .header("x-EMextraHeader123", "")
                .body(dto_MergePatchResource_1)
                .post("${baseUrlOfSut}/api/mergepatch/buggy")
                .then()
                .statusCode(201)
                .assertThat()
                .body("'name'", nullValue())
                .body("'value'", numberMatches(526))
        val location_buggy__2 = res_0.extract().header("location")
        assertTrue(isValidURIorEmpty(location_buggy__2));
        
        given().accept("*/*")
                .header("x-EMextraHeader123", "")
                .get(resolveLocation(location_buggy__2, baseUrlOfSut + "/api/mergepatch/buggy/856"))
                .then()
                .statusCode(200)
                .assertThat()
                .body("'name'", nullValue())
                .body("'value'", numberMatches(526))
        
        // Fault922. JSON Merge Patch changes untouched fields.
        val dto_MergeRequest_4 = MergeRequest()
        
        given().accept("*/*")
                .header("x-EMextraHeader123", "")
                .contentType("application/merge-patch+json")
                .body(dto_MergeRequest_4)
                .patch(resolveLocation(location_buggy__2, baseUrlOfSut + "/api/mergepatch/buggy/856?EMextraParam123=42"))
                .then()
                .statusCode(200)
                .assertThat()
                .body("'name'", nullValue())
                .body("'value'", nullValue())
        
        given().accept("*/*")
                .header("x-EMextraHeader123", "")
                .get(resolveLocation(location_buggy__2, baseUrlOfSut + "/api/mergepatch/buggy/856"))
                .then()
                .statusCode(200)
                .assertThat()
                .body("'name'", nullValue())
                .body("'value'", nullValue())
    }
    
```


core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -725,6 +725,81 @@ object HttpSemanticsOracle {
...
+        if (before.verb != HttpVerb.GET) return false
+        if (patch.verb != HttpVerb.PATCH) return false
+        if (after.verb != HttpVerb.GET) return false
+
+        if (!before.usingSameResolvedPath(patch) || !after.usingSameResolvedPath(patch)) return false
+        // the two GETs must use the same auth for a meaningful state comparison
+        if (before.auth.isDifferentFrom(after.auth)) return false
+
+        // merge-patch bodies are JSON (incl. application/merge-patch+json)
+        val bodyParam = patch.parameters.find { it is BodyParam } as BodyParam?
```
arcuri82, line 763: here you should explicitly check for ` application/merge-patch+json` or `application/json`.  other formats such as `application/json-patch+json` would be wrong, and so cannot just use `isJson()`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -725,6 +725,81 @@ object HttpSemanticsOracle {
...
+
+        val resBefore = actionResults.find { it.sourceLocalId == before.getLocalId() } as RestCallResult? ?: return false
+        val resPatch  = actionResults.find { it.sourceLocalId == patch.getLocalId() } as RestCallResult? ?: return false
+        val resAfter  = actionResults.find { it.sourceLocalId == after.getLocalId() } as RestCallResult? ?: return false
+
+        if (!StatusGroup.G_2xx.isInGroup(resBefore.getStatusCode())) return false
+        if (!StatusGroup.G_2xx.isInGroup(resPatch.getStatusCode())) return false
+        if (!StatusGroup.G_2xx.isInGroup(resAfter.getStatusCode())) return false
+
+        val untouched = extractModifiedFieldNames(patch) - extractSentFieldNames(patch)
```
arcuri82, line 782: can add the following comment:
`If there are flaky fields, eg, timestamps, there could be a different value between the 2 GETs, regardless of the PATCH. However, we are only looking at the fields of PATCH. it would be weird to set a value with a PATCH that the API can change on a whim... not impossible, but most likely very unlikely. so, in theory field flakiness should not be a problem here `

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -480,6 +483,60 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+
+            val getDef = actionDefinitions.find { it.verb == HttpVerb.GET && it.path == patchOp.path }
+                ?: return@forEach
+
+            val successPatches = RestIndividualSelectorUtils.findIndividuals(
+                individualsInSolution, HttpVerb.PATCH, patchOp.path, statusGroup = StatusGroup.G_2xx
+            )
+            if (successPatches.isEmpty()) return@forEach
+
+            for (candidate in successPatches.sortedBy { it.individual.size() }) {
```
arcuri82, line 511: are we checking this for every possible candidate? or shouldn't we break as soon as we get one that works?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
omursahin: @arcuri82 It checks every possible solution. We can return early if we find a fault, if you want, but it can also be valuable to report different combinations of the fault. What do you think?

arcuri82: are we differentiating these potential different cases in the fitness function using the `context` for the detected faults? if not, then it is no point, as anyway any other new test would be discarded in archive as not improving fitness

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
omursahin: No, you are right. It will not improve fitness.

arcuri82 (approved):

PR #1628, opened 2026-07-04
Adding MCP problem type basic sampler infrastructure

guidorc:
# Summary

Adding the basic search infrastructure for blackbox fuzzing of Model Context Protocol (MCP) servers. The approach is: connect to the target server, discover its capabilities via the standard MCP handshake, automatically construct a mutable input representation (gene tree) for each tool and resource from their JSON Schema definitions, then use a search algorithm to evolve a sequences of calls that maximise coverage; reaching each capability and obtaining a non-error response

###  MCP client (HttpMcpClient, McpClient, McpDTOs)
An HTTP-based client that implements the MCP JSON-RPC protocol: initialization handshake, tool invocation (tools/call), static resource reads (resources/read), and capability discovery (tools/list, resources/list, resources/templates/list).

### Action model (McpAction, McpToolCallAction, McpResourceReadAction)
Two concrete action types mirror the two MCP interaction patterns: calling a named tool with structured arguments, and reading a resource by URI (including template URIs with mutable parameters).

### Individual (McpIndividual)
A test case represented as an ordered sequence of McpActions, integrating with EvoMaster's existing EnterpriseIndividual and mutation infrastructure.

### Sampler (McpSampler)
On startup, connects to the target MCP server, performs the handshake, and discovers all tools and resources. Each tool's JSON Schema input definition is converted into an ObjectGene tree used by the search engine to mutate arguments. Pre-built single-action individuals ensure every capability is exercised at least once before random search begins.

### Fitness function (McpBlackBoxFitness)
Evaluates individuals by executing their actions sequentially and scoring coverage targets based on observable protocol signals: 1.0 for a successful tool call or resource read, 0.5 for a server-reported error (partial credit keeps the search gradient useful). Exceptions break the action sequence early.

### McpBlackBoxModule
Wires all the above into EvoMaster's dependency injection container and registers the problem type in the main entry point.

## Follow Up

  - McpTestCaseWriter — the test output pipeline currently uses NoTestCaseWriter, so generated test files have empty bodies. A dedicated writer that emits executable MCP
  client calls is needed.
  - Format-aware gene building — string fields carrying semantic formats (date-time, email, uuid) are built as plain StringGene. The sampler needs a three-layer cascade
  (explicit JSON Schema format → type switch → name-based heuristics) to generate valid values for these fields, matching the approach used by the REST builder.
  - SortingHelper / TestCaseNamingStrategyFactory — McpIndividual is not yet registered in the test sorting and naming strategies, causing a handled warning during output.


core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+    private fun openConnection(method: String, params: Map<String, Any?>): Pair<HttpURLConnection, String> {
+        val body = mapper.writeValueAsString(
+            mapOf(
+                "jsonrpc" to "2.0",
+                "method" to method,
+                "params" to params,
+                "id" to nextId()
+            )
+        )
+        val conn = URL(baseUrl).openConnection() as HttpURLConnection
```
Pgarrett: Any particular reason for using this connection method and not something like the Jersey client? Like [GraphQL](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/problem/graphql/IntrospectiveQuery.kt#L28) or [REST](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/problem/rest/schema/OpenApiAccess.kt#L200) do so

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+            mapOf(
+                "jsonrpc" to "2.0",
+                "method" to method,
+                "params" to params
+                // notifications have no "id"
+            )
+        )
+        val conn = URL(baseUrl).openConnection() as HttpURLConnection
+        conn.requestMethod = "POST"
+        conn.setRequestProperty("Content-Type", "application/json")
```
Pgarrett: If we can use the jersey http client, we should be using the constants provided by the library, in order to avoid these strings

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+        // Read (and discard) the response to complete the HTTP exchange
+        try { conn.inputStream.close() } catch (_: Exception) {}
+    }
+
+    /** Send a JSON-RPC request. Returns null on 4xx (method not supported). */
+    private fun post(method: String, params: Map<String, Any?> = emptyMap()): Map<String, Any?>? {
+        val (conn, _) = openConnection(method, params)
+        val status = conn.responseCode
+        if (status == 400 || status == 404 || status == 405) {
+            return null
```
Pgarrett: Under which circumstances would these status code mean anything valid? In other words, is `null` a valid return here? As an empty map or a map with content are? If not, then we should throw an exception with the corresponding error. Maybe there's an error on the MCP, or at least log a warning for the user

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+        var cursor: String? = null
+        do {
+            val params: Map<String, Any?> = if (cursor != null) mapOf("cursor" to cursor) else emptyMap()
+            val response = post("tools/list", params) ?: break
+            val result = response["result"] as? Map<String, Any?> ?: break
+            val items = result["tools"] as? List<*> ?: emptyList<Any>()
+            items.filterIsInstance<Map<String, Any?>>().forEach { t ->
+                tools.add(
+                    McpToolDefinition(
+                        name = t["name"] as? String ?: "",
```
Pgarrett: Is a tool with no name a valid state?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+    /**
+     * Perform the MCP initialization handshake (initialize + notifications/initialized).
+     * Must be called once before any other method.
+     */
     fun initialize() {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val (conn, _) = openConnection(
+            "initialize",
+            mapOf(
+                "protocolVersion" to "2024-11-05",
```
Pgarrett: Extract to constants. Is this the latest protocol version? Or the first one using streamable HTTP and as such we support as from there? Add a comment mentioning why that protocol version

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+                "capabilities" to emptyMap<String, Any>(),
+                "clientInfo" to mapOf("name" to "EvoMaster", "version" to "1.0.0")
```
Pgarrett: Should this be the EvoMaster version? Rather than the MCP implementation version. So as to avoid confusion. How to get EM version: https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/Main.kt#L225

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+        val responseBody = conn.inputStream.use { it.readBytes().toString(Charsets.UTF_8) }
+        return mapper.readValue(responseBody, Map::class.java) as Map<String, Any?>
     }
 
     override fun listTools(): List<McpToolDefinition> {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val tools = mutableListOf<McpToolDefinition>()
+        var cursor: String? = null
+        do {
+            val params: Map<String, Any?> = if (cursor != null) mapOf("cursor" to cursor) else emptyMap()
```
Pgarrett: Add a comment on `cursor` being related to pagination handling

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+            val response = post("tools/list", params) ?: break
```
Pgarrett: I find these two `break` rather odd, something's off with this type of handling

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+        } while (cursor != null)
+        return tools
     }
 
     override fun listResources(): List<McpResourceDefinition> {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val resources = mutableListOf<McpResourceDefinition>()
+        var cursor: String? = null
+        do {
+            val params: Map<String, Any?> = if (cursor != null) mapOf("cursor" to cursor) else emptyMap()
```
Pgarrett: Rather same handling as above, wondering if this could actually be extracted to a method of the kind `handleListResponseWithPagination` or something along those lines

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+            val response = post("resources/list", params) ?: break
+            val result = response["result"] as? Map<String, Any?> ?: break
+            val items = result["resources"] as? List<*> ?: emptyList<Any>()
+            items.filterIsInstance<Map<String, Any?>>().forEach { r ->
+                resources.add(
+                    McpResourceDefinition(
+                        uri = r["uri"] as? String ?: "",
```
Pgarrett: Same comments as above

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
+        } while (cursor != null)
+        return resources
     }
 
     override fun listResourceTemplates(): List<McpResourceTemplate> {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val templates = mutableListOf<McpResourceTemplate>()
+        var cursor: String? = null
+        do {
+            val params: Map<String, Any?> = if (cursor != null) mapOf("cursor" to cursor) else emptyMap()
```
Pgarrett: Naturally, same comments

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
 
     override fun callTool(name: String, arguments: Map<String, Any?>): McpToolResult {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val response = post("tools/call", mapOf("name" to name, "arguments" to arguments))
+            ?: return McpToolResult(isError = true)
+        val result = response["result"] as? Map<String, Any?> ?: return McpToolResult(isError = true)
+        val rawContent = result["content"] as? List<*> ?: emptyList<Any>()
+        val content = rawContent.filterIsInstance<Map<String, Any?>>().map { c ->
+            McpContent(
+                type = c["type"] as? String ?: "text",
```
Pgarrett: Safe to assume this type? Or should we log a warning/throw an exception?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,214 @@
...
 
     override fun readResource(uri: String): McpResourceResult {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val response = post("resources/read", mapOf("uri" to uri))
+            ?: return McpResourceResult()
+        val result = response["result"] as? Map<String, Any?> ?: return McpResourceResult()
+        val rawContents = result["contents"] as? List<*> ?: emptyList<Any>()
+        val contents = rawContents.filterIsInstance<Map<String, Any?>>().map { c ->
+            McpContent(
+                type = c["type"] as? String ?: "text",
```
Pgarrett: Same

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
```diff
@@ -1,17 +1,128 @@
...
+                        result.setIsError(false)
+                        result.stopping = false
+
+                        val targetId = idMapper.handleLocalTarget("resource:${action.id}")
+                        // A successful read (even empty contents) scores 1.0
+                        fv.updateTarget(targetId, 1.0, i)
+                    }
+
+                    else -> {
+                        result.stopping = false
```
Pgarrett, line 80: If it's neither `McpToolCallAction` nor `McpResourceReadAction`, what can it be that's valid?

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
```diff
@@ -1,17 +1,128 @@
...
+    // -------------------------------------------------------------------------
+    // Helpers
+    // -------------------------------------------------------------------------
+
+    /**
+     * Convert an [ObjectGene] to a [Map] of argument values suitable for passing
+     * to [McpClient.callTool]. Uses typed gene values when available, falling
+     * back to printable string representation for complex types.
+     */
+    private fun geneToMap(gene: ObjectGene): Map<String, Any?> {
```
Pgarrett, line 110: I feel this should actually be provided be the `ObjectGene` itself. In this implementation you'll be missing the handling of `EnumGene` or even `OptionalGene` to check if a gene is active or not.

Maybe we should have something like [ArrayGene.getViewOfElements](getViewOfElements)

Another option, if you already know which fields you'll be needing is to use [ObjectGene.getField](https://github.com/WebFuzzing/EvoMaster/blob/master/core/src/main/kotlin/org/evomaster/core/search/gene/ObjectGene.kt#L776)

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
```diff
@@ -2,35 +2,233 @@ package org.evomaster.core.problem.mcp.service
...
     }
 
-    override fun hasSpecialInitForSmartSampler(): Boolean {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
-    }
+    override fun hasSpecialInitForSmartSampler(): Boolean = adHocInitialIndividuals.isNotEmpty()
 
     override fun initSeededTests(infoDto: SutInfoDto?) {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        // Not supported in Phase 3
```
Pgarrett: Why remove the exception if it's still not supported?

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
```diff
@@ -2,35 +2,233 @@ package org.evomaster.core.problem.mcp.service
...
+
+        for ((_, action) in toolActionCluster) {
+            val copy = action.copy() as McpAction
+            copy.doInitialize(randomness)
+            val ind = McpIndividual(SampleType.RANDOM, mutableListOf(makeGroup(copy)))
+            ind.doGlobalInitialize(searchGlobalState)
+            adHocInitialIndividuals.add(ind)
+        }
+
+        for ((_, action) in resourceActionCluster) {
```
Pgarrett: Avoid duplicate code

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
```diff
@@ -2,35 +2,233 @@ package org.evomaster.core.problem.mcp.service
...
+    internal fun buildGeneFromSchema(name: String, schema: Map<String, Any?>): Gene {
+        val type = schema["type"] as? String
+
+        return when (type) {
+            "string" -> StringGene(name)
+            "integer", "number" -> IntegerGene(name)
+            "boolean" -> BooleanGene(name)
+            "object" -> buildObjectGeneFromSchema(name, schema)
+            "array" -> ArrayGene(name, StringGene("element"))
+            else -> StringGene(name) // fallback for unknown/null types
```
Pgarrett: Rather throw an exception I'd say

core/src/main/kotlin/org/evomaster/core/problem/mcp/McpIndividual.kt
```diff
@@ -2,12 +2,23 @@ package org.evomaster.core.problem.mcp
 
 import org.evomaster.core.problem.api.ApiWsIndividual
 import org.evomaster.core.problem.enterprise.EnterpriseChildTypeVerifier
+import org.evomaster.core.problem.enterprise.EnterpriseIndividual.Companion.getEnterpriseTopGroups
```
Pgarrett: What's this import for? Seems unused

core/src/test/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClientTest.kt
```diff
@@ -0,0 +1,183 @@
...
+                    aResponse()
+                        .withStatus(200)
+                        .withHeader("Content-Type", "application/json")
+                        .withBody(responseBody)
+                )
+        )
+    }
+
+    @Test
+    fun `listTools parses tool definitions correctly`() {
```
Pgarrett: Not very fond of this type of test naming, I can see them only in one class

Pgarrett (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: No particular reason, just unfamiliar with the Jersey client. We should be able to use it here as well 👍

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Will extract into shared MCP constants file. The version was a mistake due to testing with a sample MCP server, updated it to the latest protocol version.

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: No, will remove the fallback values for required properties

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: The idea was to break early if there was an invalid response from the server, will change it to throw a clear exception instead

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Extracted the navigation logic into an aux private function

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: We should throw an exception since tool results should always include a type as per the MCP specification

core/src/main/kotlin/org/evomaster/core/problem/mcp/McpIndividual.kt
guidorc: Left over from testing, removed it

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Left over from debugging with invalid MCP server, removed the catch since we should not silently drop these cases

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
guidorc: removed it for testing, will add it back

core/src/main/kotlin/org/evomaster/core/problem/mcp/McpConst.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * JSON-RPC version used for all MCP messages, as required by the MCP specification.
+     */
+    const val JSONRPC_VERSION = "2.0"
+
+    /**
+     * MCP protocol version negotiated during the initialize handshake.
+     */
+    const val PROTOCOL_VERSION = "2025-11-25"
```
Pgarrett, line 13: Just as a question, could this be a parameter for when fuzzing an MCP API? Or are protocol versions not backward compatible?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
Pgarrett: Unaddressed comment @guidorc

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,207 @@
...
+        })
     }
 
     override fun callTool(name: String, arguments: Map<String, Any?>): McpToolResult {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val response = post("tools/call", mapOf("name" to name, "arguments" to arguments))
+            ?: return McpToolResult(isError = true)
+        val result = response["result"] as? Map<String, Any?> ?: return McpToolResult(isError = true)
+        val rawContent = result["content"] as? List<Map<String, Any?>> ?: emptyList()
+        val content = rawContent.map { c ->
```
Pgarrett: Favour readability, this could be extracted to a `getMcpContent` method

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,207 @@
...
+            val type = c["type"] as? String
+                ?: throw IllegalStateException("tools/call response content item missing required 'type' field")
```
Pgarrett: Can we add more information on which tool call is actually causing this error? If we're fuzzing to test an API, it's best if we provide the most information to our users. Applicable to the other exceptions being thrown

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,207 @@
...
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.evomaster.core.problem.mcp.McpConst
+import org.evomaster.core.remote.HttpClientFactory
+import java.util.concurrent.atomic.AtomicInteger
+import javax.ws.rs.client.Client
+import javax.ws.rs.client.Entity
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response
+import kotlin.String
```
Pgarrett: Are all this imports needed?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,207 @@
...
     }
 
     override fun readResource(uri: String): McpResourceResult {
-        throw UnsupportedOperationException("MCP server analysis is not yet supported")
+        val response = post("resources/read", mapOf("uri" to uri))
+            ?: return McpResourceResult()
+        val result = response["result"] as? Map<String, Any?> ?: return McpResourceResult()
+        val rawContents = result["contents"] as? List<Map<String, Any?>> ?: emptyList()
+        val contents = rawContents.map { c ->
+            McpContent(
```
Pgarrett: How come this `McpContent` can be created without `type` and we throw an exception for the one in `callTool`? It's not a good practice to use the same object to represent different things. Either this should have a type assigned, or it should be a different object

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/McpDTOs.kt
```diff
@@ -1,35 +1,43 @@
...
 
+/** Result of a `resources/read` invocation, as defined by the MCP specification. */
 data class McpResourceResult(
     val contents: List<McpContent> = emptyList()
 )
 
+/** Content item within a tool or resource response. */
 data class McpContent(
-    val type: String,
+    val type: String? = null,
```
Pgarrett: Everything is nullable? Why? I fear we might hide issues due to this flexibility

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
```diff
@@ -1,36 +1,220 @@
...
+        try {
+            mcpClient.initialize()
+        } catch (e: Exception) {
+            throw SutProblemException(
+                "Failed to initialize MCP session at '${config.base}'. Cause: ${e.message}"
+            )
+        }
+
+        // Discover tools
+        val tools = mcpClient.listTools()
```
Pgarrett, line 49: Maybe extract this and subsequent discoveries to their own methods. That way we have a more concise and declarative `initialize` method

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpSampler.kt
```diff
@@ -1,36 +1,220 @@
...
+            toolActionCluster.values.toList() + resourceActionCluster.values.toList()
+
+        if (allActions.isEmpty()) {
+            // Edge case: no capabilities discovered — return empty individual
+            val ind = McpIndividual(SampleType.RANDOM, mutableListOf())
+            ind.doGlobalInitialize(searchGlobalState)
+            return ind
+        }
+
+        val n = randomness.nextInt(1, getMaxTestSizeDuringSampler())
```
Pgarrett: What's `n`? Choose a better name

Pgarrett (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/mcp/McpConst.kt
guidorc: The protocol versions are not backwards compatible, every new version contains breaking changes. That said we could look into if those changes affect our analysis

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Will add the tool name in the error message

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Removed the redundant ones

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Split McpContent into McpResourceContent and McpToolContent to address this. Resources don't have a `type` property.

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/McpDTOs.kt
guidorc: Refactored it to match the possible types defined in the MCP spec: https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-result

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Copied over the version resolution from Main.kt

Pgarrett (approved):

Pgarrett:
@jgaleotti can you review please?

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
Pgarrett: Hey @guidorc, to avoid this, could we use the `MapGene`? Or do you need it to be an `ObjectGene`?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,236 @@
...
+        }
+        // Send the required follow-up notification (fire-and-forget)
+        postNotification("notifications/initialized", emptyMap())
+    }
+
+    /** Send a JSON-RPC notification (no response expected). */
+    private fun postNotification(method: String, params: Map<String, Any?>) {
+        val response = sendJsonRpc(method, params, id = null, acceptEventStream = false)
+        // Discard the response to complete the HTTP exchange and release the connection
+        try { response.close() } catch (_: Exception) {}
```
jgaleotti: should exceptions go unnoticed here? Shouldn't this be logged?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,236 @@
...
+
+        do {
+            // On first request cursor is null.
+            // On subsequent requests, include the cursor returned by the previous response.
+            val params = if (cursor != null) mapOf("cursor" to cursor) else emptyMap()
+
+            val response = post(method, params)
+                ?: throw IllegalStateException("$method request failed: received null response")
+
+            val result = response["result"] as? Map<String, Any?>
```
jgaleotti, line 145: should we have "result" as a constant?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,236 @@
...
+            )
+        }
+        // Capture session ID before reading the body
+        response.getHeaderString(McpConst.SESSION_ID_HEADER)?.let { sessionId = it }
+        val responseBody = response.readEntity(String::class.java)
+        if (responseBody.isBlank()) {
+            throw IllegalStateException("MCP initialize handshake returned empty body")
+        }
+        // Send the required follow-up notification (fire-and-forget)
+        postNotification("notifications/initialized", emptyMap())
```
jgaleotti: should this string be a constant value?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,236 @@
...
+        this::class.java.`package`?.implementationVersion ?: "SNAPSHOT"
+
+    /** Send a JSON-RPC message. [id] is omitted for notifications. */
+    private fun sendJsonRpc(method: String, params: Map<String, Any?>, id: Int?, acceptEventStream: Boolean): Response {
+        val payload = mutableMapOf<String, Any?>(
+            "jsonrpc" to McpConst.JSONRPC_VERSION,
+            "method" to method,
+            "params" to params
+        )
+        id?.let { payload["id"] = it }
```
jgaleotti, line 55: should "id" be a constant?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,236 @@
...
+        return results
+    }
+
+    private fun getToolResponseContent(name: String, mcpResponse: Map<String, Any?>?) : List<McpToolContent> {
+        val content = mcpResponse?.get("content") as? List<Map<String, Any?>> ?: emptyList()
+        return content.map { item ->
+            val type = item["type"] as? String
+                ?: throw IllegalStateException("Tool: $name response content item is missing required 'type' field")
+            when (type) {
+                "text" -> McpTextToolContent(item["text"] as? String ?: "")
```
jgaleotti: plenty of constant strings are used here, should they be constant values?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Here we are only silencing a failure while closing the connection for an already received response. If the actual request fails that will be propagated and fail the initialization. WIll add a log for tracing purposes.

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Extracted MCP specific types to constants

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: I don't think we gain much in this case

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: Same as for "response"

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
guidorc: For the sake of reducing the scope of this PR, I propose to return a simple object gene here. In the next PR we can focus on adding an "McpActionBuilder", analog to what's implemented on the GraphQl sampler

guidorc:
@jgaleotti @Pgarrett ready for re-review when you're available

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,246 @@
...
+    companion object {
+        private val log: Logger = LoggerFactory.getLogger(HttpMcpClient::class.java)
+    }
+
+    private val mapper: ObjectMapper = ObjectMapper()
+    private val idCounter = AtomicInteger(1)
+
+    private val client: Client = HttpClientFactory.createTrustingJerseyClient(true, readTimeoutMs)
+
+    @Volatile private var sessionId: String? = null
```
arcuri82, line 40: why marking this as volatile?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,246 @@
...
+    private fun sendJsonRpc(method: String, params: Map<String, Any?>, id: Int?, acceptEventStream: Boolean): Response {
+        val payload = mutableMapOf<String, Any?>(
+            "jsonrpc" to McpConst.JSONRPC_VERSION,
+            "method" to method,
+            "params" to params
+        )
+        id?.let { payload["id"] = it }
+        val body = mapper.writeValueAsString(payload)
+
+        val acceptTypes = if (acceptEventStream) arrayOf(MediaType.APPLICATION_JSON, MediaType.SERVER_SENT_EVENTS) else arrayOf(MediaType.APPLICATION_JSON)
```
arcuri82: `if/else` should be multiline, unless short

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,246 @@
...
+            acceptEventStream = true
+        )
+        val status = response.status
+        if (status >= 400) {
+            throw IllegalStateException(
+                "MCP initialize handshake failed with HTTP $status at '$baseUrl'"
+            )
+        }
+        // Capture session ID before reading the body
+        response.getHeaderString(McpConst.SESSION_ID_HEADER)?.let { sessionId = it }
```
arcuri82, line 94: what if there is no session id? is that optional or mandatory?

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
```diff
@@ -1,28 +1,246 @@
...
+    }
+
+    /** Send a JSON-RPC notification (no response expected). */
+    private fun postNotification(method: String, params: Map<String, Any?>) {
+        val response = sendJsonRpc(method, params, id = null, acceptEventStream = false)
+        // Discard the response to complete the HTTP exchange and release the connection
+        try {
+            response.close()
+        } catch (e: Exception) {
+            log.trace("Failed to close MCP notification response for '$method'", e)
```
arcuri82: `trace`? shouldn't this be a `warn`?

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
arcuri82: @guidorc  in that case, add a brief `TODO` comment here to state it

core/src/main/kotlin/org/evomaster/core/problem/mcp/service/McpBlackBoxFitness.kt
```diff
@@ -1,17 +1,128 @@
...
+        val map = mutableMapOf<String, Any?>()
+        for (field in gene.fields) {
+            map[field.name] = extractGeneValue(field)
+        }
+        return map
+    }
+
+    private fun extractGeneValue(gene: Gene): Any? {
+        return when (gene) {
```
arcuri82, line 127: add `TODO` comment here related to the handling of nullable, optional and choice genes

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: So that any thread that reads `sessionId` sees the most recent value. Currently `sessionId` is written only once on initialize, but it may be written over in case a "reconnect" is needed (session expired).

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: The `sessionId` is optional, stateless servers don't include a `sessionId` in the initialize response.
In case we receive no `sessionId`, the rest of MCP requests just omit the header.
Ref: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management

guidorc:
@arcuri82 thanks for the feedback, ready for re-review

jgaleotti:
@guidorc please ask for a re-review when your changes are finished. Also, it seems the branch needs updating.

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
arcuri82: @guidorc  in such a case, you need to add a comment to specify it. eg `// the session id is optional, as missing for example in stateless servers: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management`

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
arcuri82: add a comment: `// we aim at having this class being thread-safe`... however, is `client` thread-safer?

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: > is client thread-safe?

Not entirely, two threads may read a stale session id at the same time. However the reconnect logic is not yet implemented. Added a TO-DO comment on that.

core/src/main/kotlin/org/evomaster/core/problem/mcp/client/HttpMcpClient.kt
guidorc: @arcuri82 done 👍

arcuri82 (approved):

PR #1603, opened 2026-06-20
json patch dto conversion

suarezrominajulieta:
In this pr we fix the TO-DO on json patch dtoWritter:

Idea: Create two separate applications and two end-to-end (E2E) tests, clearly documenting what is being tested. These tests should follow the format outlined in core-tests/e2e-tests/spring/spring-rest-openapi-v3/src/test/kotlin/org/evomaster/e2etests/spring/openapi/v3/dtoreflectiveassert/DtoReflectiveAssertEMTest.kt.

The first test should verify that Evomaster can generate the DTO for JSON patch when the values ​​of its operations are primitive( and dtoForRequestPayload=true). The second test should verify that the DTO is generated when the object's value is complex, not primitive.

Additionally, the DTOs can be checked in the generated test suite files.

PR #1630, opened 2026-07-07
 json patch output in tests still string

suarezrominajulieta:
In base to Andrea's comment:

"fine for now. not in this PR, but make new branch PR in which you create the following E2E for an API that:

has a POST to create a resource
has a PATCH to manipulate such created resource (404 if no resource)
and then make sure using DTO (there is a EMConfig for it), so that, in same generates test, we have a DTO used for POST following a call on PATCH without DTO (ie Patch must use a string representation, at least for now). any case, test should not fail to compile"

i created this pr with a test that proves that. this is one of the success test generated:

    /**
    * Calls:
    * 1 - (201) POST:/api/jsonpatchdto/resources
    * 2 - (200) PATCH:/api/jsonpatchdto/resources/{id}
    */
    @Test @Timeout(60)
    public void test_4_patchOnResourcReturns200() throws Exception {
        
        BBJsonPatchResourceDto dto_BBJsonPatchResourceDto_1 = new BBJsonPatchResourceDto();
        dto_BBJsonPatchResourceDto_1.setName("1GmNB");
        dto_BBJsonPatchResourceDto_1.setAge(618);
        
        ValidatableResponse res_0 = given().accept("application/json")
                .body(dto_BBJsonPatchResourceDto_1)
                .post(baseUrlOfSut + "/api/jsonpatchdto/resources")
                .then()
                .statusCode(201)
                .assertThat()
                .contentType("application/json");
        String location_resources__2 = "/api/jsonpatchdto/resources" + "/" + res_0.extract().body().path("id").toString();
        
        given().accept("text/plain")
                .contentType("application/json-patch+json")
                .body(" [ { " + 
                    " \"op\" : \"test\", " + 
                    " \"path\" : \"/c\", " + 
                    " \"value\" : \"eDgXhicy5FjjbV7h\" " + 
                    " }, { " + 
                    " \"op\" : \"copy\", " + 
                    " \"from\" : \"/b\", " + 
                    " \"path\" : \"/a\" " + 
                    " }, { " + 
                    " \"op\" : \"replace\", " + 
                    " \"path\" : \"/d\", " + 
                    " \"value\" : -1207061644 " + 
                    " } ] ")
                .patch(resolveLocation(location_resources__2, baseUrlOfSut + "/api/jsonpatchdto/resources/-2191309024963475557"))
                .then()
                .statusCode(200)
                .assertThat()
                .contentType("text/plain")
                .body(containsString("patched"));
    }


core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/jsonpatchdto/BBJsonPatchDtoTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        @BeforeAll
+        @JvmStatic
+        fun init() {
+            val config = EMConfig()
+            initClass(BBJsonPatchDtoController(), config)
+        }
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = OutputFormat::class, names = ["JAVA_JUNIT_5", "KOTLIN_JUNIT_5"])
```
arcuri82, line 33: add a comment to specify why test is only run for those formats

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/jsonpatchdto/BBJsonPatchDtoTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+            assertTrue(solution.individuals.size >= 1)
+            assertHasAtLeastOne(solution, HttpVerb.POST, 201, "/api/jsonpatchdto/resources", null)
+            assertHasAtLeastOne(solution, HttpVerb.PATCH, 200, "/api/jsonpatchdto/resources/{id}", "patched")
+            assertHasAtLeastOne(solution, HttpVerb.PATCH, 404, "/api/jsonpatchdto/resources/{id}", "missing")
+
+            assertPostThenPatchInSameTest(solution)
+        }
+
+        val generatedCode = readGeneratedCode(outputFormat, outputFolderName)
```
arcuri82, line 61: the inherited class `EnterpriseTestBase` already has methods such as `assertTextInTests` that should work here. use those instead of creating new one, unless those do not match your needs. also, as this would be a general function that could be re-used in different tests, shouldn't be here, but rather moved higher up in the hierarchy (eg under `EnterpriseTestBase`)

arcuri82: or into `SpringTestBase` if specific to BB

core-tests/e2e-tests/spring/spring-rest-bb/src/test/kotlin/org/evomaster/e2etests/spring/rest/bb/jsonpatchdto/BBJsonPatchDtoTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        )
+    }
+
+    /**
+     * Verifies that the solution contains at least one generated test (individual) in which a successful POST
+     * that creates the resource (201) is followed, later in the same test, by a successful PATCH (200) on the
+     * created resource. This is stronger than [assertHasAtLeastOne], which only checks each call exists somewhere
+     * in the solution independently and in any order.
+     */
+    private fun assertPostThenPatchInSameTest(solution: Solution<RestIndividual>) {
```
arcuri82, line 79: this could be generalized, with 2 input verbs, and moved up to `RestTestBase`

arcuri82 (changes requested):

suarezrominajulieta:
Hi @jgaleotti and @arcuri82!

I’m closing this PR because it has been superseded by #1603.

When I opened this PR, the goal was to address Andrea’s earlier request by ensuring that a generated test combining a POST request (rendered as a DTO) and a PATCH request (rendered as a raw JSON string) compiled correctly. At that point, JSON Patch DTO conversion had not been implemented yet, so even with dtoForRequestPayload=true, PATCH requests still fell back to strings. This PR simply verified that behavior.

Since then, #1603 has implemented full JSON Patch DTO conversion. With dtoForRequestPayload=true, PATCH request bodies are now correctly rendered as List<JsonPatchOperation> instead of raw JSON strings, and the implementation is covered by both unit and E2E tests for primitive and complex value scenarios.

Because of that, the assertion in this PR (assertFalse { generatedCode.contains("JsonPatchOperation") }) now validates behavior that is no longer correct. After merging #1603, this test would fail by design.

The fallback behavior (dtoForRequestPayload=false → raw JSON string) was already implemented before either PR and remains unchanged.

Sorry for the confusion, and thanks for your patience!

PR #1633, opened 2026-07-07
Centralize Postgres container initialization utilities

jgaleotti:
Fixes bug causing "FATAL: the database system is starting up" error during container setup.


core-tests/e2e-tests/spring/spring-rest-postgres/src/test/kotlin/com/foo/spring/rest/postgres/PostgresContainerUtils.kt
```diff
@@ -0,0 +1,48 @@
+package com.foo.spring.rest.postgres
```
arcuri82: we should really avoid code-duplication.
here, the same class is repeated 3 times.
better to put it in its own module (`client-java/postgres-test-utils`), and import it where needed

arcuri82 (changes requested):

client-java/postgres-test-utils/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <artifactId>evomaster-client-java-postgres-test-utils</artifactId>
+    <packaging>jar</packaging>
+    <name>${project.groupId}:${project.artifactId}</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>provided</scope>
```
arcuri82: why `provided`?

arcuri82 (changes requested):

client-java/postgres-test-utils/pom.xml
jgaleotti: I followed the same solution that in client-java/test-utils-java/pom.xml for resolving testcontainers. I cannot use the <scope>test</scope> in this case because class PostgresContainerUtils is located within src/main/java. Should I do something different in this scenario?

arcuri82: @jgaleotti  remove the `provided`. the default `compile` is fine in this case, as it is a test utility module. `client-java/test-utils-java/pom.xml` is very special, and `provided` there has its reason... which i guess i should write down in a comment there :)

arcuri82 (approved):

PR #1640, opened 2026-07-13
Json Patch and XML key config or feature flag

suarezrominajulieta:
We create this pr to introduce 2 keys to compare coverage of endpoints with and without json patch and xml support.


suarezrominajulieta:
hi @jgaleotti, I created this PR as we discussed, to make the two keys available: the JSON patch key and the XML key. I'm not sure how to use them for analysis, so for now I created two new tests that return false when the keys are disabled, thus verifying that the endpoints are not covered.

PR #1646, opened 2026-07-20
Java regex simple lookaheads

lmasroca:
Added support for basic java regex lookaheads.

This PR introduces a check to `RegexGene`s for which the `RegexType` is `JVM`, in which we check its randomized value after each `randomize()` for a match against its compiled java pattern. If the sampled value does not match we trigger repairs, which aim at repairing the assertions in the regex. These are repaired by sampling a value from the lookahead's internal `DisjunctionListRxGene` and attempting to force that value into the following genes. This is safely done as we check if the genes can actually take that value or not before forcing them. We attempt these repairs a couple of times before giving up and throwing.

We still do not support nested lookaheads (nested inside a group or nested within another lookahead) as the repair mechanism can not handle them yet. The current repairs this PR introduces can only repair regex genes that are on the same level as the lookahead itself (the `terms` of a `DisjunctionRxGene`). This also does not yet properly handle multiple concatenated lookaheads as the current repair logic steps over the previous repairs if a new repair is triggered.

Note: this PR adds a new gene `AssertionRxGene` as well as a new interface `RxAbsorbable` which is implemented by regex genes.


core/src/test/kotlin/org/evomaster/core/parser/RegexHandlerTest.kt
```diff
@@ -14,7 +14,6 @@ import java.util.regex.Pattern
 
 internal class RegexHandlerTest{
 
-    @Disabled("Needs to hande lookahead in regex")
```
lmasroca, line 17: This test needed `(?iu)`, which is not actually lookahead but embedded flags (in particular Unicode case insensitive), which we support already.

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
```diff
@@ -87,6 +89,24 @@ class GeneRegexJavaVisitor(externalRegexFlags: RegexFlags = RegexFlags()) : Rege
         return disjList
     }
 
+    /**
+     * Walks up [ctx]'s ancestry towards the top-level pattern, throwing as soon as it
+     * finds one of the currently-unsupported ways an assertion's ancestry can appear (nested).
+     * Returns normally if it reaches the top-level pattern without hitting either.
+     */
+    private fun rejectIfNestedAssertion(ctx: RegexJavaParser.AssertionContext) {
```
jgaleotti: Shouldn't this method return a boolen and then took action if nested assertions are found?

core/src/main/kotlin/org/evomaster/core/parser/GeneRegexJavaVisitor.kt
```diff
@@ -231,7 +253,16 @@ class GeneRegexJavaVisitor(externalRegexFlags: RegexFlags = RegexFlags()) : Rege
         val res = VisitResult()
 
         if(ctx.assertion() != null){
-            res.data = ctx.assertion().text
+            val assertionCtx = ctx.assertion()
+            if (assertionCtx.CARET() != null || assertionCtx.DOLLAR() != null) {
+                res.data = ctx.assertion().text
+            } else {
+                rejectIfNestedAssertion(ctx.assertion())
```
jgaleotti: Should the exception be thrown here? Is that an illegalArgumentException since ctx is a parameter?

core/src/main/kotlin/org/evomaster/core/search/gene/regex/AnyCharacterRxGene.kt
```diff
@@ -131,4 +131,25 @@ class AnyCharacterRxGene(
         return true
     }
 
+    override fun absorbableCount(value: String): Int {
```
jgaleotti, line 139: Add javadoc for these methods

core/src/main/kotlin/org/evomaster/core/search/gene/regex/AssertionRxGene.kt
```diff
@@ -0,0 +1,108 @@
...
+ * This gene is placed in the tree at the position where the assertion appeared in the
+ * source regex. It produces no characters ([getValueAsPrintableString] always returns
+ * "").
+ *
+ * Repair is triggered from [DisjunctionRxGene.attemptAssertionRepair], invoked by
+ * [RegexGene.randomize] after the disjunction's own sampled value is checked against
+ * the source pattern and found not to match.
+ */
+class AssertionRxGene(
+    val innerGene: DisjunctionListRxGene?
```
jgaleotti: why is the innerGene nullable  ? Add some javadoc to explain it.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
```diff
@@ -180,4 +187,36 @@ class CharacterRangeRxGene(
...
+        if (flags.isCaseable(c) && validRanges.contains(c.swapCase())) {
+            return 1
+        }
+        return 0
+    }
+
+    override val canBeZeroWidth: Boolean = false
+
+    override fun tryForce(value: String): Int {
+        require(value.isNotEmpty())
```
jgaleotti, line 222: is this require always used? does it throw illegalArgumentException if not satisfied?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/CharacterRangeRxGene.kt
lmasroca: `require(value: Boolean)` throws `IllegalArgumentException` when `value` is false, no-op if `value` is true. Here it should always be true unless we have a bug, as `tryForce`'s call sites should not allow this.

lmasroca: Should we replace `require(bool)` usages with `if(!bool) throw IllegalArgumentException`?

jgaleotti: no need

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/AssertionRxGene.kt
```diff
@@ -0,0 +1,126 @@
...
+ * Repair is triggered from [DisjunctionRxGene.attemptAssertionRepair], invoked by
+ * [RegexGene.randomize] after the disjunction's own sampled value is checked against
+ * the source pattern and found not to match.
+ */
+class AssertionRxGene(
+    /**
+     * The assertion's inner disjunction gene, can be null as the disjunction can be unsatisfiable,
+     * in that case [innerGene] is null.
+     */
+    val innerGene: DisjunctionListRxGene?
```
arcuri82: is the handling of this inner gene correct? i don't see passed as input to parent's constructor, nor added directly to children in an `init` method

core/src/main/kotlin/org/evomaster/core/search/gene/regex/AssertionRxGene.kt
```diff
@@ -0,0 +1,126 @@
...
+) : RxTerm, SimpleGene("assertion") {
```
arcuri82: the use of `SimpleGene` here seems incorrect, as this gene has a `innerGene`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/AssertionRxGene.kt
lmasroca: Should this be a `CompositeFixedGene` instead? If so how would we handle `innerGene` being null?

arcuri82: i guess could use `CompositeGene`?

lmasroca: Okay, we can probably use `listOfNotNull(innerGene)` and `CompositeGene` (or `CompositeFixedGene` if we allow it to be childless, as once created the number of children does not mutate). Should the same be done for `BackReferenceRxGene` too?

lmasroca: For now made both `AssertionRxGene` and `BackReferenceRxGene` both extend `CompositeFixedGene` with `canBeChildless() = true`, as separate commits. These use `listOfNotNull(nullableGene)` to set their children (either no children or a single child).

lmasroca: Reverted the `BackReferenceRxGene` extending `CompositeFixedGene` (instead of `SimpleGene`) commit.
`AssertionRxGene` owns its `innerGene` while `BackReferenceRxGene` does not own its `captureGroup`, it is just a reference to a `DisjunctionListRxGene` that exists elsewhere in the regex gene tree.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
```diff
@@ -29,21 +36,59 @@ class RegexGene(
...
+
     override fun randomize(randomness: Randomness, tryToForceNewValue: Boolean) {
         usingFixedValue = if(fixedValue == null){
             false
         } else {
             randomness.nextBoolean()
         }
-        disjunctions.randomize(randomness, tryToForceNewValue)
+
+        if (regexType != RegexType.JVM) {
```
arcuri82: maybe for readability this should be refactored with a method, like:
`if(! requiresAssertionHandling(regexType))`

arcuri82: or add a comment to explain it. otherwise unclear why we are single out JVM

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
```diff
@@ -91,6 +91,11 @@ class RegexGene(
         throw IllegalStateException("Could not repair regex value")
     }
 
+    /**
+     * If regex requires its assertions to be repaired, currently only JVM regexes.
+     */
+    private val requiresAssertionHandling: Boolean = regexType == RegexType.JVM
```
arcuri82: recall from `docs/for_developers.md`: 
> When writing a new class (or re-factoring a current one), fields should come first, 
followed by class constructors and then the other methods.

shouldn't have a field declaration between methods

arcuri82 (changes requested):

arcuri82 (approved):

PR #1641, opened 2026-07-14
Neo4j heuristics calculator

andyfelder16:

PR #1647, opened 2026-07-21
add Neo4j heuristics calculator

andyfelder16:
First of a few PRs splitting #1641, per review feedback.

- Computes `H(Q,G)`: how close a graph is to satisfying a parsed Cypher MATCH query, as a Truthness `<ofTrue, ofFalse>`
- Neo4jHeuristicsCalculator, Neo4jStructuralMatcher (`matched_elements`), Neo4jConditionEvaluator (condition + operand evaluation), Neo4jMapping
- Neo4jNode/Neo4jEdge/Neo4jGraph: in-memory graph model
- Works in Truthness end-to-end, mirroring the SQL heuristics calculator
- 11 tests


client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+            return aggregate(((OrCondition) condition).getConditions(), mapping, false);
+        }
+        if (condition instanceof XorCondition) {
+            return evaluateXor(((XorCondition) condition).getConditions(), mapping);
+        }
+        if (condition instanceof NotCondition) {
+            Truthness inner = evaluateCondition(((NotCondition) condition).getCondition(), mapping);
+            return inner == null ? null : inner.invert();
+        }
+        return null;
```
jgaleotti: instead of null shouldn't this throw an exception signalling that a case was not supported?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+
+    Neo4jConditionEvaluator(TaintHandler taintHandler) {
+        this.taintHandler = taintHandler;
+    }
+
+    /**
+     * Returns {@code ρ(condition, mapping)}, or {@code null} when the condition cannot be evaluated
+     * and must be skipped by the aggregation.
+     */
+    Truthness evaluateCondition(CypherCondition condition, Neo4jMapping mapping) {
```
jgaleotti: have you considered applying a visitor pattern to CypherCondition AST ?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+        if (taintHandler != null && a != null && b != null) {
+            taintHandler.handleTaintForStringEquals(a, b, false);
+        }
+    }
+
+    private Truthness numericLessThan(Object a, Object b) {
+        Double da = asDouble(a);
+        Double db = asDouble(b);
+        if (da == null || db == null) {
+            return null;
```
jgaleotti: is null a valid input for this method?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+        PropertyOperand po = (PropertyOperand) operand;
+        Object value = resolveProperty(po.getVariableName(), po.getPropertyKey(), mapping);
+        boolean present = value != UNRESOLVED && value != null;
+        boolean satisfied = wantPresent == present;
+        return satisfied ? TRUE_TRUTHNESS : FALSE_TRUTHNESS;
+    }
+
+    private Truthness equalityTruthness(Object a, Object b) {
+        if (a == null || b == null) {
+            return null;
```
jgaleotti: what if two references are null? what is the Neo4J semantics for this?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+
+    /**
+     * Direct string equality truthness: {@code TRUE} when the strings are equal, otherwise {@code ofTrue}
+     * is the left-alignment edit distance scaled from base {@code C} (so it never drops below {@code C})
+     * and {@code ofFalse = 1}. Used where a string is compared for equality on its own — relationship
+     * types and string-valued property/WHERE equality.
+     */
+    static Truthness stringEqualityTruthness(String a, String b) {
+        if (a == null || b == null) {
+            return null;
```
jgaleotti: is a and b nullable? what is the semantics?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,419 @@
...
+
+    private static Double asDouble(Object value) {
+        if (value instanceof Number) {
+            return ((Number) value).doubleValue();
+        }
+        return null;
+    }
+
+    static Truthness getStartsWith(String str, String prefix) {
+        if (str.startsWith(prefix)) {
```
jgaleotti, line 475: add Objects.requireNonNull(...) whenever the arguments are intended not to be null.

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,190 @@
...
+        Truthness hWhere = computeHeuristicWhere(conditions, mappings);
+        return TruthnessUtils.buildAndAggregationTruthness(hMatch, hWhere);
+    }
+
+    /**
+     * Converts a heuristic to the distance form: {@code 1 - ofTrue}, in
+     * {@code [0,1]}, where 0 means the query is satisfied.
+     */
+    public double computeDistance(MatchOperation query, Neo4jGraph graph) {
+        return 1.0d - computeHeuristic(query, graph).getOfTrue();
```
jgaleotti: add a variable to store computeHeuristic(query, graph)

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,190 @@
...
+        Truthness nodes = computeHeuristicMatchNodes(pattern.nodeCount(), graph.nodeCount());
+        Truthness edges = computeHeuristicMatchEdges(pattern.getEdges(), graph, mappings);
+        return TruthnessUtils.buildAndAggregationTruthness(nodes, edges);
+    }
+
+    /**
+     * Count-based node availability: enough graph nodes to bind the pattern's nodes. Pure cardinality,
+     * no label/property check (those are conditions evaluated by H_where).
+     */
+    Truthness computeHeuristicMatchNodes(int required, int available) {
```
jgaleotti: can values be negative? Add checks with IllegalArgumentException to signal invalid inputs.

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jMapping.java
```diff
@@ -0,0 +1,68 @@
...
+    private Neo4jMapping(Map<String, Neo4jNode> nodeBindings, Map<String, Neo4jEdge> edgeBindings) {
+        this.nodeBindings = new LinkedHashMap<>(nodeBindings);
+        this.edgeBindings = new LinkedHashMap<>(edgeBindings);
+    }
+
+    Neo4jMapping copy() {
+        return new Neo4jMapping(nodeBindings, edgeBindings);
+    }
+
+    Neo4jNode getNode(String variable) {
```
jgaleotti: add Objects.requireNonNull(...) if needed

jgaleotti (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+        Objects.requireNonNull(mapping, "mapping must not be null");
+        return condition.accept(new TruthnessVisitor(mapping));
+    }
+
+    /**
+     * Computes {@code ρ} for one condition under a fixed mapping. A fresh instance is created per
+     * top-level condition (it carries the mapping), and it recurses through {@link #evaluateCondition}
+     * for the boolean-tree children, so nested conditions dispatch through the same visitor.
+     */
+    private final class TruthnessVisitor implements CypherConditionVisitor<Truthness> {
```
jgaleotti: shouldn't this be a public class?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jMapping.java
jgaleotti: can variable be null here? what would that mean?

jgaleotti (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
andyfelder16: It's private because it's nested inside a class that's already package-private (`Neo4jConditionEvaluator`), the only thing `heuristics/` exposes is `Neo4jHeuristicsCalculator`, as the single public API. I think making it public would not change much here, since a nested class can't be more visible than the class holding it. If you think it's better to expose the visitor, I'll change it

jgaleotti:
@andyfelder16  is this ready for re-review? Please do not forget to ask for a re-review whenever ready.

jgaleotti (approved):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/data/Neo4jEdge.java
```diff
@@ -0,0 +1,63 @@
...
+ * endpoint nodes (source and target), and its property map. A relationship in Neo4j is always stored
+ * with a direction (source → target).
+ */
+public class Neo4jEdge {
+
+    private final String id;
+    private final String type;
+    private final String sourceId;
+    private final String targetId;
+    private final Map<String, Object> properties;
```
arcuri82, line 39: add JavaDocs for each of these fields. also, for collections, you must always specify in comments what the `key` and `value` are.  eg, in `Map<String, Object>`, what are used as keys?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/data/Neo4jNode.java
```diff
@@ -0,0 +1,59 @@
...
+ * <p>
+ * This is a plain data holder built either by hand (tests) or by reading the live database; it has
+ * no relation to {@link org.evomaster.client.java.controller.neo4j.operations.PatternNode}, which is
+ * the structural node of a parsed query pattern.
+ */
+public class Neo4jNode {
+
+    private final String id;
+    private final Set<String> labels;
+    private final Map<String, Object> properties;
```
arcuri82, line 40: javadocs

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+            case IS_NOT_NULL:
+                return presenceTruthness(cc.getLeft(), mapping, /*wantPresent=*/true);
+            case IN:
+                return evaluateIn(cc, mapping);
+            case STARTS_WITH:
+            case ENDS_WITH:
+            case CONTAINS:
+                return evaluateStringPredicate(cc, mapping);
+            default:
+                return evaluateBinaryComparison(cc, mapping);
```
arcuri82, line 186: `default` is usually to throw exception, as each case should be explicit. eg, if one day a new operator is added, this should crash, and not silently handle it as a binary comparison

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+            case LESS_THAN_OR_EQUALS: {
+                Truthness t = numericLessThan(r, l);
+                return t == null ? null : t.invert();
+            }
+            case GREATER_THAN_OR_EQUALS: {
+                Truthness t = numericLessThan(l, r);
+                return t == null ? null : t.invert();
+            }
+            default:
+                return null;
```
arcuri82: this should throw an exception

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+        }
+        switch (cc.getOperator()) {
+            case STARTS_WITH:
+                return getStartsWith((String) l, (String) r);
+            case ENDS_WITH:
+                return getEndsWith((String) l, (String) r);
+            case CONTAINS:
+                return getContains((String) l, (String) r);
+            default:
+                return null;
```
arcuri82: as for all of these switches, throw an exception, eg, `throw new IllegalArgumentException("Not supported operator: " + cc.getOperator())`, with a comment. unless returning null is part of the expected, normal behavior of this function if called with a non-supported operator. if so, needs to add a clarifying comment

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+    /**
+     * ρ for an equality {@code a = b} where either side is an already-valuated operand. The
+     * right-hand value may be the Cypher {@code null} literal (a {@link LiteralOperand} whose value
+     * is {@code null}); a {@code null} on either side yields {@code null} here, mirroring Cypher's
+     * ternary logic where any comparison against {@code null} is {@code null} (unknown) rather than
+     * true or false, so the aggregation skips it instead of scoring a gradient.
+     */
+    private Truthness equalityTruthness(Object a, Object b) {
+        if (a == null || b == null) {
+            return null;
```
arcuri82: in line of princple, i don't like passing `null` like this when computing heuristics score.  in `DistanceHelper`, we have things like `H_REACHED_BUT_NULL` which is lower than `H_NOT_NULL`. i think we should "score a gradient" in these cases, and not rely on null

arcuri82: if it makes sense, you could even introduce a new value that is lower than `H_REACHED_BUT_NULL` (but still > 0), eg `H_INVALID`, and have a special new boolean field in `Truthness` called like `invalid`. if a truthness object is `invalid`, its heuristics is `H_INVALID`. any operation on invalid (eg AND or OR) would result in an `invalid` result. this would avoid passing around `null`. @jgaleotti  what do you think?

arcuri82: introducing this new `H_INVALID` could be done in a new `PR` before this one is merged

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
arcuri82: i m not going to write it in all functions, but we need to remove all these `null`

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,486 @@
...
+     * Direct string equality truthness: {@code TRUE} when the strings are equal, otherwise {@code ofTrue}
+     * is the left-alignment edit distance scaled from base {@code C} (so it never drops below {@code C})
+     * and {@code ofFalse = 1}. Used where a string is compared for equality on its own — relationship
+     * types and string-valued property/WHERE equality.
+     * <p>
+     * Both arguments must be non-null: this is only reached with two resolved strings (a relationship
+     * type, or two string values already confirmed present by {@link #equalityTruthness}). A Cypher
+     * {@code null} literal never reaches here — it is handled one level up by {@link #equalityTruthness}.
+     */
+    static Truthness stringEqualityTruthness(String a, String b) {
```
arcuri82: any reason for `package` level scope here? should be `public` unless some specific reason. we usually avoid package level scoping. this apply to all methos in this PR (i ll not repeat this comment in all places)

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jMapping.java
```diff
@@ -0,0 +1,91 @@
...
+ * variable to a graph relationship ({@code ε}). Bindings are keyed by the variable name the parser
+ * assigned (a real name like {@code n}, or an anonymous {@code _anon_node_0} / {@code _anon_rel_0}),
+ * which is shared between the {@code PatternEdge}/{@code PatternNode} and the conditions that refer
+ * to it, so condition valuation can resolve a variable to its bound graph element.
+ */
+class Neo4jMapping {
+
+    private final Map<String, Neo4jNode> nodeBindings;
+    private final Map<String, Neo4jEdge> edgeBindings;
+
```
arcuri82, line 34: what are the `keys` in these 2 maps?

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
andyfelder16: Agreed on not returning `null`, and scoring a low value instead would already help the gradient.

The part I'm less sure about is having `invalid` propagate through AND/OR. Small example:

```cypher
MATCH (p:Person) WHERE p.age = 25 AND p.nickname = "bob"
```

against a node `(:Person {age: 27})`, which has no `nickname` property. The conditions score:

| condition | ρ (ofTrue) | |
|---|---|---|
| `p:Person` | `1.0` | label matches |
| `p.age = 25` | `0.333` | `1 - normalizeValue(\|27-25\|)` |
| `p.nickname = "bob"` | ? | property is absent, nothing to measure |

Depending on what we do with that third one:

- today, skipped: `avgOfTrue = (1 + 0.333) / 2 = 0.667`, and mutating age to 25 moves it to `1.0`
- scored low and kept in the average: `avgOfTrue = (1 + 0.333 + 0.05) / 3 = 0.461`, and the same mutation moves it to `0.683`
- `invalid` absorbing the AND: the whole conjunction is `H_INVALID`, and mutating age changes nothing

So a low `ofTrue` that still takes part in the aggregation gives us the "no nulls" property without losing the gradient, while an absorbing `invalid` loses it exactly on the queries where some predicate can't be valuated.

Happy to go either way, just let me know which one you prefer and I'll implement it

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
andyfelder16: Changed it. Andrea pointed out that the project avoids package level scoping in general, so `Neo4jConditionEvaluator` is public now

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
arcuri82: @andyfelder16  indeed, we should not lose the gradient! ;)
however, I need some clarifications here. Assume same query but with an `OR`:

`MATCH (p:Person) WHERE p.age = 25 OR p.nickname = "bob"`

now, consider the node `(:Person {age: 25})`... what would be the result here? `null`? we need to handle such case properly (and have at least 1 test to verify it)

andyfelder16: Not null, it gives `<1.0, 0.1>` and distance `0`, which matches Cypher semantics: `true OR null` is `true`, and Neo4j does return that node. The unvaluatable branch is dropped and the OR is satisfied by the age one

I also ran the neighbouring cases to see where the current skipping breaks down:

| query | node | distance today | Neo4j returns |
|---|---|---|---|
| `p.age = 25 OR p.nickname = "bob"` | `{age: 25}` | `0.0` | the node |
| `p.age = 25 AND p.nickname = "bob"` | `{age: 25}` | `0.0` | nothing |
| `p.nickname = "bob"` | `{age: 25}` | `0.0` | nothing |

The last two are false positives. In Cypher `true AND null` is `null`, so neither query matches anything, but we report them as fully satisfied. That is the worst case for the search

My take is that this settles it in favour of your suggestion. Scoring a low value instead of skipping fixes both cases: taking `0.05` as that value, the AND goes from `0.0` to about `0.107`, while the OR stays at `0.0` because the aggregation still picks up the satisfied branch

I'll add tests for the three cases above

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+     * Converts a heuristic to the distance form: {@code 1 - ofTrue}, in
+     * {@code [0,1]}, where 0 means the query is satisfied.
+     */
+    public double computeDistance(MatchOperation query, Neo4jGraph graph) {
+        Truthness heuristic = computeHeuristic(query, graph);
+        return 1.0d - heuristic.getOfTrue();
+    }
+
+    private Truthness computeHeuristicPattern(MatchPattern pattern, Neo4jGraph graph, List<Neo4jMapping> mappings) {
+        Truthness nodes = computeHeuristicMatchNodes(pattern.nodeCount(), graph.nodeCount());
```
jgaleotti: rename to hNodes and hEdges. Otherwise it might be confused with the actual nodes and edges

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+            if (t.getOfTrue() > maxOfTrue) {
+                maxOfTrue = t.getOfTrue();
+            }
+        }
+        return TruthnessUtils.buildScaledTruthness(C, maxOfTrue);
+    }
+
+    private Truthness edgesForMapping(List<PatternEdge> patternEdges, Neo4jGraph graph,
+                                      Neo4jMapping mapping) {
+        Truthness[] perEdge = new Truthness[patternEdges.size()];
```
jgaleotti: rename to truthnessPerEdge

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+     * {@code [0,1]}, where 0 means the query is satisfied.
+     */
+    public double computeDistance(MatchOperation query, Neo4jGraph graph) {
+        Truthness heuristic = computeHeuristic(query, graph);
+        return 1.0d - heuristic.getOfTrue();
+    }
+
+    private Truthness computeHeuristicPattern(MatchPattern pattern, Neo4jGraph graph, List<Neo4jMapping> mappings) {
+        Truthness nodes = computeHeuristicMatchNodes(pattern.nodeCount(), graph.nodeCount());
+        Truthness edges = computeHeuristicMatchEdges(pattern.getEdges(), graph, mappings);
```
jgaleotti: rename to hEdges or similar

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+                                      Neo4jMapping mapping) {
+        Truthness[] perEdge = new Truthness[patternEdges.size()];
+        for (int i = 0; i < patternEdges.size(); i++) {
+            perEdge[i] = edgeMatch(patternEdges.get(i), graph, mapping);
+        }
+        return TruthnessUtils.buildAndAggregationTruthness(perEdge);
+    }
+
+    /** Existence-only edge check: TRUE if some graph relationship matches the edge's endpoints. */
+    private Truthness edgeMatch(PatternEdge edge, Neo4jGraph graph, Neo4jMapping mapping) {
```
jgaleotti: edge -> patternEdge ?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+        Neo4jNode source = mapping.getNode(edge.getSourceVariable());
+        Neo4jNode target = mapping.getNode(edge.getTargetVariable());
+        if (source == null || target == null) {
```
jgaleotti: can source or taget be null?

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+            return FALSE_TRUTHNESS;
+        }
+        for (Neo4jEdge rel : graph.getEdges()) {
```
jgaleotti: rel is an actual edge

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * AND-aggregates the truthness of every condition under one mapping. Conditions that cannot be
+     * valuated score {@link Neo4jConditionEvaluator#UNVALUATABLE} and are aggregated like the rest, so
+     * they lower the result instead of being dropped. A query with no conditions is vacuously satisfied.
+     */
+    private Truthness matchConditions(List<CypherCondition> conditions, Neo4jMapping mapping) {
+        if (conditions.isEmpty()) {
+            return TRUE_TRUTHNESS;
+        }
+        List<Truthness> truths = new ArrayList<>();
```
jgaleotti: rename truths to listOfTruthness

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,192 @@
...
+        for (CypherCondition c : conditions) {
+            truths.add(evaluator.evaluateCondition(c, mapping));
```
jgaleotti: store evaluator.evaluateCondition to a local variable before adding to the list

jgaleotti (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
andyfelder16: I added 5 tests: the three cases from the table above, the NOT one, and one checking that a sibling's gradient survives.

Conditions that can't be valuated score `UNVALUATABLE` (`<H_REACHED_BUT_NULL, 1>`) instead of being skipped, and they take part in the aggregation like any other condition, so they lower the result without hiding the gradient of their siblings. NOT propagates the marker rather than inverting it, since NOT null is null in Cypher.

jgaleotti (approved):

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,517 @@
...
+ * structural mapping {@code m}, recursively over the typed boolean tree the parser produces
+ * (And/Or/Xor/Not, comparisons, label/type/property leaves).
+ * <p>
+ * A condition that cannot be valuated under the mapping (an absent property, an unbound variable, an
+ * opaque/raw operand) scores {@link #UNVALUATABLE} rather than {@code null}: a low but non-zero
+ * {@code ofTrue} that still takes part in the aggregation. Never returns {@code null}.
+ */
+public class Neo4jConditionEvaluator {
+
+    private static final Object UNRESOLVED = new Object();
```
arcuri82, line 38: what is this object for? add short javadoc for it

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jConditionEvaluator.java
```diff
@@ -0,0 +1,517 @@
...
+        if (l == null || r == null) {
+            return UNRESOLVED;
+        }
+        switch (ao.getOperator()) {
+            case PLUS: return l + r;
+            case MINUS: return l - r;
+            case TIMES: return l * r;
+            case DIVIDE: return r == 0d ? UNRESOLVED : l / r;
+            case MODULO: return r == 0d ? UNRESOLVED : l % r;
+            case POWER: return Math.pow(l, r);
```
arcuri82, line 330: what if `r` is 0? wouldn't those `/` and `%` throw exceptions? either way, add explicit test cases to check this

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -0,0 +1,199 @@
...
+ * <p>
+ * Expects {@code query} to already be in canonical form: no quantified path patterns and no
+ * variable-length edges (both are expanded to a plain pattern by a separate canonization step before
+ * reaching this calculator).
+ */
+public class Neo4jHeuristicsCalculator {
+
+    public static final double C = 0.1;
+    public static final Truthness TRUE_TRUTHNESS = new Truthness(1, C);
+    public static final Truthness FALSE_TRUTHNESS = new Truthness(C, 1);
```
arcuri82: some of these constants are now available in `TruthnessUtils` with the same name. should use those instead of redeclaring them here

arcuri82 (changes requested):

arcuri82 (approved):

PR #1649, opened 2026-07-21
Replacements for Spring Data for Apache Cassandra

gonzalotguerrero:


jgaleotti:
Hi @gonzalotguerrero , is this ready for review? I do not see that my review was requested yet.

gonzalotguerrero:
Hi @jgaleotti ! It is, sorry! I was waiting for the checks to finish running and then I forgot to request it :p

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/CassandraTemplateClassReplacement.java
```diff
@@ -0,0 +1,105 @@
...
+ * instantiates a {@code CassandraRepository}; a SUT calling {@code CassandraTemplate} directly (bypassing
+ * the repository layer) needs this replacement to have that mapping recorded too.
+ */
+public class CassandraTemplateClassReplacement extends ThirdPartyMethodReplacementClass {
+
+    private static final CassandraTemplateClassReplacement singleton = new CassandraTemplateClassReplacement();
+
+    @Override
+    protected String getNameOfThirdPartyTargetClass() {
+        return "org.springframework.data.cassandra.core.CassandraTemplate";
```
jgaleotti, line 29: Is it possible to monitor the schema directly from the cassandra library without having to have hooks in the spring-data library? For MongoDB this was done since the schema was lost from the spring-data library to the MongoDB library.

jgaleotti (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/CassandraTemplateClassReplacement.java
gonzalotguerrero: Hi! I've been doing some more research on the use of [Session.getMetadata()](https://apache.github.io/cassandra-java-driver/4.19.0/upgrade-README/?h=getmeta#metadata) we discussed in person.

Extracting the relevant columns directly from every intercepted query in `CqlSessionClassReplacement` initially sounds OK, but if there's no WHERE clause or if it does not contain all clustering columns, we'd be missing critical metadata, as all columns belonging to the table's primary key must be set during an insertion.

The `TableMetadata` interface inherits a `getPrimaryKey()` method from `RelationMetadata`. There, we could obtain the metadata for all columns that must be set in every insertion. Also these columns cannot be altered, so they could be cached.

Given that,  I think with the use of `getMetadata()` we could cache the primary key of every queried table, and, for every query against a table whose primary key has already been cached, only search for the metadata regarding the non-PK columns present in the WHERE clause, if any.

What do you think of this approach?

gonzalotguerrero:
Hi @jgaleotti !

I created a new PR in #1657 with a more appropriate branch/PR name, so I'll close this one

PR #1653, opened 2026-07-22
Fix REST path encoding, Java Unicode escaping, and monitor initialization

hoco-scy:
## Summary

This PR fixes four independent failures observed while generating and running EvoMaster tests across the WebFuzzing SUT suite:

- Resolved REST path fragments can fail or be corrupted for lone surrogates, supplementary Unicode characters, and non-ASCII characters.
- Java Mongo initialization source can contain literal backslash-u sequences that interact with Java Unicode preprocessing.
- Java request-body source can collapse literal backslash-u sequences into invalid or value-changing Unicode escapes.
- Disabled process monitoring still initializes and deletes shared monitor output state, creating a concurrent filesystem race.

## Changes

- Encode non-ASCII resolved REST path fragments by Unicode code point and use URI raw paths, preserving valid surrogate pairs and handling isolated surrogates safely.
- Escape Java-only Mongo EJSON literal backslash-u runs with octal backslash escapes while preserving genuine Unicode escapes and runtime values.
- Preserve doubled source backslashes for Java BODY output so literal backslash-u values survive Java Unicode preprocessing.
- Initialize process-monitor output only when process monitoring is enabled.

## Scope

- REST encoding applies to resolved/static path fragments, including fragments around OpenAPI link variables. Runtime link-variable values remain outside this change.
- Mongo octal escaping applies only to Java output. Kotlin keeps its existing escaping behavior.
- BODY escaping changes only Java JUnit 4/5 output. Kotlin, JavaScript, and Python behavior is unchanged.
- Process-monitor behavior changes only when the monitor is disabled.

## Tests

- Added and extended regression coverage for BMP, supplementary, and isolated-surrogate REST path values; static fragments around runtime REST link variables; literal and genuine Unicode sequences in Java Mongo EJSON; Java BODY escaping through the JSON writer; and disabled process-monitor output isolation.
- Local targeted suite: 64 tests passed with no failures, errors, or skips.
- Additional local core coverage passed for all modified areas. Cases requiring unavailable external LLM/HTTP services or uncached Docker images are left to the fork CI environment.
- Experimental validation confirmed the affected SUT generation and execution paths.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
@@ -295,6 +295,12 @@ object GeneUtils {
                 .replace("\b", "\\b")
                 .replace("\t", "\\t")
 
+        if (format.isJava()) {
+            // Java translates eligible Unicode escapes before tokenizing string literals.
+            // Keep the even source-backslash count so a backslash-u sequence stays literal.
+            return ret
+        }
+
         if (format.isKotlin()) return ret.replace("\$", "\\\$")
                 .replace("\\\\u", "\\u")
         //.replace("\$", "\${\'\$\'}")
--- a/core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
+++ b/core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
@@ -295,27 +295,20 @@ object GeneUtils {
                 .replace("\b", "\\b")
                 .replace("\t", "\\t")
 
-        if (format.isJava()) {
-            // Java translates eligible Unicode escapes before tokenizing string literals.
-            // Keep the even source-backslash count so a backslash-u sequence stays literal.
-            return ret
+        return when {
+            format.isKotlin() -> {
+                // JVM source must retain the doubled backslash. For example, input \uquWOnj must
+                // be emitted as "\\uquWOnj": "\uquWOnj" is invalid, while a valid sequence
+                // such as "\u0041" would silently change the runtime value to "A".
+                ret.replace("\$", "\\\$")
+            }
+            format.isJava() -> {
+                // Java translates eligible Unicode escapes before tokenizing string literals, so
+                // retaining the doubled source backslash also keeps the runtime sequence literal.
+                ret
+            }
+            else -> ret.replace("\\\\u", "\\u")
         }
-
-        if (format.isKotlin()) return ret.replace("\$", "\\\$")
-                .replace("\\\\u", "\\u")
-        //.replace("\$", "\${\'\$\'}")
-        //ret.replace("\$", "\\\$")
-        else return ret.replace("\\\\u", "\\u")
-
-        /*
-                   The \u denote unicode characters. For some reason, escaping the \\ leads to these being invalid.
-                     Since they are valid in the back end (and they should, arguably, be possible), this leads to inconsistent behaviour.
-                     This fix is a hack. It may be that some \u chars are not valid. E.g. \uAndSomeRubbish.
-
-                     As far as I understand, the addition of an \ in the \unicode should not really happen.
-                     They should be their own chars, and the .replace("\\", """\\""" should be fine, but for some reason
-                     they are not.
-                     */
     }
 
     private fun applySqlEscapes(string: String, format: OutputFormat): String {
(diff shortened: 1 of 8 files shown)


arcuri82:
hi @hoco-scy !
  you are a student of @man-zhang  isn't it?
  thanks for this PR ;)
  as Man told you about `docs/for_msc_bsc_students.md`? once CI passes, ask @man-zhang  for a review, and, when she will accept your updates/changes, she will ask for my review here on GitHub.
  also, no need to create entries under `issues` (that is mainly for external users), you can just discuss them here directly in the PR description
  thanks ;)

hoco-scy:
Hi @arcuri82,

Yes, I am a student of @man-zhang.

Thank you for the guidance. I apologize that I had not read `docs/for_msc_bsc_students.md` before opening this PR, so I followed my usual personal development workflow and created separate issues for the fixes.

I have now read the document and understand the expected internal contribution process. I will close the issues, keep the technical context in the PR description, make sure the CI process is completed on my fork, and then follow the review flow through @man-zhang.

Thank you again for your time and help.

core/src/test/kotlin/org/evomaster/core/output/MongoWriterTest.kt
```diff
@@ -0,0 +1,36 @@
+@file:Suppress("DEPRECATION")
```
man-zhang: remove unnecessary annotation

core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
```diff
@@ -295,6 +295,12 @@ object GeneUtils {
                 .replace("\b", "\\b")
                 .replace("\t", "\\t")
 
+        if (format.isJava()) {
```
man-zhang: This handle needs to move with the `format.isKotlin()`. 

This fix seems to relate to `ret.replace("\\\\u", "\\u")`. 
Does it also happen in Kotlin? can you add a test for Kotlin as well.

In addition, can you add an example in the comment to explain this fix?

man-zhang (changes requested):

core/src/test/kotlin/org/evomaster/core/output/MongoWriterTest.kt
hoco-scy: Removed the unnecessary `@file:Suppress("DEPRECATION")` annotation in fe79f8e6a7. I also confirmed that the affected tests compile and pass without it.

Thank you.

core/src/main/kotlin/org/evomaster/core/search/gene/utils/GeneUtils.kt
hoco-scy: Thank you for pointing this out. I investigated the Kotlin behavior and addressed it in fe79f8e6a7.

Kotlin was affected by the same issue. The previous Kotlin branch called `ret.replace("\\\\u", "\\u")`, which reduced the doubled source backslash to a single backslash-u sequence. As a result:

- a malformed sequence such as `\uquWOnj` could make the generated Kotlin source fail to compile;
- a valid sequence such as `\u0041` could compile but change the runtime value to `A`, instead of preserving the literal `\u0041` text.

I verified the fix in the following ways:

1. Extended `bodyEscapesKeepLiteralBackslashUForJvmFormats` to cover Java JUnit 4/5 and Kotlin JUnit 4/5.
2. Added `kotlinBodyWriterPreservesLiteralBackslashUAtRuntime`, which generates an actual Kotlin class, compiles it with `CompilerForTestGenerated`, loads the compiled class, invokes its methods, and checks the runtime string values.
3. Covered malformed backslash-u text, valid `\u0041` text, repeated backslashes, and Kotlin dollar-sign escaping.
4. Kept the existing non-JVM behavior for JavaScript and Python unchanged.

The verification confirms that the generated Kotlin source now compiles successfully and preserves the original runtime strings, including literal `\u0041`, rather than converting it to `A`.

I also added concrete malformed and valid examples to the source comment as requested.

hoco-scy:
Hi @man-zhang,

I have addressed both review comments in fe79f8e6a7:

- removed the unnecessary suppression annotation;
- investigated the Kotlin behavior and confirmed that it was affected by the same backslash-u issue;
- extended the fix to Kotlin output while preserving Kotlin dollar-sign escaping;
- added Kotlin JUnit 4/5 coverage and an actual compile-and-runtime regression test;
- added concrete examples to the source comment.

The latest full CI run for the same PR head commit has completed successfully: 22 jobs passed, with only the conditional `debug` job skipped.

CI run:
https://github.com/hoco-scy/EvoMaster/actions/runs/30246503337

Could you please take another look when convenient?

Thank you.

arcuri82 (approved):

PR #1654, opened 2026-07-23
External arazzo

arcuri82:


arcuri82:
hi @daniellopera15 (and @jgaleotti ) 
  thx! i ll be traveling till Monday. if it passes CI, i ll merge it once I am back next week.
I now invited you to the rep, so you can directly work here.
if CI will fail, you can modify directly this branch

daniellopera15:
@arcuri82 
Hi, how are you? Just letting you know that the CI failed, so I pushed a change regarding the POM versions. The CI has passed now. Does the change look good to you?

core-parent/pom.xml
```diff
@@ -86,16 +85,6 @@
                 <artifactId>swagger-annotations</artifactId>
                 <version>${swagger.annotations.version}</version>
             </dependency>
```
arcuri82, line 86: @daniellopera15  why are those dependencies removed here in `core-parent`?

daniellopera15: @arcuri82 I had to modify the dependencies because it was overwriting the Swagger version, causing the CI to fail.

I suppose this comment has something to do with it.

imagen

The parser currently has no limitations regarding the Swagger version. Does it look good as it is now, or where should I move the dependency?

core-parent/pom.xml
```diff
@@ -84,6 +85,11 @@
...
+            <dependency>
+                <groupId>com.graphql-java</groupId>
+                <artifactId>java-dataloader</artifactId>
+                <version>${java.dataloader}</version>
+            </dependency>
```
arcuri82: this shouldn't be here. why? dependencies conflicts shouldn't be handled by adding unrelated libraries (GraphQL in this case), but by library declaration order and pinned versions in the affected modules.

daniellopera15: I'm going to revert the POM to the original version. That was the dependency I cleaned up earlier, but I forgot to remove it from the parent.

daniellopera15:
@arcuri82 The change is ready for review

PR #1661, opened 2026-08-04
Java regex nested assertions

lmasroca:
Allowed assertions (currently only lookaheads #1646 and lookbehinds #1656) to appear nested (within a group `()`) on java regex. Allowed assertion requirements and repairs to escape their groups into neighboring terms. The outwards requirements of a group must be resolved by that group's neighboring genes during repairs for the repairs to be successful.

Note: Does not allow assertions to appear on quantified groups.


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/utils/AssertionRepairResult.kt
```diff
@@ -0,0 +1,60 @@
...
+            return FAILURE
+        }
+        return AssertionRepairResult(
+            success = true,
+            neededPrefix = next.neededPrefix ?: neededPrefix,
+            neededPostfix = next.neededPostfix ?: neededPostfix
+        )
+    }
+
+    companion object {
```
arcuri82: companion object should come in top of the class file, not on bottom

arcuri82 (changes requested):

arcuri82 (approved):

PR #1666, opened 2026-08-10
fix invalid location

omursahin:


core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -17,6 +17,19 @@ import org.evomaster.core.search.gene.wrapper.OptionalGene
 
 object HttpSemanticsOracle {
 
+    /**
+     * Verbs that the Location follow-up call may use (see HttpSemanticsService).
+     */
+    private val LOCATION_FOLLOWUP_VERBS = setOf(
```
arcuri82, line 24: this will not work as soon as we introduce the new verb `QUERY`.
add a `TODO` stating to update when `QUERY` is going to be added

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -729,12 +738,42 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+    private fun locationPathOnly(location: String): String? {
+        return try {
+            val p = java.net.URI(location).rawPath
+            if (p.isNullOrBlank()) null else p
+        } catch (e: Exception) {
+            null
+        }
+    }
+
+    private fun declaredPathMatching(resolvedPath: String): RestPath? {
```
arcuri82: a function like this should be in `CallGraphService`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/HttpSemanticsService.kt
```diff
@@ -729,12 +738,42 @@ class HttpSemanticsService : TimeBoxedPhase{
...
+
+    /**
+     * Verb to use for the Location follow-up call. If the Location points to a path
+     * declared in the schema, use its declared verbs by [locationFollowUpVerbPriority];
+     * otherwise default to GET.
+     */
+    private fun followUpVerbForLocation(location: String): HttpVerb {
+        val pathOnly = locationPathOnly(location) ?: return HttpVerb.GET
+        val declaredPath = declaredPathMatching(pathOnly) ?: return HttpVerb.GET
+        val verbs = actionDefinitions.filter { it.path == declaredPath }.map { it.verb }.toSet()
```
arcuri82: could use `endpointsForPath` in `CallGraphService`

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/CallGraphService.kt
```diff
@@ -70,6 +70,24 @@ class CallGraphService {
             .filter { it.path == path }
     }
 
+    /**
+     * Resolve a raw path or URL (e.g. a Location header value, absolute or relative)
+     * against the RestPath templates declared in the schema.
+     * Returns the most specific matching template (fewest path parameters), or null if none match.
+     */
+    fun resolveDeclaredPath(rawPathOrUrl: String): RestPath? {
```
arcuri82, line 78: we need integration tests for this function. see for example `CallGraphDeleteTest`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1667, opened 2026-08-10
improving info for 904 oracle

arcuri82:


core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -383,28 +417,53 @@ object HttpSemanticsOracle {
...
-                    if (valuePut != valueGet) return true
+            val fieldsPut = readPutFields(putBody, bodyParam, sentFields)
+            if (!fieldsPut.isNullOrEmpty()) {
+                val fieldsGet = readGetFields(getBody, fieldsPut.keys)
+                if(fieldsGet.isNullOrEmpty()){
+                    errors.addAll(fieldsPut.keys)
+                } else {
+                    for ((field, valuePut) in fieldsPut) {
+                        val valueGet = fieldsGet[field]
+                        //TODO but what if the values are 'null'??? was this correct?
```
omursahin: Yes, it should be correct because if it is null, the readPutFields function never returns that field. 

in src/main/kotlin/org/evomaster/core/output/formatter/OutputFormatter.kt line 89:

` if (value.isNull) continue
`

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -280,42 +280,76 @@ object HttpSemanticsOracle {
...
+                /*
+                    TODO what was motivation for this edge case?
+                    extra fields in GET that are not declared in PUT are ignored in this oracle...
+                    however, here, if there is not field in PUT, we do the opposite, and mark as
+                    failure if there are extra fields in GET?
+                    why this flipped behavior?
+                 */
+                return "PUT has no data to set, but GET returns data"
+            } else {
+                //TODO is this branch even possible?
```
omursahin: yes possible, if the put body is scalar like "just a text without name field", it happens. We can add this test to 
src/test/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracleTest.kt

```
    @Test
    fun testPut_scalarBody_noFieldInfoButPutSentData_returnsFalse() {
        val scalarTypeGene = EnumGene("contentType", listOf("application/json"))
        scalarTypeGene.index = 0
        val scalarBody = BodyParam(StringGene("body", "hello"), scalarTypeGene)

        val mismatch = runMismatchedPutOracle(
            path = "/users",
            putBody = scalarBody,
            getResponseBody = """{"name":"Alice"}"""
        )
        assertFalse(mismatch)
    }
```

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -280,42 +280,76 @@ object HttpSemanticsOracle {
...
+            assert(getBody.isNotEmpty())
             // no information to verify against; flag only when PUT sent nothing either
-            return putBody.isNullOrEmpty()
+            if(putBody.isNullOrEmpty()){
+                /*
+                    TODO what was motivation for this edge case?
+                    extra fields in GET that are not declared in PUT are ignored in this oracle...
+                    however, here, if there is not field in PUT, we do the opposite, and mark as
+                    failure if there are extra fields in GET?
+                    why this flipped behavior?
```
omursahin: I missed this behavior. I think it would be better not to mark it as a failure because it can lead to false positives.

I wrote that test but I write an assertion wrong:

```
    @Test
    fun testPut_noBodyParam_getHasServerDefaults_returnsFalse() {
        val mismatch = runMismatchedPutOracle(
            path = "/users",
            putBody = null,
            getResponseBody = """{"id":42,"name":"default","createdAt":"2026-01-01"}"""
        )
        assertTrue(mismatch)
    }
```

it should be false.

core/src/main/kotlin/org/evomaster/core/problem/rest/oracle/HttpSemanticsOracle.kt
```diff
@@ -280,42 +280,76 @@ object HttpSemanticsOracle {
...
+                return "PUT has no data to set, but GET returns data"
+            } else {
+                //TODO is this branch even possible?
+                return null
+            }
         }
 
         val wipedFields = computeWipedFields(allPutSchemaFields - sentFields, schema, get)
 
-        return hasMismatchedPutFields(putBody ?: "", getBody, sentFields, wipedFields, bodyParam)
```
omursahin, line 318: this is the only place that we use the hasMismatchedPutFields function so we can delete it.

PR #1673, opened 2026-08-11
Update SqlExpressionEvaluator.java

aschenzle:
Fixing a copy & paste mistake

### DIFF
--- a/client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
+++ b/client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
@@ -117,7 +117,9 @@ class SqlExpressionEvaluatorTest {
 
     }
 
-    private void assertSqlExpressionEvaluatesToTrue(String sqlCommand, QueryResult queryResult) {
+/* TODO: Check if we can remove this code
+
+   private void assertSqlExpressionEvaluatesToTrue(String sqlCommand, QueryResult queryResult) {
         Statement parsedSqlCommand = SqlParserUtils.parseSqlCommand(sqlCommand);
         Select select = (Select) parsedSqlCommand;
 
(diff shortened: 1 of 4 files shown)


jgaleotti (changes requested):
Please add a test case to exhibit the behaviour (i.e. a test case that fails since the expression checks only concreteLeftValue instead of concreteLeftValue and concreteRightValue

client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
```diff
@@ -117,7 +117,9 @@ void setUp() {
 
     }
 
-    private void assertSqlExpressionEvaluatesToTrue(String sqlCommand, QueryResult queryResult) {
+/* TODO: Check if we can remove this code
```
jgaleotti: if no longer used, please remove this

client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
```diff
@@ -1713,7 +1715,7 @@ public void testLowerCaseInsensitive() {
     }
 
     @Test
-    public void testUpperCaseInsenstive() {
+    public void testUpperCaseInsensitive() {
         String sql = "upper('hello')";
         Expression expression = assertDoesNotThrow(() -> CCJSqlParserUtil.parseExpression(sql));
```
jgaleotti, line 1696: I am a bit confused, where is the change to fix testEqualsToTimeWithTimeZone ?

jgaleotti (changes requested):

client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
aschenzle: No test was modified, the fix is in [ColumnTypeParser.java]

client-java/sql/src/test/java/org/evomaster/client/java/sql/heuristic/SqlExpressionEvaluatorTest.java
aschenzle: Done

jgaleotti (approved):

arcuri82 (approved):

PR #1681, opened 2026-08-14
AsyncAPI 3.x: parse a document's messages and their schemas

LautaroPetaccio:
> **Restructured following @arcuri82's review.** Everything that is not EvoMaster-specific now lives in a new module `core-extra/asyncapi-parser`, written in Java 8 under `com.webfuzzing.asyncapi`, following `arazzo-parser`. The module has **no dependency on anything in EvoMaster**, so it can be moved out into a library of its own. Nothing in EvoMaster depends on it yet either — the first consumer comes later in the stack.
>
> The stack shrank by one as a result: the PR that lifted `SchemaLocation` and friends into a shared package existed only so the Kotlin parser could reach them, and a self-contained module cannot. It has been closed, and the module carries its own `DocumentLocation` and `RefLocations`.

## What this is

The first step towards supporting **AsyncAPI** — the event-driven counterpart of OpenAPI, describing services that talk over Kafka, AMQP, MQTT or WebSocket — as a new problem type in EvoMaster.

This PR adds a parser for the **message layer** of an AsyncAPI 3.x document, and nothing else. There is no `ProblemType`, no sampler, no fitness function and no gene creation, so **nothing user-facing changes**: the code is additive, in a new module, and no existing file is touched apart from registering the module.

I am splitting the work into a stack of small PRs rather than one large one. This is the first; the others (channels and operations, references across documents, servers and protocol bindings, then the EvoMaster-side gene and action building) build on it. I would rather find out early if the direction or the conventions are wrong.

## Why a hand-written parser

What is needed is a small, well-understood subset of AsyncAPI: what a service can be sent, and the JSON Schema of it. It also has to keep going when a document is partly unusable, since real published documents routinely are. The available AsyncAPI libraries are all-or-nothing validators and pull in large dependency trees, so this reads the document with Jackson and takes only what is needed. A keyword it does not know is simply not read, and is never an error.

## The module

`core-extra/asyncapi-parser`, Java 8, `com.webfuzzing.asyncapi`. Its only dependencies are `jackson-databind` and `jackson-dataformat-yaml`.

| | |
|---|---|
| `access/AsyncApiAccess` | retrieval from a URL, a file or the classpath |
| `mapper/AsyncApiMapper` | one reader for both YAML and JSON, as JSON is valid YAML |
| `parser/AsyncApiParser` | document text to the model |
| `resolver/AsyncApiRefResolver` | local JSON Pointer resolution |
| `models/` | `AsyncApiDocument`, `AsyncApiMessage`, `AsyncApiCorrelationId`, `DocumentLocation` |

Models are immutable, with a nested builder where there are many optional fields, following `arazzo-parser`.

## Two decisions worth reviewing

**Payloads stay as JSON Schema, not genes.** A message payload is kept as a raw Jackson node with its `$ref` intact, and everything it can reach is checked to be present in `componentSchemas` — transitively, not just at the top level. So whatever builds genes later can resolve any reference it meets against that one map. A message whose schema reaches a reference that cannot be followed is dropped rather than handed on half-broken.

The transitive part is not theoretical. Avro payloads in the wild declare `schemaFormat` on the *component schema*, not on the payload, so a message can reference a perfectly good schema whose own properties reference one that had to be dropped. Checking only the payload's own `$ref` misses it.

**Failures are graded.** An unreadable document, the wrong version, or a document that is not AsyncAPI at all raises `AsyncApiParsingException`. Anything narrower costs only the element it affects and is recorded on `AsyncApiDocument.warnings`, so one exotic message cannot make a whole document unusable. Reporting those warnings is left to whatever consumes them, so that on the EvoMaster side they can reach the warnings aggregator the way REST's do, rather than being logged twice.

Only AsyncAPI 3.x is parsed. 2.x has no first-class reply and is rejected with a message saying so.

## Testing

37 unit tests in `AsyncApiParserTest`, `AsyncApiRefResolverTest` and `AsyncApiAccessTest`, with fixtures under `core-extra/asyncapi-parser/src/test/resources/asyncapi/artificial/`. Several of them exist because a real published document did what they describe: a payload written in Avro, a schema pointing at one that was dropped, references that go round in a circle (which took down the whole run with a `StackOverflowError` before there was a guard).

The parser has also been run over ~460 real AsyncAPI documents collected from public repositories. That corpus is too large to check in, so it is not part of this PR; a later one in the stack adds a handful of representative documents as fixtures.


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/schema/AsyncApiAccess.kt
```diff
@@ -0,0 +1,145 @@
+package org.evomaster.core.problem.asyncapi.schema
```
arcuri82: hi @LautaroPetaccio  @jgaleotti  looks like we forgot to specify it... but we should follow same approach as all the other parsers (eg Arazzo and Overlay). all code that is not specific to EvoMaster should be in a new module called `asyncapi-parser` under `core-extra`, and should be in Java (best if 8), not Kotlin, with goal in future to move it out of EvoMaster.  package name should be `com.webfuzzing.asyncapi`. sorry if this info was lost / not told :(  after that is done, i can review the changes

arcuri82 (changes requested):

jgaleotti (approved):

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/access/AsyncApiAccess.java
```diff
@@ -0,0 +1,195 @@
...
+import java.nio.file.Paths;
+import java.util.Locale;
+
+/**
+ * Reads AsyncAPI documents and hands them to {@link AsyncApiParser}.
+ *
+ * A document may come from a URL, from disk, or from the classpath, and where it came from is
+ * remembered so that references to neighbouring documents can be followed.
+ */
+public class AsyncApiAccess {
```
arcuri82, line 27: leave a `TODO` comment here, stating that this class will likely need to be refactored in the future (it is fine for now though). point is that retrieving schema is a common functionality. we could have it in EM, or as a shared third-party library that is re-used by the different document parsers. but it is not an architectural decision I want to make now

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/mapper/AsyncApiMapper.java
```diff
@@ -0,0 +1,51 @@
...
+ * limits are far too low for real documents, so they are raised here, in one place rather than
+ * at each call site.
+ */
+public class AsyncApiMapper {
+
+    private static final int CODE_POINT_LIMIT = 50 * 1024 * 1024;
+
+    private static final int MAX_ALIASES_FOR_COLLECTIONS = 1000;
+
+    private static final int NESTING_DEPTH_LIMIT = 100;
```
arcuri82, line 33: add TODO comment stating that, once parser is more stabilized, those could be configuration params, instead of hard-coded constants

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiDocument.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private final String rawText;
+
+    private final DocumentLocation sourceLocation;
+
+    private final String version;
+
+    private final String defaultContentType;
+
+    private final Map<String, AsyncApiMessage> messages;
```
arcuri82, line 51: recall what written under `docs/for_developers.md`: "for `Map`s, should add a comment stating what is the _key_, and what is the _value_". do it for all map fields in this PR

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
```diff
@@ -0,0 +1,586 @@
...
+    private static final String REF = "$ref";
+
+    /**
+     * Schema formats that are JSON Schema by another name, and so can be read here. Compared as
+     * prefixes, since the format string carries a version suffix.
+     *
+     * Anything else -- Avro and Protobuf being the ones that actually turn up -- describes a
+     * payload in a language this parser does not speak, and the message is dropped.
+     */
+    private static final List<String> JSON_SCHEMA_FORMATS = Arrays.asList(
```
arcuri82, line 57: what about `application/json` and `application/yaml`?

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
```diff
@@ -0,0 +1,586 @@
...
+
+        return resolved;
+    }
+
+    /**
+     * Every field of a node, with any {@code $ref} followed -- both one on the node itself and
+     * one on each of its values. Bindings are routinely shared through {@code components}, and
+     * reading them without following the reference is worse than not reading them at all.
+     */
+    private static Map<String, JsonNode> dereferencedFields(
```
arcuri82, line 404: what are the `keys` of this returned map? add such info in the javadoc comment above

arcuri82 (changes requested):

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/access/AsyncApiAccess.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/mapper/AsyncApiMapper.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiDocument.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
LautaroPetaccio: Added both. They're content types rather than schemaFormat values, which is why they weren't there, but hand-written documents use them here anyway, and the only sensible reading is JSON Schema. Rejecting them was dropping messages that parse fine.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
LautaroPetaccio: Done.

arcuri82 (approved):

PR #1682, opened 2026-08-15
Arazzo Sampler V1 

daniellopera15:


core/src/main/kotlin/org/evomaster/core/problem/rest/service/module/ArazzoRestModule.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.evomaster.core.problem.rest.service.sampler.ArazzoSampler
+import org.evomaster.core.remote.service.RemoteController
+import org.evomaster.core.remote.service.RemoteControllerImplementation
+import org.evomaster.core.search.service.FitnessFunction
+import org.evomaster.core.search.service.Sampler
+import org.evomaster.core.search.service.mutator.Mutator
+import org.evomaster.core.search.service.mutator.StandardMutator
+import org.evomaster.core.search.service.mutator.StructureMutator
+
+class ArazzoRestModule(private val bindRemote: Boolean = true) : RestBaseModule() {
```
jgaleotti: Add @Javadoc of this class

core/src/main/kotlin/org/evomaster/core/problem/rest/service/mutator/ArazzoStructureMutator.kt
```diff
@@ -0,0 +1,29 @@
+package org.evomaster.core.problem.rest.service.mutator
+
+import org.evomaster.core.problem.api.service.ApiWsStructureMutator
+import org.evomaster.core.search.EvaluatedIndividual
+import org.evomaster.core.search.Individual
+import org.evomaster.core.search.service.mutator.MutatedGeneSpecification
+import org.evomaster.core.sql.SqlInsertBuilder
+
+class ArazzoStructureMutator : ApiWsStructureMutator() {
```
jgaleotti: I am not sure if we need a specific mutator for Arazzo? I think we need to discuss this with Andrea

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -1,6 +1,9 @@
 package org.evomaster.core.problem.rest.service.sampler
 
 import com.google.inject.Inject
+import com.webfuzzing.arazzo.access.ArazzoAccess
+import com.webfuzzing.arazzo.models.domain.Workflow
```
jgaleotti: in a future PR, rename Workflow to ArazzoWorkflow

jgaleotti: add as a //TODO to rename it to ArazzoWorkflow

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -90,6 +93,11 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
     lateinit var skippedEndpoints : List<Endpoint>
         private set
 
+    var workflowsArazzo = mutableListOf<Workflow>()
```
jgaleotti: rename field to arazzoWorkflows

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -90,6 +93,11 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
...
+        private set
+
+    lateinit var workflowsArazzoById: Map<String, Workflow>
```
jgaleotti: rename field to to arazzoWorkflowsById

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -491,4 +505,12 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
             }
     }
 
+    /**
+     * Parse Arazzo and looking for Workflows
+     */
+    private fun readWorkflowsArazzo(): List<Workflow> {
```
jgaleotti: rename to readArazzoWorkflows()

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -491,4 +505,12 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
...
+        val arazzoText = ArazzoAccess.readFromDisk(config.arazzoLocation)
```
jgaleotti: add a Objects.requireNonNull() to config.arazzoLocation

jgaleotti: also, config.arazzoLocation should be a parameter in this method

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -3352,6 +3365,10 @@ class EMConfig {
     var inferFormatFromNames = false
 
 
+    @Experimental
+    @Cfg("arazzo location in disk")
+    var arazzoLocation = ""
```
jgaleotti: should this be null by default instead of the "" empty string?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -151,6 +159,13 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
             initializeDerivedParamRules(problem.derivedParams)
         }
 
+        if (config.arazzoStrategy != EMConfig.ArazzoStrategy.NONE) {
+            val arazzo = readWorkflowsArazzo()
```
jgaleotti: if EMConfig.arazzoLocation is not defined, but config.arazzoStrategy != EMConfig.ArazzoStrategy.NONE, then the configuration is invalid

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
+package org.evomaster.core.problem.rest.service.sampler
+
+import com.webfuzzing.arazzo.models.domain.Step
+import com.webfuzzing.arazzo.models.domain.Workflow
+import org.evomaster.core.problem.enterprise.SampleType
+import org.evomaster.core.problem.rest.data.RestCallAction
+import org.evomaster.core.problem.rest.data.RestIndividual
+
+class ArazzoSampler : AbstractRestSampler() {
```
jgaleotti: why workflowsArazzo, readWorkflowsArazzo and workflowsArazzoById are defined in AbstractRestSampler instead of within ArazzoSampler?

docs/options.md
```diff
@@ -275,6 +275,8 @@ There are 3 types of options:
...
 |`aiResponseClassifierWarmup`| __Int__. Number of training iterations required to update classifier parameters. For example, in the Gaussian model this affects mean and variance updates. For neural network (NN) models, the warm-up should typically be larger than 1000. *Default value*: `100`.|
 |`appendToTargetHeuristicsFile`| __Boolean__. Whether should add to an existing target heuristics file, instead of replacing it. It is only used when processFormat is TARGET_HEURISTIC. *Default value*: `false`.|
+|`arazzoLocation`| __String__. arazzo location in disk. *Default value*: `""`.|
```
jgaleotti: I am not usre that "" is a good default value (since it is not a valid location). It should be null instead.

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -579,6 +580,8 @@ class Main {
                 EMConfig.ProblemType.REST -> {
                     if (config.blackBox) {
                         BlackBoxRestModule(config.bbExperiments)
+                    } else if (config.isEnabledArazzoStrategy()) {
```
jgaleotti: but it could be blackbox and the arazzoStrategy be enabled, right?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,63 @@
+package org.evomaster.core.problem.rest.service.sampler
+
+import com.webfuzzing.arazzo.models.domain.Step
```
jgaleotti, line 5: add a //TODO to rename class Step to ArazzoStep in a future PR

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
...
+            }
+            .toMutableList()
+
+        return createIndividual(SampleType.RANDOM, actions)
+    }
+
+    /**
+     * The Steps can reference a sub-workflow or OpenApi
+     */
+    private fun resolveStep(step: Step): List<RestCallAction> {
```
jgaleotti: this method name is not declarative, since it builds a List<RestCallAction> from an ArazzoStep, it should be renamed to buildArazzoRestCallActions(ArazzoStep)

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
...
+        if (!step.operationId.isNullOrBlank()) {
+            return listOfNotNull(findActionForOperation(step.operationId))
+        }
+        if (!step.workflowId.isNullOrBlank()) {
+            val nested = workflowsArazzoById[step.workflowId] ?: return emptyList()
```
jgaleotti: but it is expected for the workflowId to be missing? Shouldn't this signal an exception?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
...
+            return nested.steps.flatMap { resolveStep(it) }
+        }
+        return emptyList()
+    }
+
+    /**
+     * Every Step Arazzo has its corresponding RestCallAction in the actionCluster
+     */
+    private fun findActionForOperation(operationId: String): RestCallAction? {
```
jgaleotti: can operationId be null? If not, please add a Objects.requireNonNull()

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
...
+     */
+    private fun resolveStep(step: Step): List<RestCallAction> {
+        if (!step.operationId.isNullOrBlank()) {
+            return listOfNotNull(findActionForOperation(step.operationId))
+        }
+        if (!step.workflowId.isNullOrBlank()) {
+            val nested = workflowsArazzoById[step.workflowId] ?: return emptyList()
+            return nested.steps.flatMap { resolveStep(it) }
+        }
+        return emptyList()
```
jgaleotti: I am not sure about this recursive approach, could you change it to a iterative approach to build the RestCallActions?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Every Step Arazzo has its corresponding RestCallAction in the actionCluster
+     */
+    private fun findActionForOperation(operationId: String): RestCallAction? {
+        val template = actionCluster.values
+            .filterIsInstance<RestCallAction>()
+            .find { it.operationId == operationId }
+        return template?.copy() as? RestCallAction
```
jgaleotti: what happen if no operationId is found? Is it expected?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2251,6 +2251,19 @@ class EMConfig {
             "Note that resource-based sampling is only applicable for REST problem with MIO algorithm.")
     var resourceSampleStrategy = ResourceSamplingStrategy.ConArchive
 
+
+    enum class ArazzoStrategy {
+        NONE,
+        ENABLED
```
jgaleotti, line 2297: ENABLED is not declarative enough. Please replace with a more declarative name and add a Javadoc to describe this strategy

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/mutator/ArazzoStructureMutator.kt
daniellopera15: We can discuss it. However, the MIO algorithm needs to use the mutator. We could use an existing mutator, but that might break the workflow logic.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: readWorkflowsArazzo: This was implemented in `AbstractRestSampler` to follow the OpenAPI pattern. Additionally, the parsed schema is being used, though it could be moved inside the Sampler itself; however, I didn't see that logic in the other Samplers.

workflowsArazzo and workflowsArazzoById: It can be moved to ArazzoSampler. It had been left in the Abstract due to a previous test, but there is no longer a need to keep it there.

core/src/main/kotlin/org/evomaster/core/Main.kt
daniellopera15: Yes, it is possible.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: Yes, it is better for it to throw an exception. However, the ideal approach would be for the ArazzoParser to perform the validations and guarantee a valid Arazzo document.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: I am adding an exception.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: I am adding an exception.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: How about calling it WORKFLOW?

core/src/main/kotlin/org/evomaster/core/Main.kt
jgaleotti: But then, arazzo only works for white-box?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/module/ArazzoRestModule.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: I added a validation in readArazzoWorkflows.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: I'm moving it to the ArazzoSampler.

docs/options.md
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/Main.kt
daniellopera15: I changed the location of the if statement.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/ArazzoSampler.kt
daniellopera15: Done. Refactored to an iterative solution using ArrayDeque.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -49,8 +49,6 @@ import org.evomaster.core.search.warning.WarningCategory
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import javax.annotation.PostConstruct
-import kotlin.sequences.forEach
-
```
jgaleotti, line 55: Please revert changes to this file if they are not needed (this seems refactoring such as formatting and unused imports)

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
jgaleotti: ok

jgaleotti (changes requested):

core/pom.xml
```diff
@@ -362,6 +362,12 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Arazzo Parser from core-extra module -->
+        <dependency>
+            <groupId>org.evomaster</groupId>
+            <artifactId>arazzo-parser</artifactId>
+            <scope>compile</scope>
```
jgaleotti: why the compiler scope? The parser will be required in the distribution to be able to parse Arazzo specs within the evomaster.jar, right?

core/src/main/kotlin/org/evomaster/core/Main.kt
```diff
@@ -579,7 +580,9 @@ class Main {
 
             val problemModule = when (config.problemType) {
                 EMConfig.ProblemType.REST -> {
-                    if (config.blackBox) {
+                    if (config.isEnabledArazzoStrategy()) {
+                        ArazzoRestModule()
```
jgaleotti: I am not sure that an ArazzoRestModule is required. I think Arazzo sampling should be possible to enable in BlackBoxRestModule, ResourceRestModule and RestModule. The Arazzo sampling looks more like a sampling strategy.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/module/ArazzoRestModule.kt
jgaleotti: After a more careful consideration, this class should be deleted.

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/Main.kt
daniellopera15: That’s a possibility. I’m going to look into making a commit that includes the Sampler but not the Module. That way, we’d avoid the Mutator issue.

core/pom.xml
daniellopera15: The default scope for dependencies is `compile`. I'm going to remove it.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -176,8 +184,14 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
         log.debug("Done initializing {}", AbstractRestSampler::class.simpleName)
     }
 
+    override fun sampleAtRandom(): RestIndividual {
+        if (shouldUseArazzoSampling()) {
+            return arazzoWorkflowsService.sampleAtRandom(actionCluster, ::createIndividual)
+        }
```
jgaleotti, line 190: add else { explicitly here

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,112 @@
...
+
+class ArazzoWorkflowsService {
+
+    @Inject
+    private lateinit var randomness: Randomness
+
+    var arazzoWorkflows = mutableListOf<Workflow>()
+        private set
+
+    lateinit var arazzoWorkflowsById: Map<String, Workflow>
```
jgaleotti, line 25: every field instance of map needs to be described in comments. Please refer to  docs/for_developers.md

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,112 @@
...
+        private set
+
+    fun load(openAPI: OpenAPI, location: String?) {
```
jgaleotti, line 28: add javadoc

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: @daniellopera15  was it modified to WORKFLOW? i still see `ENABLED` here

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2288,6 +2288,27 @@ class EMConfig {
             "Note that resource-based sampling is only applicable for REST problem with MIO algorithm.")
     var resourceSampleStrategy = ResourceSamplingStrategy.ConArchive
 
+
+    /**
+     * Enum that enables and disables the generation of individuals based on Arazzo Workflows.
+     */
+    enum class ArazzoStrategy {
```
arcuri82: if it is only two option on/off, why is this an enum and not a boolean?

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -2288,6 +2288,27 @@ class EMConfig {
...
+    @Experimental
+    @Cfg("Enable workflow-based sampling from an Arazzo document. " +
+            "Only applicable for REST with MIO.")
+    var arazzoStrategy = ArazzoStrategy.NONE
+
+    fun isEnabledArazzoStrategy() = isUsingAdvancedTechniques() && arazzoStrategy != ArazzoStrategy.NONE
+
+    @Experimental
+    @Cfg("Probability of controlling the creation of Arazzo individuals.")
+    @Probability
```
arcuri82: i guess that should add `activating=false`, as controlled by `arazzoStrategy`.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -3389,6 +3410,10 @@ class EMConfig {
     var inferFormatFromNames = false
 
 
+    @Experimental
+    @Cfg("arazzo location in disk")
```
arcuri82: `Arazzo path location on disk`

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -3389,6 +3410,10 @@ class EMConfig {
...
+    var arazzoLocation: String? = null
```
arcuri82: can be `=""`, with ` ExistingPath(val canBeBlank: Boolean, val shouldBeWritable: Boolean)` having `true,false`

arcuri82: same as `overlay` option

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -153,6 +157,10 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
 
         updateDataPoolBasedOnSchema(actionCluster)
 
+        if (config.isEnabledArazzoStrategy()) {
```
arcuri82: ` && config.arazzoLocation.isNotEmpty()`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -176,8 +184,15 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
...
+    override fun sampleAtRandom(): RestIndividual {
+        return if (shouldUseArazzoSampling()) {
+            arazzoWorkflowsService.sampleAtRandom(actionCluster, ::createIndividual)
+        } else {
+            doSampleAtRandom()
+        }
+    }
 
-
+    protected abstract fun doSampleAtRandom(): RestIndividual
```
arcuri82: the name is a bit confusing when considering there is already a `sampleAtRandom()` function. maybe use something like `sampleAtRandomeOnlyBasedOnOpenAPISchema`, with a JavaDoc stating that super-class samplers might use other info as well, like from Arazzo

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
```diff
@@ -495,4 +510,12 @@ abstract class AbstractRestSampler : HttpWsSampler<RestIndividual>() {
             }
     }
 
+    private fun shouldUseArazzoSampling(): Boolean {
+        if (!config.isEnabledArazzoStrategy()) {
```
arcuri82: this should also check that file location is not empty. in future, after experiments and fixes, this option will be on by default, and Arazzo will be activated just if a file location is activated

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+import io.swagger.v3.oas.models.OpenAPI
+import org.evomaster.core.config.ConfigProblemException
+import org.evomaster.core.problem.enterprise.SampleType
+import org.evomaster.core.problem.rest.data.RestCallAction
+import org.evomaster.core.problem.rest.data.RestIndividual
+import org.evomaster.core.search.action.Action
+import org.evomaster.core.search.service.Randomness
+import java.util.ArrayDeque
+
+class ArazzoWorkflowsService {
```
arcuri82, line 22: add JavaDoc with a couple of sentences of what Arazzo is

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+     * Map containing each Arazzo workflow associated with its corresponding ID.
+     * Used to resolve nested workflow references in steps.
+     */
+    lateinit var arazzoWorkflowsById: Map<String, Workflow>
+        private set
+
+    /**
+     * Load Arazzo workflows from disk
+     */
+    fun load(openAPI: OpenAPI, location: String?) {
```
arcuri82: why is the `AbstractRestSample` that calls the `load`, instead of doing it directly in this singleton with a `@PostConstruct` ?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+        if (location.isNullOrBlank()) {
```
arcuri82: `location` shouldn't be a nullable strings. see other comments

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+            throw ConfigProblemException("arazzoLocation must not be null when Arazzo strategy is enabled")
```
arcuri82, line 50: Arazzon strategy will be on by default in future, with location empty by default

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+        val workflows = readArazzoWorkflows(openAPI, location)
+        if (workflows.isEmpty()) {
+            throw ConfigProblemException("There must be at least one Arazzo workflow.")
```
arcuri82: in this error message, add info on the `location`

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+        arazzoWorkflows.clear()
+        arazzoWorkflows.addAll(workflows)
+        arazzoWorkflowsById = workflows.associateBy { it.workflowId }
+    }
+
+    private fun readArazzoWorkflows(openAPI: OpenAPI, location: String): List<Workflow> {
+        val arazzoText = ArazzoAccess.readFromDisk(location)
+        return ArazzoParser.parse(arazzoText, openAPI).workflows
```
arcuri82: maybe the `parse` should be in a try/catch, throwing a `ConfigProblemException` if any error? eg, what happens if user specify wrong non-Arazzo file by mistake?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Choose a random workflow
+     */
+    fun sampleAtRandom(
+        actionCluster: Map<String, Action>,
```
arcuri82: why is this given as input, and not read directly from an injected field holding ref to the singleton sampler?

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Create workflows individuals
+     */
+    fun buildIndividualFromWorkflow(
+        workflow: Workflow,
+        actionCluster: Map<String, Action>,
+        createIndividual: (SampleType, MutableList<RestCallAction>) -> RestIndividual,
+    ): RestIndividual {
+        val actions = buildArazzoRestCallActions(workflow.steps, actionCluster)
```
arcuri82: should clarify here in a comment that, for the time being, we are only considering one single OpenAPI schema fro the workflow, and ignoring cases of arazzo spanning different APIs

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
arcuri82: or are there already plans to add more in future? in that case, leave a TODO comment

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: No. At the moment, there are no plans to add more in the future. We're going to switch it to a boolean.

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/sampler/AbstractRestSampler.kt
daniellopera15: ok

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: ok

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: ArazzoWorkflowsService requires the OpenAPI to have already been processed—a step that takes place during initialization. We could use @PostConstruct, but I am not sure if the order in which beans are constructed is guaranteed.

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Are you talking about doing something like this?

@Inject
private lateinit var sampler: AbstractRestSampler

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
arcuri82: good point. didn't think about it. currently there is no way to enforce any ordering in @PostConstruct execution. can keep it like that. but add a comment to clarify this issue

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
arcuri82: yes. by the time sampleAtRandom is called, the singletons should all be already initialized

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

core/src/main/kotlin/org/evomaster/core/problem/rest/service/ArazzoWorkflowsService.kt
daniellopera15: Done

arcuri82 (approved):

PR #1690, opened 2026-08-20
Fix the dialect transform for PostgreSQL enum CHECK constraints

agusaldasoro:
Part of the work in #1688, split into one change per defect.

## The defect

`JSqlConditionParser` rewrites a few dialect constructs before handing the expression to JSQLParser. One of those rules strips the `ARRAY[...]` wrapper using a greedy group:

```java
transformedStr.replaceAll("ARRAY\s*\[([^<]*)\]", "$1")
```

`[^<]*` is greedy, so the match runs to the **last** `]` in the whole string. In the shape PostgreSQL emits for an enum column, that bracket does not belong to the array — it belongs to the cast wrapping it:

```
((c)::text = ANY ((ARRAY['A'::character varying])::text[]))
```

So the substitution consumes the wrong bracket and produces `...])::text[))`, which is not valid SQL. JSQLParser explores alternatives over that for a very long time and then throws, and `appendTableConstraints` catches the exception and continues without a constraint. **The entire cost bought nothing.**

## Measured

Taken from one real schema of 59 tables, recovered through the driver's `infoSUT` endpoint. Each of its 48 CHECK constraints parsed individually:

| | Before | After |
|---|---|---|
| constraints that parse | 4 of 48 | **48 of 48** |
| total parse cost | 317.6 s | **72 ms** |
| the single 24 KB constraint | 275.0 s | — |

End to end on that system, SMT-LIB generation for one query went from 331,695 ms to 313 ms, and the number of queries reaching the solver in a three-minute run went from 1 to 276.

Speed is the smaller half. These constraints are now **applied** rather than silently discarded: generated rows respect the enums the schema declares, where before they were free to violate them without any counter recording it.

## The change

Two parts. The complete shape — array, cast and `ANY` comparison — is recognised ahead of the general rules and rewritten to an `IN`. And the general ARRAY rule now excludes `]` from its group, so the match ends at the array's own closing bracket instead of the last one anywhere in the string.

`ARRAY` therefore appears in two substitutions now. Both are reachable, because they cover disjoint shapes: the new rule only matches when the array is wrapped in a cast, and PostgreSQL also emits `= ANY (ARRAY[...])` without one, while a bare `ARRAY[...]` can appear on its own.

| Input | Full-shape rule | General ARRAY rule |
|---|---|---|
| `= ANY ((ARRAY[...])::text[])` | handles it | no-op, the `ARRAY[` is already gone |
| `= ANY (ARRAY['A','B'])` | no match | **required** |
| `x = ARRAY[1,2]` | no match | **required** |

## A second case the greedy group broke

Tracing the rules over those shapes turned up a defect independent of the enum one. With two arrays in a single expression, the match ran from the first `ARRAY[` to the last `]` anywhere in the string and swallowed everything between them:

```
input:    (a = ARRAY[1,2] AND b = ARRAY[3,4])
before:   (a = 1,2] AND b = ARRAY[3,4)     <- corrupted
after:    (a = 1,2 AND b = 3,4)
```

This change fixes that too.

## A note on the previous diagnosis

This was a known symptom, attributed to the parser library. `SqlConditionParserTest` carried a test asserting that such a constraint throws, and a disabled one pointing at an upstream JSQLParser issue. The cause was on our side, in the transform: given well-formed input the library parses these in milliseconds.

That test now asserts a successful parse and keeps its `@Timeout` as the regression guard — a transform broken this way does not fail visibly, it just gets slow and quietly stops constraining anything.

## Tests

`CheckConstraintParsingCostTest` covers the enum shape at realistic widths and bounds the cost.

`SqlConditionParserTest` gains two cases for the shapes above. `testTwoArraysInOneExpression` is a real regression guard: it fails on the previous regex, throwing `Encountered unexpected token: "IN"`, and passes on this one. `testArrayWithoutCastIsStillTransformed` passes on both by design — it guards that the general rule stays reachable, so a later cleanup does not delete it as dead code.

Verified with `mvn clean verify -DskipTests --fae`; `dbconstraint` is fully green.


jgaleotti (approved):

core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlConditionParser.java
```diff
@@ -81,19 +81,39 @@ public SqlCondition parse(String sqlConditionStr, ConstraintDatabaseType databas
...
+                " IN ($1)");
+
         /*
          * The JSQL parser does not properly parse the Postgresql SQL dialect function "ANY"
-         * We can work aroung this limitation by replacing the "= ANY (...)" with a valid " IN (...)"
+         * We can work around this limitation by replacing the "= ANY (...)" with a valid " IN (...)"
          * string
          */
-        String transformedStr = originalSqlStr.replaceAll("=\\s*ANY\\s*\\(([^<]*)\\)", " IN ($1)");
+        transformedStr = transformedStr.replaceAll("=\\s*ANY\\s*\\(([^<]*)\\)", " IN ($1)");
```
arcuri82, line 125: are these cases due to limitation of parser? if so, should add explicit `TODO` comments, stating this might not longer be needed in future when parser is updated.

arcuri82 (changes requested):

core-extra/dbconstraint/src/main/java/org/evomaster/dbconstraint/parser/jsql/JSqlConditionParser.java
agusaldasoro: Fixed the comments

arcuri82 (approved):

PR #1691, opened 2026-08-20
Make what the solver costs, and why it fails, observable

agusaldasoro:
Part of the work in #1688, split into one change per defect. This one adds no behaviour: it only makes existing behaviour measurable. It is the change the other diagnoses were built on, so it is probably the easiest of the set to review.

Three gaps made the last round of experiments hard to read.

## Failure attribution

One counter grouped SQL parse failures together with SMT-LIB generation failures. The two live in different components and are corrected in different places, so the aggregate could not direct a fix — it said that half the invocations failed, but not at what.

Split into `sqlZ3SqlParseFailures` and `sqlZ3SmtlibGenFailures`, whose sum is the counter that already existed. It was this split that attributed one SUT's failures to their real cause: every one of them came from SMT-LIB generation, none from the SQL parser, which pointed straight at the query shape responsible.

## Solve time

The two timers already in place bracket the Z3 call and formula generation. Between them they leave out writing the `.smt2` file, rebuilding the gene tree from a solution — which also happens on **every cache hit** — and the call overhead itself. Measuring only those two understates what the solver costs the search.

`sqlZ3SolveTotalMs` times the whole `solve()` call, in a `finally` so every exit path is counted. It deliberately overlaps the other two rather than replacing them: the difference between them is exactly the part that was previously unaccounted for.

## Insertion time

Rows a solver call generates are inserted on **every later evaluation** of an individual that carries them. One successful call therefore imposes a cost that repeats for the rest of the search, and no solver-side counter can see it, because it is paid in `EnterpriseFitness` rather than inside `solve()`.

`sqlInsertionExecutionTotalMs` and `sqlInsertionExecutions` record it. Per execution as well as in total, because the average is the quantity of interest: a total alone cannot distinguish a few slow insertions from many cheap ones.

## Tests

`SqlZ3StatisticsCountersTest` pins the two relations the statistics file invites a reader to rely on — the failure breakdown summing to its aggregate, and every query seen ending up classified as either a hit or a miss, which requires a translation failure to close the accounting the same way a solver answer does.

`Z3InvocationCostTest` measures where the time in one invocation actually goes, separating the container round-trip from process creation from Z3 itself.


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -149,6 +149,32 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
...
+            only those understates what the solver costs the search.
+         */
+        if (stats == null) {
+            return doSolve(schemaDto, sqlQuery, numberOfRows, null)
+        }
+
+        val solveStart = System.currentTimeMillis()
+        try {
+            return doSolve(schemaDto, sqlQuery, numberOfRows, stats)
+        } finally {
```
arcuri82: instead of writing this time measuring manually, should reuse what already available in `TimeUtils`, like `measureTimeMillis`

core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -166,8 +192,8 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
         val queryStatement = try {
             parseStatement(sqlQuery)
         } catch (e: RuntimeException) {
-            LoggingUtil.getInfoLogger().warn("SQL-Z3: failed to parse SQL query as SMT-LIB: '$sqlQuery'")
-            stats?.reportSqlZ3ParseFailure()
+            LoggingUtil.getInfoLogger().warn("SQL-Z3: failed to parse SQL query: '$sqlQuery'")
```
arcuri82: can this query happen often? in that case, could use `LoggingUtil.uniqueWarn`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1692, opened 2026-08-20
Let each SQL data strategy evaluate its own precondition

agusaldasoro:
Part of the work in #1688, split into one change per defect.

**This is a latent defect.** No case was observed where it actually suppressed an invocation. It is corrected because it is a real inconsistency and the correction is small — not because any observed experimental result is attributed to it. Flagging that up front so it is reviewed on its own merits.

## The inconsistency

Before deciding whether there is any work to do for a failed WHERE clause, `ApiWsStructureMutator` builds the failed-WHERE map grouped per table and filtered by what the sampler can insert into. If that map comes out empty, it returns.

That is the correct precondition for the **search-based** strategy: it builds `INSERT` statements table by table, so with no table it genuinely has nothing to do.

The **constraint solver** does not consume that map at all. It works from the query text and the schema. Sharing one entry point made it inherit a precondition that is not its own.

The consequence, if it were hit: a failed WHERE over something the filter rejects — a view, or a table whose identifier does not match the schema exactly — would leave the solver unreachable even though there is a query recorded and translatable. And it would do so silently, since no counter is incremented, making the case indistinguishable in the statistics from a system that simply issues no failing queries.

## The change

Each strategy now evaluates its own precondition. The precedence mirrors what `handleFailedWhereSQL` already does — the search-based strategy wins when both are enabled — so the solver's precondition only applies when the solver is the one that will actually run.

The predicate is extracted as a pure function so it can be tested without constructing an individual, a sampler and a fitness value.

## Tests

`FailedWhereStrategyGateTest` covers each strategy in isolation, the precedence when both are enabled, and that neither runs without input of its own kind.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
@@ -57,8 +57,10 @@ abstract class ApiWsStructureMutator : StructureMutator() {
          * inconsistency is fixed because it is one, not because it was shown to have suppressed
          * anything.
          *
-         * The precedence mirrors [handleFailedWhereSQL]: the search-based strategy wins when both are
-         * enabled, so the solver's own precondition only applies when it is the one that will run.
+         * The two are mutually exclusive — [EMConfig] rejects a configuration with both enabled — so
+         * the order below is not a precedence rule but a total definition: with neither strategy
+         * enabled there is nothing to do whatever the inputs say, which keeps the predicate honest
+         * independently of the `shouldGenerateSqlData` check that currently guards the call site.
          */
         internal fun nothingToDoForFailedWhere(
             noInsertableTables: Boolean,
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/org/evomaster/core/problem/api/service/ApiWsStructureMutator.kt
```diff
@@ -42,6 +42,36 @@ abstract class ApiWsStructureMutator : StructureMutator() {
...
+         * enabled there is nothing to do whatever the inputs say, which keeps the predicate honest
+         * independently of the `shouldGenerateSqlData` check that currently guards the call site.
+         */
+        internal fun nothingToDoForFailedWhere(
+            noInsertableTables: Boolean,
+            noFailedWhereQueries: Boolean,
+            generateSqlDataWithSearch: Boolean,
+            generateSqlDataWithZ3: Boolean
+        ): Boolean {
+            if (generateSqlDataWithSearch) return noInsertableTables
```
jgaleotti, line 71: can generateSqlDataWithSearch and generateSqlDataWithZ3 be enabled simultaneously?

jgaleotti: EMConfig currently rejects that configuration (EMConfig.kt:761-763)

jgaleotti (approved):

arcuri82 (approved):

PR #1694, opened 2026-08-20
Stop a derived table from discarding the whole query

agusaldasoro:
## The defect

`SmtLibGenerator` assumed every item in a `FROM` clause was a schema table and cast it without checking. A parenthesised sub-select, which can be generated by an ORM for inheritance with `UNION ALL`, caused the cast to fail.

The exception was caught too far upstream, so the **entire query** was discarded instead of just the unsupported item. This meant that queries joining a real table with a derived table could lose perfectly valid constraints as well.

## The fix

Derived tables are now identified using the aliases declared by the query's `FROM` and `JOIN` clauses. If a column cannot be mapped to a supported table, only that condition is dropped and the loss is counted.

We also translate top-level `AND` conditions independently, so one unsupported condition no longer causes all the others to be dropped. We don't split `OR`, since dropping part of a disjunction could make the resulting formula stronger than the original query.

The generated SMT-LIB changes from:

```text
(assert (and (> (SCORE account__1) 30) (= (NAME account__1) "a")))
```

to:

```text
(assert (> (SCORE account__1) 30))
(assert (= (NAME account__1) "a"))
```

These are equivalent in SMT-LIB since top-level assertions are conjunctive.

## What this does not do

This does **not** add support for derived tables. Unsupported conditions are still dropped; the difference is that the decision is now correct, local to the condition, and counted.

## Tests

Added coverage for:

* derived tables using the same alias as a real table
* unsupported conditions within `AND`
* `OR` conditions
* nested `AND` inside `OR`

Tests were checked against deliberately disabled fixes to avoid vacuous coverage.

Also added a scaling test to verify that SMT-LIB generation remains linear with schema size.


core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTConditionVisitor.kt
```diff
@@ -111,6 +111,27 @@ class SMTConditionVisitor(
...
+                    the schema-level constraints of the query intact.
+
+                    Only qualified columns are checked. An unqualified one resolves to the default
+                    table, which is a schema table by construction, and the name it arrives under is
+                    the ASCII-folded one — comparing that against the schema's own spelling would
+                    reject a valid table whose name is not ASCII, quietly dropping its conditions.
+                 */
+                if (sqlCondition.tableName != null &&
+                    tables.none { it.id.name.equals(tableName, ignoreCase = true) }) {
+                    throw RuntimeException(
```
jgaleotti: This means derived tables are not supported? Have you checked TableColumnResolver? That class can handle derived tables in case you need support for that.

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTConditionVisitor.kt
agusaldasoro: Good point, thanks! It turned out to be a real bug, not just a cleaner way to do the check.

The issue was that we assumed a column was derived whenever its qualifier wasn't a schema table. That breaks when a sub-select uses the same alias as a real table: we generate a constraint for the wrong table, Z3 rejects it, and we silently lose the condition. This PR fixes that by using the aliases defined in the query itself, without changing dependencies.

I also added a proof of concept in [#1698](https://github.com/WebFuzzing/EvoMaster/pull/1698) using `TableColumnResolver`, in case we decide the dependency change is worth it later. It handles a few extra cases like CTEs and correlated subqueries.

And yes, derived tables are still not supported. The condition is still dropped, but now we detect and count it correctly.

I also applied Copilot's suggestion to translate `AND` conditions independently, so one unsupported condition doesn't make us drop the others. We don't split `OR`, since dropping an `OR` condition could make a satisfiable query unsatisfiable. There are tests for both cases.

jgaleotti (approved):

arcuri82 (approved):

PR #1696, opened 2026-08-20
Memoise what does not change, and make the cache bound configurable

agusaldasoro:
## CHECK expressions

A generator is built afresh on every solver cache miss, and each one re-parsed **every** CHECK constraint in the schema. The schema does not change during a run, so one parse per expression is enough.

The memo is owned by the solver and passed into the generator rather than living in a static field, following the convention used elsewhere in the subsystem. Failures are memoised as null: an expression that cannot be parsed will not become parseable later in the run, and the warning moves to the parse site so a remembered failure stops repeating it.

## Untranslatable queries

The failure branches returned before touching any cache, so a query that could not be translated was retried on every encounter.

This is deliberately narrower than caching everything. Z3's `UNKNOWN` and `ERROR` may be a timeout or a transient container fault, so they stay uncached. A *translation* failure is deterministic: neither the query nor the schema changes between encounters.

Failures are still counted on every encounter rather than only the first, so the reported totals keep reflecting how often the search runs into one — the memo saves the work, not the accounting. The failure kind is remembered too, so the breakdown between SQL parsing and SMT-LIB generation survives later encounters.

## Cache size

The bound was a private constant of 500. It becomes `--sqlZ3CacheSize`, defaulting to 5000.

On the benchmark, systems issuing more distinct queries than the old bound were turning a large share of their cache misses into re-solves of evicted entries rather than genuinely new work — one issued 2,153 distinct queries in a one-hour search. Entries are small, a query string and a solver result, so the higher bound costs little.

## Tests

`CheckExpressionMemoisationTest` measures the saving over several generations, checks that a warm memo does not change the generated formula, and that failures are memoised too. It uses many ordinary constraints rather than one expensive shape, so the measurement does not depend on which dialect quirks the parser currently handles.

`UntranslatableQueryCacheTest` covers the memo, the kinds it distinguishes, that counting continues on repeat encounters, and that the bound evicts.

Verified with `mvn clean verify -DskipTests --fae`; `docs/options.md` is regenerated, so `ConfigToMarkdownTest` passes.


core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -77,9 +78,20 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
     // corrupts the map. This makes the cache safe for the parallelized fitness evaluation that is planned.
     private var z3ResultCache: MutableMap<Pair<String, Int>, Z3Result>? = null
 
-    companion object {
-        private const val MAX_CACHE_SIZE = 500
+    /**
+     * Queries that could not be translated at all, remembered so they are attempted once per run.
+     *
+     * Unlike Z3's UNKNOWN and ERROR outcomes — which may be a timeout or a transient container fault
+     * and are deliberately left uncached — a translation failure is deterministic: neither the query
```
jgaleotti: left uncatched?

agusaldasoro: modified

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/database/sql/solver/SMTLibZ3DbConstraintSolver.kt
```diff
@@ -119,14 +138,43 @@ class SMTLibZ3DbConstraintSolver() : DbConstraintSolver {
     private fun postConstruct() {
```
arcuri82, line 138: not in this PR, create a new one, but is this a service that is injected? if yes, it should be under a `service` package, recall what discussed in `for_developers.md`: 

>All injectable services should be singletons, and declared under a package called `*.service` (this
is to make it easy to find out which services are available).

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
```diff
@@ -65,6 +65,17 @@ class EMConfig {
...
+        /**
+         * Default number of entries kept in the Z3 solver's memoization cache. See [sqlZ3CacheSize].
+         *
+         * Chosen from measurements: one system under test issued 2,153 distinct queries in a one-hour
+         * search, and with the previous bound of 500 nearly half of its cache misses were re-solves of
+         * entries that had been evicted rather than genuinely new queries. Entries are small — a query
+         * string and a solver result — so a higher bound costs little memory and removes that whole
+         * class of wasted work.
+         */
+        const val DEFAULT_SQL_Z3_CACHE_SIZE = 5000
```
arcuri82: why declaring it as a `val` instead of using it directly `var sqlZ3CacheSize = 5000`? using constants is needed just for when we need to check if a value has been modified by the user

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/EMConfig.kt
agusaldasoro: Done!

arcuri82 (approved):

PR #1697, opened 2026-08-20
Canonize Cypher patterns by expanding quantified paths and variable-length edges

andyfelder16:
Second of the PRs splitting #1641, on top of #1647 which is now merged.

- Rewrites a MATCH pattern into an equivalent one with no quantified path patterns and no variable-length relationships, so the calculator only ever deals with plain nodes and edges
- `Neo4jPatternExpander` does the expansion, always to the construct's lower bound, since satisfying that is enough for the query to match. `((a)-[]->(b)){1,3}` becomes one copy of the sub-pattern spliced onto the outer nodes, and `[:KNOWS*2]` becomes a chain with a fresh intermediate node
- `CypherConditionRenamer` rebinds the conditions of a cloned edge to the fresh variable names, so labels and properties follow their element
- A zero lower bound merges the two endpoints, since a zero-length path identifies them
- The only change to `Neo4jHeuristicsCalculator` is the line that calls the expander
- 19 tests


client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/ConditionRenamer.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Deep-clones a {@link CypherCondition} (or {@link Operand}) while substituting the variable names it
+ * refers to, according to a rename map. Used by {@link Neo4jPatternExpander} when it clones a
+ * variable-length edge into a chain: the edge's type/property conditions and the endpoints' labels
+ * must be re-bound to the fresh variable names of the cloned edges and intermediate nodes.
+ * <p>
+ * Only variable names change; operators, literals and structure are preserved. A name not present in
+ * the map is left unchanged.
+ */
+final class ConditionRenamer {
```
jgaleotti: rename class to CypherConditionRenamer

jgaleotti (changes requested):

jgaleotti (approved):

arcuri82 (approved):

PR #1704, opened 2026-08-24
Java regex boundary assertions

lmasroca:
This PR adds `MULTILINE` flag and some boundary assertions to Java regex, including:
- `\A` which matches at the start of the string.
- `\z` which matches at the end of the string.
- `^` which matches at the start of the string (like `\A`), unless in `MULTILINE` mode (`m` flag) in which case it also matches after any line terminator.*
- `$` which matches at the end of the string (like `\z`), unless in `MULTILINE` mode (`m` flag) in which case it also matches just before a line terminator.*

\* These already existed previously on Java regex, see more information below for changes beyond `MULTILINE` mode.

This PR also replaces (for Java regex) the previous hack used for `^$`, allowing them to appear at any point on a Java regex by using `AssertionRxGene` and related repairs to handle them.
Previously, all regex types used `DisjunctionRxGene`'s `matchStart`/`matchEnd` booleans to show that a disjunction matches the start/end of the string and if so avoid randomly appending pre/postfixes to the disjunction's value. Now JVM regexes always set those two values to `true` so that nothing is appended to disjunctions, the actual appends now happen on `RegexGene` for Java regex (non-JVM regex types still keep current behavior). This also allows remaining results of repairs to escape outwards of the regex into the pre/postfix. For example, `(?=abc)a` can now produce `abc` by forcing the remaining requirements of lookahead repair (in this case `bc`) onto the postfix of the `RegexGene`.

Note: some changes were made on the visitor, as previously `assertions` resulted in no genes and using VisitResult.data to move `String`s and `Pair` of `Boolean`s throughout the visit, to then use to set the mentioned booleans on the top level `DisjunctionRxGene`. This is no longer necessary as now `assertions` in Java always generate a gene, and those disjunction booleans always remain as `true`.


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
```diff
@@ -66,6 +67,9 @@ class RegexGene(
             null
         }
 
+    private var javaPrefix : String = ""
+    private var javaPostfix : String = ""
+
```
arcuri82, line 81: what are those fields for? add JavaDocs. also fields shoukdl not come after already started with method declarations.

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
```diff
@@ -66,6 +67,9 @@ class RegexGene(
...
     override fun copyContent(): Gene {
         return RegexGene(name, disjunctions.copy() as DisjunctionListRxGene, sourceRegex, regexType, fixedValue, usingFixedValue, externalRegexFlags, hasAssertions)
     }
```
arcuri82, line 87: shouldn't copy handle `javaPrefix`  and `javaPostfix`? and in such case, curios why no test case failed due to it

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
lmasroca: Adding JavaDocs. Regarding fields after method declarations i do not see methods declarations before it (`requiresAssertionHandling` and `pattern` are both fields) unless you are referring to `compiledPattern(...)` which is a method defined on the companion object before it. If that is it should i move the companion object after field declarations?

core/src/main/kotlin/org/evomaster/core/search/gene/regex/RegexGene.kt
lmasroca: Yes, adding that too.

Also related, since the `setPrefixAndPostfix(...)` method only got called when a JVM regex `hasAssertions` was true (which is set during Java pattern visit when an `AssertionRxGene` is built) JVM regexes with no assertions had no pre/postfix set to them. Same thing happens for `RegexGene`s not built from `GeneRegexJavaVisitor`. So JVM regex defaults to full match (both pre/postfixes are "", as we never set them) unless built with Java visitor or `createGeneForJVM` and containing at least one assertion gene. This may be related to why tests did not fail.

We can remove `hasAssertion` and its check so that `DisjunctionRxGene.attemptAssertionRepair(randomness)` is always called, since if there are no assertions it will return `AssertionRepairResult.SUCCESS` which allows pre/postfixes to be appended to JVM regex again.

The possible issue with this is `GeneSamplerForTests.sampleRegexGene(rand)`, since that creates random structures there is a chance this creates an unsatisfiable gene tree by using assertion genes. For this we could hardcode `GeneSamplerForTests.sampleAssertionRxGene` to always have a lookahead of `""` ( equivalent to `(?=)`), which is always valid and repairs are successful, instead of sampling a random `innerGene` for lookahead.

We may also want to change `sampleRegexGene`'s `sourceRegex` to `(?s).*` instead of `.*`, as otherwise line terminators are not accepted by `.`. This enables the `DOT_ALL` flag, which makes `.` match all chars including line terminators.

lmasroca: Just tried implementing these changes and ran `GeneRandomizedTest` locally, one of those tests failed because the random structure generated an unmounted `SqlPrimaryKeyGene`. This failed during `GeneRandomizedTest.verifyCopyValueFrom(...)`, as `x` was globally valid (the SQL gene was not on the active gene of the `ChoiceGene` wrapping it) but `y` was not (SQL gene was now on the active gene), resulting in `x.unsafeCopyValueFrom(y)` returning false.

Since we are using `unsafeCopyValueFrom(other)` for which "there is no check if constraints are kept satisfied" I suppose not checking here for `y.isGloballyValid()` before calling `unsafeCopyValueFrom` is fine. However since `ChoiceGene.unsafeCopyValueFrom(other)` internally calls `Gene.copyValueFrom(gene)` a validity check happens here.

One option to avoid this issue could be adding a check for `y.isGloballyValid()` before calling `unsafeCopyValueFrom` so that this does not happen here.

arcuri82 (approved):

arcuri82 (approved):

PR #1706, opened 2026-08-24
AsyncAPI 3.x: parse channels and operations

LautaroPetaccio:
Second in the AsyncAPI stack, on top of #1 (which sits on WebFuzzing/EvoMaster#1681).

#1681 read the message layer: what a service can be sent, and the JSON Schema of it. This adds where those messages travel and which of them a given interaction uses.

## What it adds

- **`AsyncApiChannel`** an addressable place on the broker (a Kafka topic, an AMQP queue, an MQTT topic, a WebSocket endpoint) plus the messages it carries. Its `address` is nullable on purpose: the specification allows an explicit `address: null` to say it is only known at run time, and published documents do use that.
- **`AsyncApiOperation`** and **`AsyncApiReply`**. The reply is the construct AsyncAPI 3.0 added that makes an asynchronous interaction observable without instrumentation.
- **`AsyncApiDocument`** gains `channels`, `operations`, and accessors for resolving an operation to its channel and messages.

## Three things that are less obvious than they look

**Messages can be written inline in a channel.** Not every document uses `components.messages`; plenty declare the message right where it is used. Those are promoted into the one message map under a synthetic `<channel>.<localKey>` id, so nothing downstream has to care where a message was declared.

**A `$ref` may address a message through its channel** `#/channels/<c>/messages/<k>` and that channel-local key is frequently *not* the component id. A channel may expose `#/components/messages/errorMessage` as just `error`. Hence the `messageKeys` indirection rather than a plain list.

**A narrowed selection is never widened again.** An operation's `messages` array picks a subset of what the channel offers. When an entry cannot be resolved, typically because that message itself had to be dropped, for instance for being written in Avro, the operation is left with fewer messages and does *not* fall back to the whole channel. Widening a selection the document deliberately narrowed would have the operation drive messages it never claimed to. There is a test pinning this, because the obvious `ifEmpty { channel.messageIds }` is exactly the wrong thing and I wrote it that way first.

## Degrading rather than failing

The same principle as #1681: an operation with no valid `action`, or naming a channel that does not exist, is dropped with a warning rather than taking the document down. `broken-parts.yaml` is a fixture with five operations of which two are well-formed, and the test asserts that exactly those two survive and that each of the other three is explained.

`validate()` arrives here rather than in #1681, since operations are the thing there has to be at least one of.

## Testing

`AsyncApiParserTest` grows to 35 tests (49 across the package).
---

*Second in the AsyncAPI stack, on top of #1681 which is now merged. It also carries two one-line
comments that landed on the message-parsing branch just after #1681 was merged, saying what
`AsyncApiDocument.warnings` and `AsyncApiMessage.examples` hold.*

*The rest of the stack follows: external references, then servers and bindings, then the
real-document suite, then the EvoMaster-side gene and action building.*

### DIFF
--- a/core-extra/asyncapi-parser/src/test/resources/asyncapi/artificial/broken-parts.yaml
+++ b/core-extra/asyncapi-parser/src/test/resources/asyncapi/artificial/broken-parts.yaml
@@ -1,0 +1,52 @@
+asyncapi: 3.0.0
+info:
+  title: A document with several unusable parts
+  version: 1.0.0
+
+channels:
+  good:
+    address: good.request
+    messages:
+      request:
+        $ref: '#/components/messages/request'
+      dangling:
+        $ref: '#/components/messages/doesNotExist'
+
+operations:
+  # this one is fine, and must survive everything below
+  works:
+    action: receive
+    channel:
+      $ref: '#/channels/good'
+
+  noAction:
+    channel:
+      $ref: '#/channels/good'
+
+  wrongAction:
+    action: publish
+    channel:
+      $ref: '#/channels/good'
+
+  missingChannel:
+    action: receive
+    channel:
+      $ref: '#/channels/notThere'
+
+  badCorrelationTarget:
+    action: send
+    channel:
+      $ref: '#/channels/good'
+
+components:
+  messages:
+    request:
+      name: Request
+      # not one of the two runtime expressions the specification defines
+      correlationId:
+        location: 'somewhere/else'
+      payload:
+        type: object
+        properties:
+          value:
+            type: integer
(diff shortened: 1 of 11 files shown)


core-extra/asyncapi-parser/src/test/resources/asyncapi/artificial/broken-parts.yaml
```diff
@@ -0,0 +1,52 @@
+asyncapi: 3.0.0
+info:
+  title: A document with several unusable parts
```
jgaleotti, line 3: I am not sure if project-wide it is more used synthetic than artificial?

jgaleotti (approved):

core-extra/asyncapi-parser/src/test/resources/asyncapi/artificial/broken-parts.yaml
LautaroPetaccio: I have followed the `swagger/artificial` + `swagger/sut` and `graphql/artificial` that were done in the repo, but I'd gladly change it if you find it better.

arcuri82 (approved):

PR #1707, opened 2026-08-24
AsyncAPI 3.x: follow references into other documents

LautaroPetaccio:
Fourth in the AsyncAPI stack, on top of #2.

Documents are routinely split across files — a message payload points at a schema that lives next door. Until now such a reference could not be followed, and the message was dropped.

## Inlined, not linked

Rather than keeping a second lookup path for "schemas that live elsewhere", the other document's `components/schemas` and `components/messages` are **copied into** the primary one under keys prefixed `_ext_<hash>_`, and every reference that pointed at them is rewritten to the local form.

That preserves the invariant the message layer rests on: one flat map in which any reference a payload makes can be resolved. Whoever builds genes later still only has to look in `componentSchemas`.

The hash is derived from the absolute location, so the names are stable across runs and generated output stays diffable.

## Three cases that are easy to get wrong

**A reference is resolved against the document that makes it.** An imported document referring to `shared.yaml` means the one next to *itself*, not next to the primary document. So the primary document is rewritten *before* anything is copied into it — walking it afterwards would re-resolve an imported document's relative references against the wrong directory, and if that happened to hit another loaded document it would bind them to a completely different schema with no sign anything was wrong. `nested/` is a fixture built to reproduce exactly that, and the test asserts the reference is left unresolved and reported rather than silently misbound.

**A document that names itself is not imported into itself.** Some generators write every reference as an absolute one, including those that stay inside the file. `self-reference.yaml#/components/schemas/Thing` from inside `self-reference.yaml` is just a local reference written the long way; it is turned back into one, rather than importing the whole file into itself and doubling every schema and message it declares.

**Only `components/schemas` and `components/messages` can be imported.** A pointer into some other part of another document — an OpenAPI `x-webhooks` section, say, which is a real case — is reported, and the message depending on it is dropped rather than left holding a reference that nothing downstream can follow.

## Bounded

There is a ceiling of 100 imported documents. References are paths, and a server that answers every path, or a symlink loop, would otherwise be followed forever.

### DIFF
--- a/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
+++ b/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
@@ -346,7 +346,7 @@ public class AsyncApiRefResolver {
      */
     private static String locationOf(String ref, DocumentLocation from, List<String> warnings) {
         try {
-            return RefLocations.computeLocation(ref, from, warnings);
+            return RefLocations.resolveDocumentLocation(ref, from, warnings);
         } catch (Exception e) {
             warnings.add("Cannot work out what document '" + ref + "' refers to: " + e.getMessage());
             return null;
--- a/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
+++ b/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
@@ -18,7 +18,6 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -49,8 +48,14 @@ public class AsyncApiRefResolver {
 
     private static final String MESSAGES = "messages";
 
-    public static final String SCHEMA_PREFIX = RefLocations.FRAGMENT_SEPARATOR + RefLocations.PATH_SEPARATOR
-            + COMPONENTS + RefLocations.PATH_SEPARATOR + SCHEMAS + RefLocations.PATH_SEPARATOR;
+    /**
+     * How a pointer at any component begins, e.g. the "#/components/" of
+     * "#/components/schemas/Order".
+     */
+    private static final String COMPONENT_PREFIX = RefLocations.FRAGMENT_SEPARATOR
+            + RefLocations.PATH_SEPARATOR + COMPONENTS + RefLocations.PATH_SEPARATOR;
+
+    public static final String SCHEMA_PREFIX = COMPONENT_PREFIX + SCHEMAS + RefLocations.PATH_SEPARATOR;
 
     /**
      * The two component kinds that are worth pulling out of an external document. Schemas are
@@ -170,7 +175,7 @@ public class AsyncApiRefResolver {
         }
 
         String rest = ref.substring(SCHEMA_PREFIX.length());
-        int slash = rest.indexOf('/');
+        int slash = rest.indexOf(RefLocations.PATH_SEPARATOR);
         String key = slash < 0 ? rest : rest.substring(0, slash);
 
         return key.trim().isEmpty() ? null : decodePointerSegment(key);
@@ -503,18 +508,18 @@ public class AsyncApiRefResolver {
         if (segments.length < 3 || !COMPONENTS.equals(segments[0]) || !INLINABLE.contains(segments[1])) {
             warnings.add(
                     "Reference '" + original + "' points at '" + path + "' of another document."
-                            + " Only components/" + SCHEMAS + " and components/" + MESSAGES
+                            + " Only " + COMPONENTS + RefLocations.PATH_SEPARATOR + SCHEMAS
+                            + " and " + COMPONENTS + RefLocations.PATH_SEPARATOR + MESSAGES
                             + " can be imported.");
             return null;
         }
 
-        StringBuilder renamed = new StringBuilder("#/")
-                .append(COMPONENTS).append('/')
-                .append(segments[1]).append('/')
+        StringBuilder renamed = new StringBuilder(COMPONENT_PREFIX)
+                .append(segments[1]).append(RefLocations.PATH_SEPARATOR)
                 .append(prefix).append(segments[2]);
 
         for (int i = 3; i < segments.length; i++) {
-            renamed.append('/').append(segments[i]);
+            renamed.append(RefLocations.PATH_SEPARATOR).append(segments[i]);
         }
 
         return renamed.toString();
@@ -569,9 +574,7 @@ public class AsyncApiRefResolver {
 
     private static DocumentLocationType locationTypeOf(String absoluteLocation, DocumentLocation from) {
 
-        String lower = absoluteLocation.toLowerCase(Locale.ENGLISH);
-
-        if (lower.startsWith("http:") || lower.startsWith("https:")) {
+        if (RefLocations.isHttpLocation(absoluteLocation)) {
             return DocumentLocationType.REMOTE;
         }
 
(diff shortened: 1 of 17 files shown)


core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
```diff
@@ -127,6 +305,271 @@ public static List<String> collectRefs(JsonNode node) {
...
+        }
+
+        /*
+            Some generators write even a reference that stays inside the document as an absolute
+            one. It names this very file, so it is just a local reference written the long way,
+            and turning it back into one is all that is needed.
+         */
+        if (absolute.equals(primary)) {
+            String fragment = fragmentOf(ref);
+            return fragment.trim().isEmpty() ? null : "#" + fragment;
```
jgaleotti: replace "#" with constant

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
```diff
@@ -127,6 +305,271 @@ public static List<String> collectRefs(JsonNode node) {
...
+     * the primary document's own component of that name, silently binding a payload to an
+     * unrelated schema.
+     */
+    private static String renameComponent(
+            String fragment,
+            String prefix,
+            String original,
+            List<String> warnings) {
+
+        String path = fragment.startsWith("/") ? fragment.substring(1) : fragment;
```
jgaleotti: replace "/" with constant

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
```diff
@@ -21,4 +28,72 @@ private RefLocations() {
     public static boolean isLocalRef(String ref) {
         return ref.startsWith("#");
```
jgaleotti: see comment above

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
```diff
@@ -21,4 +28,72 @@ private RefLocations() {
...
+
+    /**
+     * The absolute location of the document a reference points at, resolved against the
+     * document making the reference. Null when the reference is not one this can make sense of,
+     * in which case {@code messages} says why.
+     *
+     * @throws IllegalArgumentException if the referring document was supplied as text, as there
+     *                                  is then nothing for a relative location to be relative to
+     */
+    public static String computeLocation(String ref, DocumentLocation currentSource, List<String> messages) {
```
jgaleotti: extractLocation perhaps?

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
```diff
@@ -21,4 +28,72 @@ private RefLocations() {
...
+        //if arrive here, it is a relative path
+        String delimiter = csl.endsWith("/") ? "" : "/";
+        String parentFolder = "../"; // this is based on what is discussed in the specs
+
+        String location = csl + delimiter + parentFolder + rawLocation;
+
+        try {
+            return new URI(location).normalize().toString();
+        } catch (Exception e) {
+            return location;
```
jgaleotti: what does an exception mean here? Why the location is returned? Add comments to document this decision and behaviour.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
jgaleotti: computeLocation is a bit misleading, the extractLocation is within computeLocation. Find a better way to name this method.

jgaleotti (changes requested):

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
```diff
@@ -127,6 +316,273 @@ public static List<String> collectRefs(JsonNode node) {
...
+        if (segments.length < 3 || !COMPONENTS.equals(segments[0]) || !INLINABLE.contains(segments[1])) {
+            warnings.add(
+                    "Reference '" + original + "' points at '" + path + "' of another document."
+                            + " Only components/" + SCHEMAS + " and components/" + MESSAGES
+                            + " can be imported.");
+            return null;
+        }
+
+        StringBuilder renamed = new StringBuilder("#/")
+                .append(COMPONENTS).append('/')
```
jgaleotti: replace '/' with constant

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
```diff
@@ -1,24 +1,130 @@
...
+     */
+    private static final String PROTOCOL_SEPARATOR = ":";
+
+    /**
+     * A location naming a host but no protocol, which borrows the protocol of the document
+     * referring to it.
+     */
+    private static final String PROTOCOL_RELATIVE_PREFIX = "//";
+
+    private static final String HTTP_PREFIX = "http" + PROTOCOL_SEPARATOR;
```
jgaleotti: replace "http" with constants

jgaleotti (changes requested):

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
LautaroPetaccio: Done! Moved all of the constants to the `RefLocation.java` file.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/AsyncApiRefResolver.java
LautaroPetaccio: Done!

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
LautaroPetaccio: Done!

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
LautaroPetaccio: WDYT about `resolveDocumentLocation` and `extractLocationPart`? The second one is now documented to be clearer on what it does.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/resolver/RefLocations.java
LautaroPetaccio: It tried to normalize the path with URI, falling back to the raw string if that failed. But URI rejects paths with spaces or Windows separators, and the raw string was unusable too, so those references were silently dropped.

I've change the code and now a plain path resolves through `java.nio.file.Path` and a URL through `URI.resolve`.

jgaleotti (approved):

arcuri82 (approved):

PR #1708, opened 2026-08-24
AsyncAPI 3.x: parse servers, bindings, channel parameters and security

LautaroPetaccio:
Fifth in the AsyncAPI stack, on top of #3.

The previous PRs read what a service can be sent and where those messages travel. This adds the rest of what a client needs: **how to reach the service**.

## What it adds

**`AsyncApiServer`** — host, protocol, `protocolVersion`, `pathname` and variables. The protocol is also how AsyncAPI distinguishes the two incompatible AMQPs: `amqp` is 0-9-1, and 1.0 is the separate `amqp1` protocol. A document declaring no server at all is not an error — plenty describe the contract and leave the broker to deployment, so `servers` is simply empty.

**`AsyncApiChannelBindings`** — only the few fields that change what a client has to do are lifted out; everything else stays in a raw map, so nothing is lost and a later transport can read it without the model growing first.

The Kafka binding is the one that matters most, because it carries its own `topic`, and when present *that* is the topic a client publishes to rather than the channel address. Hence `effectiveAddress(protocol)`, which prefers it for Kafka only and leaves `address` untouched so the document still reads as written. Getting this wrong means publishing to the wrong topic, so it has tests for the override, the non-Kafka case, the no-topic case and the case-insensitive match.

**Channel parameters**, kept verbatim. Resolving a templated address like `tenants/{tenantId}/orders` to a concrete topic is a run-time concern, not a parsing one.

**`AsyncApiSecurityScheme`** — `type` is a free-form string rather than an enum on purpose. AsyncAPI's catalogue is broad and broker-specific (`userPassword`, `scramSha512`, `gssapi`, `X509`, `oauth2`, …) and parsing must not fail on a scheme that is perfectly valid but that no transport here can use yet. Whether it can actually be honoured is the connecting client's question. Schemes may also be written inline where they are used rather than in `components`, which real documents do, and those are registered under a name derived from where they appear.

## Bindings and parameters are dereferenced

Both can be shared through `components`, and reading them without following the `$ref` is worse than not reading them at all: a channel whose real Kafka topic sits behind a reference would silently keep its declared address. That is a *wrong* answer rather than a missing one, so there is a test for it specifically.

## Testing

71 tests across the package, 13 of them new here.
---

*Fourth in the AsyncAPI stack. Based on #1707.*

### DIFF
--- a/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
+++ b/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
@@ -9,6 +9,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * An entry under {@code channels:}: one addressable place on the broker (a Kafka topic, an AMQP
@@ -167,24 +168,27 @@ public class AsyncApiChannel {
         private AsyncApiChannelBindings bindings = AsyncApiChannelBindings.none();
 
         private Builder(String name) {
-            this.name = name;
+            this.name = Objects.requireNonNull(name, "name");
         }
 
         public Builder address(String address) { this.address = address; return this; }
 
-        public Builder servers(List<String> servers) { this.servers = servers; return this; }
+        public Builder servers(List<String> servers) { this.servers = Objects.requireNonNull(servers, "servers"); return this; }
 
         public Builder messageKeys(Map<String, String> messageKeys) {
-            this.messageKeys = messageKeys;
+            this.messageKeys = Objects.requireNonNull(messageKeys, "messageKeys");
             return this;
         }
 
         public Builder parameters(Map<String, JsonNode> parameters) {
-            this.parameters = parameters;
+            this.parameters = Objects.requireNonNull(parameters, "parameters");
             return this;
         }
 
-        public Builder bindings(AsyncApiChannelBindings bindings) { this.bindings = bindings; return this; }
+        public Builder bindings(AsyncApiChannelBindings bindings) {
+            this.bindings = Objects.requireNonNull(bindings, "bindings");
+            return this;
+        }
 
         public AsyncApiChannel build() {
             return new AsyncApiChannel(this);
--- a/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
+++ b/core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
@@ -1,10 +1,13 @@
 package com.webfuzzing.asyncapi.models;
 
+import com.fasterxml.jackson.databind.JsonNode;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -14,9 +17,20 @@ import java.util.Map;
  */
 public class AsyncApiChannel {
 
+    /**
+     * The one protocol whose binding decides where a message actually goes.
+     */
+    public static final String KAFKA = "kafka";
+
     private final String name;
 
     private final String address;
+
+    /**
+     * Names of the servers this channel is available on, from its {@code servers} array.
+     * Empty means every server, which is the specification's default.
+     */
+    private final List<String> servers;
 
     /**
      * Key is the channel-local message key, which is what a {@code $ref} of the form
@@ -31,10 +45,21 @@ public class AsyncApiChannel {
      */
     private final List<String> messageIds;
 
+    /**
+     * Key is the parameter name, as used by a {@code {placeholder}} in the address.
+     * Value is that parameter's declaration, kept as a raw node.
+     */
+    private final Map<String, JsonNode> parameters;
+
+    private final AsyncApiChannelBindings bindings;
+
     private AsyncApiChannel(Builder builder) {
         this.name = builder.name;
         this.address = builder.address;
+        this.servers = Collections.unmodifiableList(builder.servers);
         this.messageKeys = Collections.unmodifiableMap(new LinkedHashMap<>(builder.messageKeys));
+        this.parameters = Collections.unmodifiableMap(builder.parameters);
+        this.bindings = builder.bindings;
         //distinct, as two local keys may well point at the same message definition
         this.messageIds = Collections.unmodifiableList(
                 new ArrayList<>(new LinkedHashSet<>(this.messageKeys.values())));
@@ -57,10 +82,19 @@ public class AsyncApiChannel {
      *
      * Null on purpose: the specification allows an explicit {@code address: null} to say the
      * address is not known statically and is determined at run time. It may also contain
-     * {@code {parameter}} placeholders, which are left in place here.
+     * {@code {parameter}} placeholders, which are left in place here -- see
+     * {@link #getParameters()}.
      */
     public String getAddress() {
         return address;
+    }
+
+    /**
+     * Names of the servers this channel is available on, from the channel's {@code servers}
+     * array. Empty means "all of them", which is the specification's default.
+     */
+    public List<String> getServers() {
+        return servers;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
```diff
@@ -84,24 +118,74 @@ public List<String> getMessageIds() {
...
+        private Map<String, JsonNode> parameters = Collections.emptyMap();
+        private AsyncApiChannelBindings bindings = AsyncApiChannelBindings.none();
 
         private Builder(String name) {
             this.name = name;
         }
 
         public Builder address(String address) { this.address = address; return this; }
 
+        public Builder servers(List<String> servers) { this.servers = servers; return this; }
```
jgaleotti: can servers be null ? Add Objects.requireNonNull() for all cases where the expected type is non-null

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
```diff
@@ -84,24 +118,74 @@ public List<String> getMessageIds() {
...
+
         public Builder messageKeys(Map<String, String> messageKeys) {
             this.messageKeys = messageKeys;
```
jgaleotti: comment above

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannelBindings.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    private final Map<String, JsonNode> raw;
+
+    private AsyncApiChannelBindings(Builder builder) {
+        this.kafkaTopic = builder.kafkaTopic;
+        this.amqpIs = builder.amqpIs;
+        this.amqpQueue = builder.amqpQueue;
+        this.amqpExchange = builder.amqpExchange;
+        this.wsMethod = builder.wsMethod;
+        this.raw = Collections.unmodifiableMap(builder.raw);
```
jgaleotti, line 40: can builder.raw be null?

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiOperation.java
```diff
@@ -49,6 +64,8 @@ private AsyncApiOperation(Builder builder) {
         this.channelName = builder.channelName;
         this.messageIds = Collections.unmodifiableList(builder.messageIds);
```
jgaleotti, line 66: can messageIds be null?

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
```diff
@@ -380,6 +411,132 @@ private static AsyncApiCorrelationId parseCorrelationId(
...
+        if (host == null || protocol == null) {
+            String missing = host == null && protocol == null
+                    ? "host and protocol"
+                    : (host == null ? "host" : "protocol");
+            warnings.add("Server '" + name + "' declares no " + missing + ", and is ignored");
+            return null;
+        }
+
+        Map<String, AsyncApiServerVariable> variables = new LinkedHashMap<>();
+        for (Map.Entry<String, JsonNode> entry : objectFieldsOf(node.get("variables")).entrySet()) {
```
jgaleotti: add local variable to node.get("variables").

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
```diff
@@ -380,6 +411,132 @@ private static AsyncApiCorrelationId parseCorrelationId(
...
+            JsonNode variable = entry.getValue();
+            variables.put(entry.getKey(), new AsyncApiServerVariable(
+                    entry.getKey(),
+                    scalarOf(variable.get("default")),
```
jgaleotti: add string constants for all strings ("default", "enum", etc)

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
```diff
@@ -147,18 +167,29 @@ public static AsyncApiDocument parse(
         Map<String, AsyncApiOperation> operations = new LinkedHashMap<>();
         for (Map.Entry<String, JsonNode> entry : objectFieldsOf(root.get("operations")).entrySet()) {
```
jgaleotti: add variable to root.get("operations")

jgaleotti (changes requested):

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
LautaroPetaccio: Done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
LautaroPetaccio: Done for all 53 keywords the parser reads, not just these.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/parser/AsyncApiParser.java
LautaroPetaccio: Done, and for the other four section loops that read the same way (channels, operations, servers, channel messages), so they all read alike.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannelBindings.java
LautaroPetaccio: Not today (the parser always passes a fresh map), but the setter now rejects it explicitly, so the constructor can't be reached with one.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
LautaroPetaccio: Not from a document. The spec makes it optional, and the parser folds an absent servers into an empty list before the model sees it. Added requireNonNull anyway on every required setter and builder argument, since these are public API on a module meant to become a library.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
LautaroPetaccio: Same as above, done.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiOperation.java
LautaroPetaccio: Same, never from the parser, now guarded at the setter.

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannel.java
```diff
@@ -84,21 +119,74 @@ public List<String> getMessageIds() {
...
+        /** @see AsyncApiChannel#parameters */
+        private Map<String, JsonNode> parameters = Collections.emptyMap();
+        private AsyncApiChannelBindings bindings = AsyncApiChannelBindings.none();
 
         private Builder(String name) {
-            this.name = name;
+            this.name = Objects.requireNonNull(name, "name");
         }
 
         public Builder address(String address) { this.address = address; return this; }
```
jgaleotti, line 174: address could be null?

core-extra/asyncapi-parser/src/main/java/com/webfuzzing/asyncapi/models/AsyncApiChannelBindings.java
jgaleotti: as a programming discipline, please consider adding non nullity checks whenever required, even when the expected input can never be null

jgaleotti (approved):

arcuri82 (approved):

PR #1711, opened 2026-08-24
AsyncAPI 3.x: the action and individual a search holds

LautaroPetaccio:
Eighth in the AsyncAPI stack, on top of #6. Turns the parsed document into something a search can sample and mutate.

## The action

`AsyncApiAction` is one message published on one channel, following the shape of `RPCCallAction` — the closest analogue in EvoMaster, being a call with no URL, made through a driver, whose mutable state is the input and whose response is read afterwards rather than searched over.

**What is deliberately not a gene is the design.**

- **The address**, which the contract fixes. Sending to a channel the service does not read would only waste executions.
- **The correlation id**, stamped fresh at each execution. Searching over it could achieve nothing — the service only echoes it back — and pairing a reply with its request needs a value unique per execution rather than one carried in the genome. So the header it is stamped into is left out of the header genes; and when it is the *only* header, there are no header genes at all, since an empty schema would otherwise build a free-form map gene and invite the search to invent headers the contract never declared.
- **The reply**, which is an observation. It is read at fitness time to decide what was covered, exactly as RPC reads its response.

## The individual

`AsyncApiIndividual` is one individual for **every** transport rather than one per transport. Kafka versus AMQP versus a socket appears nowhere in it: what is searched over is the operation and the payload, which are the same whatever moves the bytes, and which wire is used is decided below the driver interface. Were the transport to leak in here there would have to be a `KafkaIndividual` and an `AmqpIndividual` with nothing shared between them.

## The builder

`AsyncApiActionBuilder` fills an action cluster, keeping the contract `addActionsFromSwagger` and `addActionsFromSchema` already use: one entry per action name, and a list of what had to be skipped rather than an exception.

Two decisions worth review:

- **Only operations the service consumes become actions.** A `send` operation is what the service emits — something to subscribe to, not to drive. They are counted as skipped rather than built.
- **An operation carrying several messages becomes several actions.** Which message to send is a choice the search makes by *picking an action*, not by mutating a gene. The message is named in the action only when there is more than one (`recv_list_legs` rather than `recv_list_legs:listLegs`), so a generated test reads as the operation wherever it can.

## One thing not copied across

`RPCIndividual.seeIndexedRPCCalls` looks up children by action type, but the children are `EnterpriseActionGroup`s, so it returns nothing. It has no callers, so it appears to be dead. I wrote the same method here, watched a test fail on it, and removed it rather than carrying the bug over — `seeMainExecutableActions` does the job. Worth a look on the RPC side.

## Testing

15 new tests in `AsyncApiActionBuilderTest`, covering the action cluster built from NCS, microcks and a WebSocket document, what an action does and does not expose as genes, adding and removing messages from an individual, and copying one. 108 across the AsyncAPI package.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
+++ b/core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
@@ -33,17 +33,40 @@ object AsyncApiGeneBuilder {
      */
     private const val INLINE_PREFIX = "_asyncapi_"
 
+    /*
+        The JSON Schema keywords this builder reads or rewrites. These are JSON Schema's
+        vocabulary, not AsyncAPI's, which is why they live here rather than with the parser.
+     */
+    private const val PROPERTIES = "properties"
+    private const val PATTERN_PROPERTIES = "patternProperties"
+    private const val DEFINITIONS = "definitions"
+    private const val DEFS = "\$defs"
+    private const val REQUIRED = "required"
+    private const val CONST = "const"
+    private const val DEFAULT = "default"
+    private const val ENUM = "enum"
+    private const val EXAMPLE = "example"
+    private const val EXAMPLES = "examples"
+    private const val TYPE = "type"
+    private const val MINIMUM = "minimum"
+    private const val MAXIMUM = "maximum"
+
+    private const val TYPE_STRING = "string"
+    private const val TYPE_INTEGER = "integer"
+    private const val TYPE_NUMBER = "number"
+    private const val TYPE_BOOLEAN = "boolean"
+
     /**
      * The keywords whose value is literal data rather than a schema, so nothing inside them is
      * a keyword either.
      */
-    private val DATA_KEYWORDS = setOf("const", "default", "enum", "example", "examples")
+    private val DATA_KEYWORDS = setOf(CONST, DEFAULT, ENUM, EXAMPLE, EXAMPLES)
 
     /**
      * The keywords whose value maps arbitrary names to schemas. Their keys come from the
      * document, so a field a service happens to call "const" or "default" must still be walked.
      */
-    private val SCHEMA_MAPS = setOf("properties", "patternProperties", "definitions", "\$defs")
+    private val SCHEMA_MAPS = setOf(PROPERTIES, PATTERN_PROPERTIES, DEFINITIONS, DEFS)
 
     /**
      * The genes for a message's payload, or null when it declares none.
@@ -90,14 +113,14 @@ object AsyncApiGeneBuilder {
         }
 
         val field = correlation.fieldName ?: return headers
-        val properties = headers.get("properties")
+        val properties = headers.get(PROPERTIES)
 
         if (properties == null || !properties.has(field)) {
             return headers
         }
 
         val copy = headers.deepCopy<JsonNode>() as ObjectNode
-        val kept = (copy.get("properties") as ObjectNode).apply { remove(field) }
+        val kept = (copy.get(PROPERTIES) as ObjectNode).apply { remove(field) }
 
         /*
             When the stamped id was the only header, there is nothing left to vary. Returning
@@ -109,11 +132,11 @@ object AsyncApiGeneBuilder {
         }
 
         //a field that is no longer there cannot be required either
-        (copy.get("required") as? ArrayNode)?.let { required ->
+        (copy.get(REQUIRED) as? ArrayNode)?.let { required ->
             val kept = required.filter { it.asText() != field }
-            copy.remove("required")
+            copy.remove(REQUIRED)
             if (kept.isNotEmpty()) {
-                copy.putArray("required").apply { kept.forEach { add(it) } }
+                copy.putArray(REQUIRED).apply { kept.forEach { add(it) } }
             }
         }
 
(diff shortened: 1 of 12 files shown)


core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
```diff
@@ -45,12 +47,66 @@ object AsyncApiGeneBuilder {
...
+            When the stamped id was the only header, there is nothing left to vary. Returning
+            the empty schema would build a free-form map gene, which is worse than nothing: it
+            would invite the search to invent headers the contract never declared.
+         */
+        if (kept.isEmpty) {
+            return null
+        }
+
+        //a field that is no longer there cannot be required either
+        (copy.get("required") as? ArrayNode)?.let { required ->
```
jgaleotti: replace strings with constants

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.evomaster.core.database.sql.SqlAction
+import kotlin.math.max
+
+/**
+ * A sequence of messages to publish, and whatever is needed to set the service up first.
+ *
+ * There is one individual for every transport rather than one per transport: Kafka versus AMQP
+ * versus a socket appears nowhere in here. What is being searched over is the operation and the
+ * payload, which are the same whatever moves the bytes; which wire is used is decided below the
+ * driver interface. Were the transport to leak in here there would have to be a
```
jgaleotti, line 20: Where the transport?

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
```diff
@@ -0,0 +1,116 @@
...
+    override fun copyContent(): AsyncApiIndividual =
+        AsyncApiIndividual(
+            sampleType,
+            trackOperator,
+            index,
+            children.map { it.copy() }.toMutableList() as MutableList<ActionComponent>,
+            mainSize = groupsView()!!.sizeOfGroup(GroupsOfChildren.MAIN),
+            sqlSize = groupsView()!!.sizeOfGroup(GroupsOfChildren.INITIALIZATION_SQL),
+            mongoSize = groupsView()!!.sizeOfGroup(GroupsOfChildren.INITIALIZATION_MONGO),
+            redisSize = groupsView()!!.sizeOfGroup(GroupsOfChildren.INITIALIZATION_REDIS),
```
jgaleotti: I think that we should keep redis outside ASYNC API until it is fully supported in REST

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
LautaroPetaccio: Dropped the `redisSize` parameter, the group is fixed at zero, and `copyContent` no longer measures it.

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
LautaroPetaccio: Done everywhere in this PR.

jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
```diff
@@ -260,11 +339,11 @@ object AsyncApiGeneBuilder {
      * The keywords whose value is literal data rather than a schema, so nothing inside them is
      * a keyword either.
      */
-    private val DATA_KEYWORDS = setOf("const", "default", "enum", "example", "examples")
+    private val DATA_KEYWORDS = setOf(CONST, DEFAULT, ENUM, EXAMPLE, EXAMPLES)
```
arcuri82: recall what stated in `for_developers.md`. do not declare variables in the middle of the file, but at the top of the class

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
```diff
@@ -260,11 +339,11 @@ object AsyncApiGeneBuilder {
...
 
     /**
      * The keywords whose value maps arbitrary names to schemas. Their keys come from the
      * document, so a field a service happens to call "const" or "default" must still be walked.
      */
-    private val SCHEMA_MAPS = setOf("properties", "patternProperties", "definitions", "\$defs")
+    private val SCHEMA_MAPS = setOf(PROPERTIES, PATTERN_PROPERTIES, DEFINITIONS, DEFS)
```
arcuri82: do not declare variables in the middle/end of the file

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
```diff
@@ -0,0 +1,121 @@
...
+ */
+class AsyncApiIndividual(
+    sampleType: SampleType,
+    trackOperator: TrackOperator? = null,
+    index: Int = -1,
+    allActions: MutableList<ActionComponent>,
+    mainSize: Int = allActions.size,
+    sqlSize: Int = 0,
+    mongoSize: Int = 0,
+    dnsSize: Int = 0,
```
arcuri82, line 41: we are adding several more DB to EvoMaster these days. no need to handle them here, for the moment. but can leave a TODO comment about it

arcuri82 (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
LautaroPetaccio: Done.

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
LautaroPetaccio: Done.

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/builder/AsyncApiGeneBuilder.kt
LautaroPetaccio: Done.

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/data/AsyncApiIndividual.kt
LautaroPetaccio: I've simplified the comment, it was too verbose for what it needed to describe.

arcuri82 (approved):

PR #1717, opened 2026-08-27
Fix minimization bug in sql query parser

suarezrominajulieta:


client-java/sql/src/main/java/org/evomaster/client/java/sql/internal/SqlHandler.java
```diff
@@ -237,11 +237,23 @@ public List<SqlCommandWithDistance> getSqlDistances(List<InsertionDto> successfu
...
-                    SqlDistanceWithMetrics sqlDistance = computeDistance(sqlCommand,
-                            parsedStatement,
-                            successfulInitSqlInsertions,
-                            queryFromDatabase);
+                    /*
+                        There are SQL constructs we cannot analyze yet, which throw exceptions.
+                        Those must not escape from here: the heuristics are computed in the same
+                        request that returns the coverage, so an exception here would make us lose
+                        the whole coverage of the test, and not just this single heuristic.
+                     */
```
jgaleotti, line 245: What is the SQL query that is failing? Have you tried with heuristicsForSQLAdvanced = true ?

suarezrominajulieta: Hi! The querys are these:

SELECT DISTINCT t.tag              FROM ref r CROSS JOIN LATERAL jsonb_array_elements_text(...)
SELECT t.tag, COUNT(DISTINCT r.url) FROM ref r CROSS JOIN LATERAL jsonb_array_elements_text(...)

This triggers an exception and causes the SUT to restart, resulting in the loss of the list of IDs maintained within the driver/SUT and leading to duplicate counts in EvoMaster. Minimization, on the other hand, does not produce duplicates; it skips broken tests while maintaining a single list. That is why coverage is lost during minimization—in reality, those lines shouldn't be there.

I hadn't noticed that option, though; I'll give it a try.

Haven't finished the analysis and the fixes for the two bugs I found, but i could find the 414 we have talked about. 

this pr is incomplete yet.

suarezrominajulieta: Hi, yes, with that key on, the code is not fixed, but the exception is catched, thus, this problem, would be bypassed

PR #1723, opened 2026-08-29
Fix escaping response assertion field paths in generated JVM tests

hoco-scy:
## Summary

This PR fixes escaping of response assertion field paths when EvoMaster generates Java or Kotlin tests.

A response field name is embedded in two different syntactic layers: first as a single-quoted Groovy GPath segment used by RestAssured, and then as a Java/Kotlin string literal. Characters such as a backslash or a single quote must therefore be escaped for both layers, in the correct order.

## Problem observed in real experiments

We found the issue while repeatedly generating and executing tests for `tiltaksgjennomforing`.

Two concrete generated field paths caused deterministic failures:

1. A response field containing a backslash generated:

   ```java
   .body("'\g'.'name'", containsString("disabled"));
   ```

   The raw backslash reached the generated Java source and produced an `illegal escape character` compilation error.

2. A response field containing a single quote, such as the value below:

   ```text
   <img src=x onerror=alert('XSS')>
   ```

   generated:

   ```java
   .body("'<img src=x onerror=alert('XSS')>'", equalTo(false));
   ```

   The inner quote prematurely closed the Groovy GPath field segment. RestAssured then parsed the remaining text as a Groovy expression, which led to errors such as an undefined `XSS` parameter or `MissingPropertyException`.

These failures were reproducible across repeated executions because the generated source itself was invalid; they were not caused by instability in the system under test.

## Root cause

The previous implementation inserted field names into a single-quoted GPath segment and only applied the escaping needed by the generated JVM source.

That handles one layer, but not both:

1. the field name must first be escaped as content inside a single-quoted Groovy GPath segment;
2. the resulting GPath must then be escaped as content inside a Java/Kotlin string literal.

If the first step is skipped, a quote changes the GPath structure and a backslash may be interpreted by either Groovy or the JVM compiler before it reaches RestAssured.

## Changes

This PR adds a dedicated escaping path for response assertion field names:

- escape backslashes and single quotes for the single-quoted Groovy GPath segment;
- apply `GeneUtils.applyEscapes` afterwards so the already escaped GPath is represented correctly in generated Java/Kotlin source;
- use the same handling for both Java and Kotlin output;
- add regression coverage for ordinary names, quotes, backslashes, and combinations of both characters.

For example, the generated assertion for the XSS-shaped field becomes:

```java
.body("'<img src=x onerror=alert(\\'XSS\\')>'", equalTo(false));
```

and backslashes are preserved as literal field-name characters instead of becoming invalid JVM escapes.

### Unit/regression tests

The focused `TestCaseWriterTest` suite passes after rebasing onto the current upstream `master`:

- 41 tests run
- 0 failures
- 0 errors

### End-to-end validation

We also rebuilt EvoMaster with this change and ran a targeted validation against `tiltaksgjennomforing` using response paths that previously triggered both failure modes.

Results:

- 10 independently generated logical test suites
- 20 JUnit XML result files
- 3,806 executed tests
- 3,806 passed
- 0 failures
- 0 errors
- 0 skipped tests
- all 10 suite runs completed with exit code 0

The corrected generated tests contained three field paths with embedded single quotes and three with backslashes. None of the earlier `illegal escape character`, undefined `XSS` parameter, or `MissingPropertyException` signatures appeared.

### DIFF
--- a/core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
+++ b/core/src/main/kotlin/org/evomaster/core/output/service/ApiTestCaseWriter.kt
@@ -306,22 +306,23 @@ abstract class ApiTestCaseWriter : TestCaseWriter() {
 
                     var needsDot = true
 
-                    val fieldName = if (format.isJava()) {
-                        "'${it.key}'"
-                    } else if (format.isKotlin()){
-                        "'${handleDollarSign(it.key)}'"
+                    val fieldName = if (format.isJavaOrKotlin()) {
+                        "'${escapeJvmGPathFieldName(it.key)}'"
                     } else if (format.isJavaScript()) {
                         //field name could have any character... need to use [] notation then
                         if (it.key.matches(Regex("^[a-zA-Z][a-zA-Z0-9]*$"))) {
                             it.key
                         } else {
                             needsDot = false
-                            "[\"${it.key}\"]"
+                            "[\"${escapeFieldNameForStringLiteral(it.key)}\"]"
                         }
                     } else if (format.isPython()) {
                         needsDot = false
-                        "[\"${it.key}\"]"
-                    //TODO need to deal with '' C#? see EscapeRest
+                        "[\"${escapeFieldNameForStringLiteral(it.key)}\"]"
+                    // C# is no longer a supported output format, but keep the legacy branch safe.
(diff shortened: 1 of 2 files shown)


man-zhang:
@arcuri82 this PR is ready for you to review.

arcuri82 (approved):

PR #1715, opened 2026-08-25
Read the live Neo4j graph through the driver

andyfelder16:
Third of the PRs splitting #1641, on top of #1580, #1672, #1647 and #1697, all of which are merged.

- `Neo4jGraphReader` reads the whole graph into an in-memory `Neo4jGraph`, the `G` the calculator scores a query against. Two read-only Cypher queries return only primitive projections (ids, label and type names, property maps), never the driver's own `Node` and `Relationship` types, which keeps the reflected surface small
- `ReflectionBasedNeo4jClient` holds all the driver reflection behind a typed API, so the reader has none of its own. Same shape as `ReflectionBasedRedisClient`. It lives here because the reader is its first user; the Neo4j test-data insertion will be the second, and it already carries what that needs and the reader does not: a parameterised `run` overload and a `detachDeleteAll` for resetting between tests
- Reflection is used so that EvoMaster never has a compile-time dependency on `neo4j-java-driver` and cannot clash with whichever version the SUT ships
- 11 tests, including that the session is closed when a query fails, and that a driver whose API does not match fails naming the method it could not find

PR #1580, opened 2026-06-09
add Neo4j Cypher query parser

andyfelder16:
- Neo4j Cypher query parser
- Built on Neo4j's Cypher25 ANTLR grammar ([source](https://github.com/neo4j/neo4j/tree/release/5.26.0/community/cypher/front-end/parser/v25/parser/src/main/antlr4/org/neo4j/cypher/internal/parser/v25))
- Parses `MATCH` into a structural pattern + typed condition model
- Node/relationship patterns, label/type expressions, properties, and full `WHERE` support
- Variable-length and quantified path patterns
- 92 parser tests

PR #1731, opened 2026-09-01
Wire Neo4j heuristics into the search fitness function

andyfelder16:
Last of the PRs splitting #1641, on top of #1580, #1672, #1647, #1697 and #1715, all of which are merged. It closes the chain: the Cypher queries the SUT runs have been captured since #1460, and this is what finally turns them into search guidance.

- `Neo4jHandler` collects the queries intercepted from `Session.run`, reads the graph once per action rather than once per query, and scores each query against that snapshot. Only MATCH queries are scored; anything that does not parse as one, such as a write, is skipped
- `Neo4jCommandWithDistance` and `Neo4jDistanceWithMetrics` carry the result to the DTO
- `EnterpriseFitness.handleNeo4jHeuristics` feeds the distances to the search, behind the new `heuristicsForNeo4j` option, and reports evaluation counts and the average number of nodes inspected through `Statistics`
- Nothing throws back at the SUT: a driver that cannot be queried, a query that does not parse, and a scoring failure each degrade to no heuristic or to a failed evaluation, and the search carries on
- 5 handler tests plus one in `StatisticsTest`, covering the scored path, a write being skipped, no driver, heuristics disabled, and a driver that cannot be reached


client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Creates a Neo4j heuristic result.
+     *
+     * @param distance normalized distance to satisfying the query, 0 meaning satisfied
+     * @param numberOfEvaluatedNodes number of graph nodes considered
+     * @param evaluationFailure whether the evaluation failed
+     */
+    public Neo4jDistanceWithMetrics(double distance, int numberOfEvaluatedNodes, boolean evaluationFailure) {
+        if (distance < 0.0d || distance > 1.0d || Double.isNaN(distance)) {
```
arcuri82, line 20: hi @andyfelder16 .thx ;) but should first request @jgaleotti  review, and, then, once that is completed, JP asks for my review. anyway, with a  quick look here i can see the naming is confusing. see `DistanceHelper` for a definition of `distance` vs `heuristic` inside EM codebase

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
jgaleotti: Hi @arcuri82 , I am taking over the review. However, this is indeed a distance since it follows the same model as MongoDistanceWithMetrics, RedisDistanceWithMetrics, DynamoDbDistanceWithMetrics and CqlDistanceWithMetrics (i.e., Cassandra)

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jHandler.java
```diff
@@ -0,0 +1,144 @@
...
+                continue;
+            }
+
+            Neo4jDistanceWithMetrics metrics;
+            try {
+                double distance = calculator.computeDistance(parsedQuery, graph);
+                metrics = new Neo4jDistanceWithMetrics(distance, graph.nodeCount(), false);
+            } catch (Exception e) {
+                SimpleLogger.uniqueWarn("Failed to compute Neo4j heuristic for query: " + query);
+                metrics = new Neo4jDistanceWithMetrics(1.0, graph.nodeCount(), true);
```
jgaleotti: replace 1.0 with Neo4HeuristicsCalculator.MAX_NEO4J_DISTANCE = 1.0d

core/src/main/kotlin/org/evomaster/core/problem/enterprise/service/EnterpriseFitness.kt
```diff
@@ -551,4 +555,37 @@ abstract class EnterpriseFitness<T> : FitnessFunction<T>() where T : Individual
...
+            val toMinimize = extra.heuristics
+                .filter {
+                    it != null
+                            && it.objective == ExtraHeuristicEntryDto.Objective.MINIMIZE_TO_ZERO
+                            && it.type == ExtraHeuristicEntryDto.Type.NEO4J
+                }.map { it.value }
+                .toList()
+
+            if (toMinimize.isNotEmpty()) {
+                fv.setExtraToMinimize(i, toMinimize)
```
jgaleotti: this code is behind master

jgaleotti (changes requested):

jgaleotti (approved):

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
arcuri82: if this is distance, why are we throwing an exception if greater than 1? this is confusing. if it is a normalized distance, then it should be explicitely called that, eg, `double normalizedDistance`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+            throw new IllegalArgumentException("distance must be between 0 and 1, but was " + distance);
```
arcuri82: a distance does not need to be within 0  and 1. if it is a normalized distance, then explicitely state so, eg, `a normalized distance must be between 0 and 1,`

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+public final class Neo4jDistanceWithMetrics {
+
+    private final double distance;
+    private final int numberOfEvaluatedNodes;
+    private final boolean evaluationFailure;
+
+    /**
+     * Creates a Neo4j heuristic result.
+     *
+     * @param distance normalized distance to satisfying the query, 0 meaning satisfied
```
arcuri82: any specific reason why Neo4j works on normalized distances?

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jDistanceWithMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+        this.distance = distance;
+        this.numberOfEvaluatedNodes = numberOfEvaluatedNodes;
+        this.evaluationFailure = evaluationFailure;
+    }
+
+    /**
+     * @return normalized distance to satisfying the query, 0 meaning satisfied
+     */
+    public double getDistance() {
```
arcuri82, line 47: if normalized, then the name `getDistance` is confusing.

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/SutController.java
```diff
@@ -546,6 +561,34 @@ public final void computeMongoHeuristics(ExtraHeuristicsDto dto, List<Additional
...
 
+    public final void computeNeo4jHeuristics(ExtraHeuristicsDto dto, List<AdditionalInfo> additionalInfoList){
+        if(neo4jHandler.isCalculateHeuristics()){
+            if(!additionalInfoList.isEmpty()) {
+                AdditionalInfo last = additionalInfoList.get(additionalInfoList.size() - 1);
+                last.getNeo4JInfoData().forEach(it -> {
+                    try {
+                        neo4jHandler.handle(it);
+                    } catch (Exception e){
+                        SimpleLogger.error("FAILED TO HANDLE NEO4J COMMAND");
```
arcuri82: could add the `e.getMessage` to the this error log

client-java/controller/src/main/java/org/evomaster/client/java/controller/neo4j/heuristics/Neo4jHeuristicsCalculator.java
```diff
@@ -37,6 +37,13 @@ public class Neo4jHeuristicsCalculator {
      */
     public static final double C = DistanceHelper.H_NOT_NULL;
 
+    /**
+     * Largest distance {@link #computeDistance} can report: the query matched nothing at all. A caller that
+     * cannot compute a distance (for example when the evaluation throws) reports this value, so a failure
+     * never looks closer to satisfied than a genuine miss.
+     */
+    public static final double MAX_NEO4J_DISTANCE = 1.0d;
```
arcuri82: why does Neo4J require this bound, where all other systems do not?  note that normalizing distances do lose information. that is why in `DistanceHelper` there are a lot of different helper functions to deal then a maxed distance is `Double.MAX_VALUE `

arcuri82 (approved):

arcuri82 (changes requested):
approved by mistake, there are comments to address first

client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/neo4j/Neo4jHandler.java
jgaleotti: the value of MAX_NEO4J_DISTANCE should be Double.MAX_VALUE instead of 1.0d

jgaleotti (approved):

arcuri82 (approved):

PR #1738, opened 2026-09-05
AsyncAPI 3.x: the driver contract for publishing a message

LautaroPetaccio:
Ninth in the AsyncAPI stack, on top of #7. **This one is entirely in `client-java` — no core changes**, so it reviews as controller-side plumbing on its own.

Everything so far reads a document. This is the other side: what a driver must provide so the core can drive an AsyncAPI service without knowing anything about brokers.

## Why a driver at all, even for black-box

Unlike REST and GraphQL, there is no universal wire to point at — Kafka, AMQP, MQTT and WebSocket share nothing at the API level. So something has to hold a client and move the bytes, and that something is the driver. This is the one way AsyncAPI is unlike every other black-box mode EvoMaster ships, and it is the same reason RPC needs a driver.

## `AsyncApiProblem`

Carries only the document — a location to fetch it from, or the text itself. The connection to the broker stays in the driver and never crosses, which is what keeps core free of any broker dependency.

The inline-text option is worth noting, since REST has little need for it: a REST service usually serves its own contract over HTTP, whereas a service that speaks only Kafka has no endpoint to serve anything from. Its document is far more likely to be a file shipped beside it.

## `SutController.executeAsyncApiAction`

Where a driver publishes one message and, when a reply is expected, waits for the one that answers it — the counterpart of `executeAction` for RPC.

It has a **default that throws rather than being abstract**, so adding it does not break any existing driver; the message names what to override. Only publish and await are protocol-specific, and they never leave the driver.

## The driver reports, it does not judge

`AsyncApiReplyDto` distinguishes four outcomes, because they mean different things and only one is a fault:

- published, no reply expected — a fire-and-forget operation did what it could
- published, and a reply arrived — the only case with something to classify
- published, and nothing arrived within the window — the contract promised a reply and did not deliver, though a slow service and a stuck one look alike from outside
- could not be published at all — a **broken setup**, not a finding about the service, which is why it is reported separately from silence

Deciding what an outcome *means* is the core's job, so that it means the same thing whatever the transport.

The reply also carries whether the correlation id came back. Whether correlation works cannot be read off a contract — echoing the id is the service's own behaviour — so it is established by watching for it rather than assumed.

## Wiring

A field on `SutInfoDto`, a field on `ActionDto`, and two branches in `EMController`.

## Testing

Six tests in `AsyncApiProblemTest`, including one asserting that a driver pointed at an AsyncAPI service without implementing the hook fails with a message telling it what to override.

`client-java/controller` runs **740 tests, 0 failures, 34 errors** — identical with and without this change. Those 34 are testcontainers suites failing because Docker is not reachable in my environment; I verified the same counts on a pristine tree before claiming that.


client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * What the document declares the payload is encoded as, eg "application/json".
+     */
+    public String contentType;
+
+    /**
+     * Headers to publish alongside the body, for a transport that has them.
+     */
+    public Map<String, String> headers = new LinkedHashMap<>();
```
jgaleotti, line 58: add a description of what the keys and values are in maps. This programming discipline is specified in docs/for_developers.md

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * How long to wait for a reply before giving up, in milliseconds.
+     *
+     * There is no right answer here: a slow service and a stuck one look the same from outside,
+     * so this is a tuning parameter with no equivalent in a synchronous protocol. It is set
+     * generously and reported with the result.
+     */
+    public Long replyTimeoutMs;
+
+    public static final String CORRELATION_IN_HEADER = "HEADER";
```
jgaleotti: all constants should be at the beginning of the class

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,75 @@
...
+ *   <li>could not be published at all -- a broken setup rather than a finding about the
+ *       service, which is why it is reported separately from silence.</li>
+ * </ul>
+ */
+public class AsyncApiReplyDto {
+
+    /**
+     * The index of the action this answers, echoing what was asked.
+     */
+    public Integer index;
```
jgaleotti, line 26: can this value be null?

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
```diff
@@ -0,0 +1,93 @@
...
+    public String replyAddress;
+
+    /**
+     * How long to wait for a reply before giving up, in milliseconds.
+     *
+     * There is no right answer here: a slow service and a stuck one look the same from outside,
+     * so this is a tuning parameter with no equivalent in a synchronous protocol. It is set
+     * generously and reported with the result.
+     */
+    public Long replyTimeoutMs;
```
jgaleotti, line 98: can this value be null?

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * The reply body, as it arrived.
+     */
+    public String replyPayload;
+
+    /**
+     * The reply's headers, for a transport that has them.
+     */
+    public Map<String, String> replyHeaders = new LinkedHashMap<>();
```
jgaleotti, line 54: describe what key,value are as specified in docs/for_developers.md

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,75 @@
...
+     * recorded rather than treated as a fault, since from outside there is no telling a defect
+     * from a service that correlates by some business key instead.
+     */
+    public boolean correlationMatched;
+
+    /**
+     * How long the driver waited, in milliseconds, whether or not anything arrived. Reported
+     * because the verdict on silence is only meaningful alongside how long it was waited for.
+     */
+    public Long waitedMs;
```
jgaleotti, line 72: can this value be null?

jgaleotti (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Yes, when the driver did not wait at all, that is a fire and forget operation, or a message that could not be published. I've written it in the javadoc.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Done!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
LautaroPetaccio: Yes, when no reply is expected (replyAddress null), there is nothing to wait for. I've written it in the javadoc.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
LautaroPetaccio: Done!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
LautaroPetaccio: Done!

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: It can't. It basically holds the same value as `ActionDto.index`, which `newAction` has already set by then, so a null never reaches this line. It is Integer to match `ActionDto.index` and `ActionResponseDto.index`. 
EMController almost never assigns a DTO through constructors and most of the DTOs are plain classes without validations.
WDYT I should do in this case? Should I create this DTO to have a constructor with the integer with a validation?

jgaleotti: I think you should use int instead of Integer if this field is never null.

jgaleotti (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: I've converted it to int.

jgaleotti (approved):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,78 @@
...
+ *   <li>could not be published at all -- a broken setup rather than a finding about the
+ *       service, which is why it is reported separately from silence.</li>
+ * </ul>
+ */
+public class AsyncApiReplyDto {
+
+    /**
+     * The index of the action this answers, echoing what was asked.
+     */
+    public int index;
```
arcuri82: DTOs should NEVER use primitive values, as missing values would be silently converted to defaults

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Whether the message reached the broker. False means the driver could not publish, and
+     * {@link #errorMessage} says why.
+     */
+    public boolean published;
```
arcuri82: no primitive values in DTOs

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Whether a reply arrived and was recognised as answering this message.
+     */
+    public boolean replyReceived;
```
arcuri82: no primitive values in DTOs

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
arcuri82: eg, use `Integer` here

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
arcuri82: eg use `Boolean` here

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Whether the driver waited for a reply at all. False for a fire-and-forget operation, so
+     * that the absence of a reply is not mistaken for silence in answer to a promise.
+     */
+    public boolean replyExpected;
```
arcuri82: no primitive values in DTOs

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Whether the reply carried back the correlation id that was stamped on the request.
+     *
+     * This is the honest answer to "did correlation work", which cannot be read off a contract:
+     * echoing the id is the service's own behaviour. A reply that arrives without it is
+     * recorded rather than treated as a fault, since from outside there is no telling a defect
+     * from a service that correlates by some business key instead.
+     */
+    public boolean correlationMatched;
```
arcuri82: no primitive values in DTOs

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/AsyncApiProblem.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param schemaLocation where the AsyncAPI document can be fetched from: a URL, or a path
+     *                       on the machine running the driver
+     */
+    public AsyncApiProblem(String schemaLocation) {
+        this(schemaLocation, null);
+    }
+
+    private AsyncApiProblem(String schemaLocation, String schemaText) {
+
+        if ((schemaLocation == null) == (schemaText == null)) {
```
arcuri82: bit convoluted check. for readability, use explicit check, eg:
```
if ((schemaLocation == null && schemaText == null)  || (schemaLocation != null && schemaText != null))
```

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
jgaleotti: Hi @arcuri82 , this was my bad since I asked @LautaroPetaccio to explicitly change all non-nullable fields to their primitive types in Lautaro's commit. Thanks for clarifying this behaviour explicitly for DTOs.

LautaroPetaccio: Done.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Done.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Done.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Done.

client-java/controller/src/main/java/org/evomaster/client/java/controller/problem/AsyncApiProblem.java
LautaroPetaccio: Done.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiReplyDto.java
LautaroPetaccio: Done.

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public String correlationId;
+
+    /**
+     * Where the correlation id has to be written, as the document declares it. One of
+     * {@link #CORRELATION_IN_HEADER} or {@link #CORRELATION_IN_PAYLOAD}, or null when the
+     * document says nothing, in which case it is up to the driver to decide -- a transport with
+     * native correlation should use it.
+     */
+    public String correlationLocation;
```
arcuri82: in that case, shouldn't those values be in an enum?

arcuri82 (changes requested):

client-java/controller-api/src/main/java/org/evomaster/client/java/controller/api/dto/problem/asyncapi/AsyncApiActionDto.java
LautaroPetaccio: Yes, you're right, it's better as an enum. Changed it!

arcuri82 (approved):

PR #1741, opened 2026-09-07
Java regex word boundary assertions

lmasroca:
Adds support for Java regex word boundary assertions `\b` and `\B`:
- `\b` is a word boundary assertion, which means that position is a boundary between a word character and a non-word character, so either a word character followed by non-word character or a non-word character followed by a word character. For example, the regex `.\b.` matches with ` a` as `a` is a word character and ` ` is non-word.
- `\B` is non-word boundary assertion, which means that position is not a boundary between a word character and a non-word character, so either a word character followed by another word character or a non-word character followed by another non-word character. For example, `.\B.` matches with `aa` as `a` is a word character, and it also matches with `  ` (two spaces) since ` ` is non-word.

Both of these capture no actual characters since they are assertions, they check their position satisfies a condition on the matched string. As such, these use the assertion repair mechanisms to try and force their neighboring genes' values to satisfy the assertion. For these types of assertions we try to perform repairs following certain shapes, for example for `\b` we may attempt a word character to the left of the assertion and a non-word character to the right, or the other way around.

This PR also adds regex strings that were previously not supported as a new test `RegexLanguageToolTest`. These strings were collected from experiments with LanguageTool.

Note: Out-of-bounds input positions are considered non-word characters.
Note 2: Characters considered as word characters for `\b`/`\B` change with JDK version, and are also affected by flag `U` (`UNICODE_CHARACTER_CLASS`). Because of that, we collect these characters into a `MultiCharacterRange` on `UnicodeCache` and use those ranges to sample possible word/non-word characters when needed.


core/src/main/kotlin/org/evomaster/core/utils/UnicodeCache.kt
```diff
@@ -21,14 +22,12 @@ import java.util.concurrent.ConcurrentHashMap
...
-         * Furthermore, although the hashmap is mutable, the values inside are not
-         */
-        private val cache = ConcurrentHashMap<String, MultiCharacterRange>()
-    }
+object UnicodeCache {
+    /**
+     * WARNING: mutable static state. But as it is just a cache, it is not a problem.
+     * Furthermore, although the hashmap is mutable, the values inside are not
+     */
+    private val cache = ConcurrentHashMap<String, MultiCharacterRange>()
```
jgaleotti, line 41: add documentation to state which are the key,value types. Refer to docs/for_developers.md for more information

jgaleotti (changes requested):

jgaleotti (approved):

arcuri82 (approved):

PR #1750, opened 2026-09-12
Fix MongoDB query heuristic semantics

jgaleotti:
Correct MongoDB query heuristics for arrays, null and missing fields, numeric comparisons, and bitwise operations. Extract shared evaluation logic into MongoHeuristicsCalculatorHelper.
- Fix $all, $in, $nin, $not, and regex handling.
- Support numeric, bit-position, and binary bitmasks.
- Ignore query comments and represent empty queries with EmptyOperation.


client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -96,120 +98,147 @@ private static Truthness buildSafeScaledTruthness(double maxOfTrue) {
...
+            List<?> actualValueList = (List<?>) actualValue;
+            if (actualValueList.isEmpty()) {
+                return C_FALSE;
+            } else {
+                Truthness[] results = actualValueList.stream()
+                        .filter(element -> element instanceof String)
+                        .map(element -> (String) element)
+                        .map(element -> evaluateRegularExpression(element, pattern, taintHandler))
+                        .toArray(Truthness[]::new);
+                return buildOrAggregationTruthness(results);
```
LautaroPetaccio, line 217: Throws if the array holds no strings. The filter can empty the stream and [checkValidTruthnesses](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/distance-heuristics/src/main/java/org/evomaster/client/java/distance/heuristics/TruthnessUtils.java#L239) won't take an empty array.

```
{a:{$regex:"x"}} vs {a:[1,2,3]}   ->   IllegalArgumentException: null or empty Truthness instance
```

mongo 7.0.41 returns 0 documents for that, and [the pre-PR code](https://github.com/WebFuzzing/EvoMaster/blob/a0741b14152abf6a03d8711f6f6607c9d3882cc8/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L166) returned `C_FALSE`, so it's a regression rather than a gap. Nothing catches it on the way out ([MongoHandler:155](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoHandler.java#L155)), so the action loses its SQL heuristics too.

This is also the only array path not wrapped in `buildSafeScaledTruthness`, cf [385](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L385) and [489](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L489).

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
```diff
@@ -0,0 +1,59 @@
...
+                    mask |= 1L << pos;                               // pos 63 -> Long.MIN_VALUE, fine
+                } else {
+                    return OptionalLong.empty();                     // not a number
+                }
+            }
+            return OptionalLong.of(mask);
+        }
+        // value can be a byte array
+        if (value instanceof byte[]) {
+            byte[] bytes = (byte[]) value;
```
LautaroPetaccio, line 45: BinData masks NPE, only a raw `byte[]` gets through. mongo 7.0.41 matches both.

```
{a:{$bitsAllSet: BinData(0,"Ag==")}} vs {a:2}
```

Needs to take `org.bson.types.Binary` as well, reflectively like [BsonHelper.isBsonTimestamp](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BsonHelper.java#L123) so the driver stays out of the compile deps.

The `byte[]` branch isn't wasted, to be fair. The query arrives however the app built it, with no decode in between ([MongoOperationClassReplacement:20](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/MongoOperationClassReplacement.java#L20), [MongoFindCommand.getQuery()](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/MongoFindCommand.java#L48), [MongoHandler:155](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoHandler.java#L155)), so both shapes turn up. It's the decoded one that has no cover.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
```diff
@@ -0,0 +1,59 @@
...
+        if (value instanceof Number) {
+            return OptionalLong.of(((Number) value).longValue());
+        }
+        // value can be a list of positions
+        if (value instanceof List<?>) {
+            long mask = 0L;
+            for (Object p : (List<?>) value) {
+                if (p instanceof Number) {
+                    long pos = ((Number) p).longValue();
+                    if (pos >= Long.SIZE) {
```
LautaroPetaccio, line 33: No check for `pos < 0`. Shift counts get masked to 6 bits, so `1L << -1` sets bit 63 and `{a:{$bitsAllSet:[-1]}}` comes back as a match on `Long.MIN_VALUE`.

mongo 7.0.41 won't run that query at all: `Failed to parse bit position. Expected a non-negative number in: 0: -1`. Answering false would be fine. Answering match sends the search after something that can't happen.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
```diff
@@ -0,0 +1,59 @@
...
+     *   Only the least significant 64 bits are considered; higher-order bits must be zero.
+     *
+     * @param value the input value to convert. Must be a {@code Number}, a {@code List} of numbers, or a {@code byte[]}
+     *              that satisfies the described conditions.
+     * @return an {@code OptionalLong} containing the resulting bitmask if the input value can be converted successfully;
+     *         otherwise, an empty {@code OptionalLong}.
+     */
+    public static OptionalLong toBitMaskValue(Object value) {
+        // value can be a number
+        if (value instanceof Number) {
```
LautaroPetaccio, line 24: Nothing validates the mask. `3.9` truncates to 3 and matches `{a:3}`, `-1` becomes all-ones and matches `{a:-1}`. mongo 7.0.41 rejects both, `Expected an integer` and `Expected a non-negative number`.

The field side of the same comparison already validates, that's what [getIntegralLongValue](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/MongoUtils.java#L125) is for. Same call here, plus `>= 0`.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/QueryParser.java
```diff
@@ -54,18 +57,43 @@ public QueryOperation parse(Object bsonDocument) {
...
         }
 
-        Object normalizedDocument = normalizeTopLevelValueOperatorQuery(bsonDocument);
-        if (normalizedDocument == bsonDocument) {
+        Object normalizedDocument = normalizeTopLevelValueOperatorQuery(normalizedWithoutComments);
+        if (normalizedDocument == normalizedWithoutComments) {
             return operation;
         }
         return parseWithSelectors(normalizedDocument);
     }
```
LautaroPetaccio, line 72: Recurses through the whole query, so it strips the key out of values as well as operator objects.

```
doc    {a: {$comment: "note", x: 1}}
query  {a: {$eq: {$comment: "note", x: 1}}}

mongo 1 document, calculator no match
```

`$`-prefixed field names store fine since 5.0, so that pair is not hypothetical. Keeping the strip at the top level fixes it and stops the query being deep-copied on every `parse`.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/QueryParser.java
```diff
@@ -19,6 +21,7 @@
 public class QueryParser {
 
     private static final String SYNTHETIC_FIELD_NAME = "$";
+    private static final Set<String> COMMENTS_OPERATORS = new HashSet<>(Arrays.asList("$comment", "$comments"));
```
LautaroPetaccio, line 24: `$comments` isn't a thing. mongo 7.0.41 says `unknown top level operator: $comments` and refuses the query. Dropping it here means we score something that could never have run.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -475,171 +399,139 @@ private Truthness computeHeuristic(AndOperation operation, Object document) {
...
-        final String fieldName = operation.getFieldName();
-        if (!documentContainsField(document, fieldName)) {
-            return C_FALSE;
-        } else if (expectedValues.isEmpty()) {
-            return C_FALSE;
-        } else {
-            Object actualValues = getValue(document, fieldName);
-            if (actualValues == null || !(actualValues instanceof List<?>)) {
+        if (!(actualValue instanceof List<?>)) {
+            if (expectedValues.size() != 1) {
```
LautaroPetaccio, line 440: `$all` is an `$and` of `$eq`, so repeats against a scalar are fine. mongo 7.0.41 returns 1 document for both of these, we return no match.

```
{a:{$all:[1,1]}}        vs {a:1}
{a:{$all:[null,null]}}  vs {b:1}
```

Same assumption in the [single-element branch](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L433) just above. Dedupe `expectedValues` first and both go.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -69,7 +79,7 @@ private Truthness computeHeuristicOnDocuments(QueryOperation operation, Iterable
         double maxOfTrue = 0;
         boolean first = true;
         for (Object doc : documents) {
-            double ofTrue = computeHeuristicOnDocument(operation, doc).getOfTrue();
+            double ofTrue = computeHeuristicQueryOperation(operation, doc).getOfTrue();
```
LautaroPetaccio, line 82: Three walks over `documents` here: [57](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L57) counts it, [74](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L74) counts it again, then this loop. [MongoHandler:149](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoHandler.java#L149) peeks before handing it over, so four.

It's a [FindIterable](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/internal/db/mongo/MongoHandler.java#L134), so it replays and nothing is broken. But every replay is another `find` on the wire. Four collection scans per action, on the path that runs for every action. One copy into a `List` at the top of [computeDistanceDocuments](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L56) covers it, and makes the `Iterable` in the signature honest.

Pre-existing, only raising it because it's the method you rewrote.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,440 @@
...
+                    throw new IllegalArgumentException("Unsupported comparison operator type: " + comparisonOperatorType);
+            }
+        } else {
+            // if both values are not NaN, we can use the standard comparison logic
+            final Truthness truthnessOfComparison = SqlExpressionEvaluator.calculateTruthnessForNumberComparison(leftNumber, rightNumber, comparisonOperatorType);
+            return truthnessOfComparison;
+        }
+    }
+
+    Truthness compareNullableValues(Object leftValue, SqlExpressionEvaluator.ComparisonOperatorType comparisonOperatorType, Object rightValue) {
```
LautaroPetaccio, line 227: Names are the wrong way round. [243](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java#L243) calls `compareNonNullValues(rightValue, op, leftValue)`, so the contract is `(expected, op, actual)`, which is what makes `$gt` read as actual > expected.

[127](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java#L127) and [283](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java#L283) pass them the other way. That reaches the taint handler at [153](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java#L153) with the arguments swapped, which costs nothing today because [ExecutionTracer.handleTaintForStringEquals](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/staticstate/ExecutionTracer.java#L322) takes either direction. Renaming these to `expectedValue`/`actualValue` would keep it that way.

The same split shows up in tainting. `$ne` over a field holding an array reaches `computeHeuristicContainsElement`, which compares with `EQUALS_TO` and so records a specialization, while `$ne` over a scalar field goes through this method with `NOT_EQUALS_TO` and records nothing. One operator, two behaviours, decided by what the field happens to hold rather than by the query.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,440 @@
...
+        }
+        return res;
+    }
+
+    Truthness evaluateMod(long divisor, long expectedRemainder, Object actualValue) {
+        if (!(actualValue instanceof Number)) {
+            return C_FALSE;
+        }
+
+        long actualRemainder = ((Number) actualValue).longValue() % divisor;
```
LautaroPetaccio, line 316: `{a:{$mod:[0,0]}}` throws `ArithmeticException`. mongo 7.0.41 rejects that query (`divisor cannot be 0`) so it shouldn't reach us, and [the same line before this PR](https://github.com/WebFuzzing/EvoMaster/blob/a0741b14152abf6a03d8711f6f6607c9d3882cc8/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L683) does it too, but the method moved into this class so it's a cheap guard while you're here.

LautaroPetaccio (commented):
The answers quoted in the comments below are what a mongo 7.0.41 returns for those queries, not what the documentation implies.

Two queries mongo accepts make the calculator throw: `$regex` on an array holding no strings, and a BinData bitmask. The first one is a regression, [the code this replaces](https://github.com/WebFuzzing/EvoMaster/blob/a0741b14152abf6a03d8711f6f6607c9d3882cc8/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L166) returned `C_FALSE` for it. Nothing catches either on the way out of [computeDistanceDocuments](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L56), so the action loses its SQL heuristics along with the mongo ones.

The rest are wrong answers rather than crashes. No validation on the bitmask itself, `$comment` stripped out of things that aren't operators, `$all` wrong on a repeated element.

Two that look wrong and aren't: nested `$not` agrees with mongo both ways, and the reversed actual/expected in [evaluateListEquality](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java#L115) does hit the taint handler backwards, but [ExecutionTracer.handleTaintForStringEquals](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/staticstate/ExecutionTracer.java#L322) takes either direction so nothing comes of it.

Not your change, but the unparsed-query-then-NPE path that [testOperatorsThatAreNotModelledDoNotThrow](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/test/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorTest.java#L2029) documents is bigger than that disabled test implies. `{a:{$not:/x/}}` lands in it, and that one mongo answers fine.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,440 @@
...
+            if (taintHandler != null && comparisonOperatorType == SqlExpressionEvaluator.ComparisonOperatorType.EQUALS_TO) {
+                taintHandler.handleTaintForStringEquals(actualString, expectedString, false);
+            }
+            truthnessOfComparison = SqlExpressionEvaluator.calculateTruthnessForStringComparison(actualString, expectedString, comparisonOperatorType);
+
+        } else {
+            // If both types are supported, but no actual comparison logic is defined,
+            // we considered them to be incompatible, therefore the comparison returns true
+            // only if the comparison operator is NOT_EQUALS_TO. Otherwise returns false.
+            truthnessOfComparison = comparisonOperatorType == SqlExpressionEvaluator.ComparisonOperatorType.NOT_EQUALS_TO ? TRUE_C : C_FALSE;
```
LautaroPetaccio, line 187: Two sub-documents never reach a comparison. The branches above cover numbers, strings, booleans, lists, dates, timestamps and object ids, and two documents fall through to here, which answers true for `$ne` and false for everything else.

The half that hurts is `$ne` and `$nin`, reporting a match mongo 7.0.41 does not have:

```
{a:{$ne:{x:1}}}      vs {a:{x:1}}   mongo 0 documents, we match
{a:{$nin:[{x:1}]}}   vs {a:{x:1}}   mongo 0 documents, we match
```

The other half is misses, and it takes in the plainest nested-object query there is:

```
{a:{x:1}}            vs {a:{x:1}}          mongo 1, we do not match
{a:{$eq:{x:1}}}      vs {a:{x:1}}          mongo 1, we do not match
{a:{x:1}}            vs {a:[{x:1},{x:2}]}  mongo 1, we do not match
{a:{$gt:{x:0}}}      vs {a:{x:1}}          mongo 1, we do not match
```

To be fair this is already better than [what it replaces](https://github.com/WebFuzzing/EvoMaster/blob/a0741b14152abf6a03d8711f6f6607c9d3882cc8/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L232), which threw on a sub-document rather than answering at all. And [testFieldsHoldingASubDocument](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/test/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorTest.java#L1764) does cover sub-documents, but against a scalar, where false is the right answer, so the document-to-document case falls outside it.

Binary data lands in the same place, with the same split between the two directions:

```
{a:{$eq:BinData(0,"AQI=")}}  vs the same value   mongo 1, we do not match
{a:{$ne:BinData(0,"AQI=")}}  vs the same value   mongo 0, we match
```

The rest of this area agrees with 7.0.41: `$ne` and `$nin` over arrays and missing fields, five `$elemMatch` shapes, `$size`, `$exists`, int against long against double against Decimal128, negative and fractional `$mod`, `$regex` with options, all four bit operators, `$nor`, nested `$and` and `$or`.

One more cost, beyond the answer. Neither branch records a string specialization, so a tainted value inside a sub-document is never offered to the search either. `{a:{x:"..."}}` against a document whose `x` holds a tainted string records nothing, where the same comparison one level up records it.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -310,49 +262,20 @@ private static boolean isTypeSupportedForComparison(Object value) {
...
-                    return (expectedValue == null && actualValue == null) ? C_FALSE : TRUE_C;
-                case GREATER_THAN:
-                case GREATER_THAN_EQUALS:
-                case MINOR_THAN:
-                case MINOR_THAN_EQUALS:
-                    return C_FALSE;
-                default:
-                    throw new IllegalArgumentException("Unsupported comparison operator type: " + comparisonOperatorType);
-            }
+        if ((actualValue instanceof List<?>) && !(expectedValue instanceof List<?>)) {
```
LautaroPetaccio, line 272: This condition sends a list expected value straight to `compareNullableValues`, which only ever compares the two arrays as wholes. So `$eq` misses an array holding the given array as one of its elements:

```
{a:{$eq:[1,2]}} vs {a:[1,2]}       mongo 1, we match
{a:{$eq:[1,2]}} vs {a:[[1,2],3]}   mongo 1, we do not
```

[$in](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L412) and [$all](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java#L456) both find a nested array in that position, so `$eq` is the one of the three that does not.

This one also pays to stay wrong. The document scores `ofTrue` 0.48970, against 0.10891 for `{a:{$eq:100}}` over `{a:0}`, which is reachable and merely far away, and 0.52857 for `{a:{$gt:100}}` over `{a:99}`, which one mutation satisfies. So a document that can never be reported as matching outranks a reachable one by more than four times, and the search climbs towards it.

LautaroPetaccio (commented):
Two more, both in code this pull request rewrote. The answers quoted are what a mongo 7.0.41 returns for those queries.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
```diff
@@ -475,171 +399,139 @@ private Truthness computeHeuristic(AndOperation operation, Object document) {
...
-    private static Truthness buildSafeScaledTruthness(Truthness truthness) {
-        return buildSafeScaledTruthness(truthness.getOfTrue());
-    }
+        List<?> actualValues = (List<?>) actualValue;
+        if (actualValues.isEmpty()) {
+            return C_FALSE;
+        }
 
+        Truthness res = buildAndAggregationTruthness(expectedValues.stream()
+                .map(expectedValue -> helper.computeHeuristicContainsElement(expectedValue, actualValues))
```
LautaroPetaccio, line 456: `$all` accepts `$elemMatch` among its elements, and each one then has to be satisfied by some element of the array. Here every element of `$all` is compared for equality against the elements of the field, so the document holding the operator never equals any of them.

```
{a:{$all:[{$elemMatch:{$gt:2}}]}}                        vs {a:[1,5]}   mongo 1, we do not match
{a:{$all:[{$elemMatch:{$gt:4}},{$elemMatch:{$lt:2}}]}}  vs {a:[1,5]}   mongo 1, we do not match
```

The second one is the shape that makes the operator worth having: 5 satisfies the first condition and 1 the second, which no single `$elemMatch` could express.

The gradient points the same way: `ofTrue` 0.34390 for an array that cannot be reported as matching. Worth contrasting with the conditions that are equally unreachable but correctly flat, `$size` against a scalar and `$regex` against a number, which both sit at 0.10000 and offer the search nothing to climb.

LautaroPetaccio (commented):
One more in code this pull request rewrote, plus two that are not yours but sit next to the ones already noted.

`{a:{$gt:null}}` and its three siblings are not parsed, so the calculator throws. mongo 7.0.41 runs all four and matches nothing, and `$eq` and `$ne` given null answer normally, so the ordering operators are the odd ones out in their own family. `$exists` given anything but a boolean throws the same way, where mongo takes any value and reads it as a truth value, so `{$exists:1}` and `{$exists:0}` are both queries it accepts and answers. Both selectors are untouched here, so this is the same route as [testOperatorsThatAreNotModelledDoNotThrow](https://github.com/WebFuzzing/EvoMaster/blob/be1815da425faf6dea3dc3179d0a3e182a8bec23/client-java/controller/src/test/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorTest.java#L2029) rather than anything new.

Gradients are in good shape otherwise. Over fifteen families of documents ordered from nearest to furthest from matching, fourteen fall away monotonically, including `$size`, `$mod`, the bit operators, `$all`, `$elemMatch` and `$gt` over an array. The exception is string equality, where any difference in length flattens the score to the same value whatever the content, so `"help"` and `"zzzz"` score identically against `"hello"` and both score below `"zzzzz"`. That one comes from `calculateTruthnessForStringComparison`, which master calls the same way, so it is not this pull request's doing.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: Mongo parser now rejects the query. Since it is no parsed, no heuristic score is computed on this query. So the test has no meaning.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
jgaleotti: thanks for reporting. It is fixed now.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
jgaleotti: We check first for whole equality and if not met, then a elem by elem check is performed.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: I changed all the left/right to expected/actual and make sure that the propagation is correct.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
jgaleotti: added

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
jgaleotti: fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
jgaleotti: thanks

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/utils/BitmaskUtils.java
jgaleotti: thanks

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/QueryParser.java
jgaleotti: fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculator.java
jgaleotti: fixed

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/QueryParser.java
jgaleotti: fixed

jgaleotti:
@arcuri82:  you might review this PR now.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/selectors/TypeSelector.java
```diff
@@ -12,16 +16,154 @@ public class TypeSelector extends SingleConditionQuerySelector {
...
+            }
+            return Arrays.asList(bsonType);
+        } else {
+            return null;
+        }
+    }
+
+    static String getBsonTypeEnumNameFromAlias(String alias) {
+
+        switch (alias) {
```
arcuri82, line 120: i don't understand this switch. why not just returning `alias.toUpperCase()`?

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,587 @@
...
+import static org.evomaster.client.java.sql.heuristic.SqlExpressionEvaluator.ComparisonOperatorType.EQUALS_TO;
+import static org.evomaster.client.java.sql.heuristic.SqlExpressionEvaluator.ComparisonOperatorType.NOT_EQUALS_TO;
+
+public class MongoHeuristicsCalculatorHelper {
+
+    // TODO these constants should be replaced by DistanceHelper constants
+    public static final double C = 0.1;
+    public static final Truthness C_FALSE = new Truthness(C, 1.0);
+    // TODO These constants should be refactored by TruthnessUtils constants
+    public static final Truthness TRUE_C = new Truthness(1.0, C);
```
arcuri82, line 37: why TODOs and not just doing it?

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,587 @@
...
+
+    private final TaintHandler taintHandler;
+
+    public MongoHeuristicsCalculatorHelper(TaintHandler taintHandler) {
+        this.taintHandler = taintHandler;
+    }
+
+    static Truthness buildSafeScaledTruthness(double maxOfTrue) {
```
arcuri82, line 45: why are these methods package-level scoped?

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
```diff
@@ -0,0 +1,587 @@
...
+            double ofTrue = 1d / (1.1d + distance);
+            return buildSafeScaledTruthness(ofTrue);
+        }
+    }
+
+    static int toIntValue(Boolean actualValue) {
+        return actualValue ? 1 : 0;
+    }
+
+    static Truthness buildSafeScaledTruthness(Truthness truthness) {
```
arcuri82, line 99: not going to repeat comment on package-level scoped, it applies to all method here

arcuri82 (changes requested):

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/selectors/TypeSelector.java
jgaleotti: Please notice that, although alias.toUpperCase() works in some cases (e.g. "double", "string", etc.) it is not a general solution. For example, it does not work for:
- "object" -> "DOCUMENT"
- "binData" -> "BIN_DATA"
- "objectId" -> "OBJECT_ID"
- "bool" -> "BOOLEAN"
- "minKey" -> "MIN_KEY"
- "maxKey" -> "MAX_KEY"
- "int" -> "INT_32"
- "long" -> "INT_64"
- "javascriptWithScope" -> "JAVASCRIPT_WITH_SCOPE"
- and some other examples. 
Therefore, I took the decision to explicitly map each possible schema value type into its corresponding enum name in org.bson.BsonType

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: This //TODO was previously in MongoHeuristicsCalculator, and now it was copied to the MongoHeuristicsCalculatorHelper class. I can definitively do it now.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: good point. Going to change the visibility to public for all these methods.

client-java/controller/src/main/java/org/evomaster/client/java/controller/mongo/MongoHeuristicsCalculatorHelper.java
jgaleotti: ack

arcuri82 (approved):

PR #1753, opened 2026-09-12
AsyncAPI 3.x: the fitness, and --problemType ASYNCAPI runs

LautaroPetaccio:
Twelfth in the AsyncAPI stack, on top of #1752. With this one `--problemType ASYNCAPI` runs a whole search, everything but writing the tests.

## What a search over an AsyncAPI service looks for

REST covers `(status × endpoint)`. There are no status codes here, so the targets are:

- **`(outcome × operation)`** — what publishing to an operation was seen to do: `PUBLISHED` (fire-and-forget), `REPLIED`, or `NO_REPLY`.
- **`(declared reply × operation)`** — when a reply came back, which of the messages the contract declares for that reply it was. A reply channel that lists a result and an error gives the search two things to reach, and telling them apart is what makes an asynchronous service observable from outside.

Two things are faults (both `@Experimental` categories): a promised reply that never arrives, and a reply that matches none of the declared messages. A message the driver could not publish is neither — it is a broken setup, no target is registered for it, and the test stops there.

## Recognising a reply

`AsyncApiReplyClassifier` matches the observed payload against each declared payload schema. It is deliberately a classifier, not a validator: it reads the parts of JSON Schema that tell one message from another — `type` (including type lists), `required`, `const`/`enum`, `allOf`/`anyOf`/`oneOf`, `items`, `$ref` to component schemas — and gives the benefit of the doubt on anything it does not understand (formats, bounds, patterns). When several declared messages fit, the most specific wins: the one pinning down the most top-level fields. No new dependency.

## The fitness

`AsyncApiBlackBoxFitness` resolves everything the driver needs from the document — address, content type, correlation location and pointer, reply address, timeout — so the driver never reads the document. The correlation id is `runId-counter`, with `runId` drawn from `Randomness`, so a seeded run is reproducible and ids never repeat within a run. Headers travel as a map by way of the gene's own JSON printing, which is what knows which optional headers are on.

Not supported yet, and said so with a one-time warning: a reply address announced inside the request (`reply.address.location`) rather than fixed by the contract. Such an operation is published without waiting.

## Plumbing

- `RemoteController.executeNewAsyncApiActionAndGetReply`, mirroring the RPC call. It has a **default that throws**, so the four other controllers (three of them test fakes) need not each say they cannot publish.
- `EMConfig.asyncApiReplyTimeoutMs`, `@Experimental`, default 5 s.
- **A constraint: `--createTests false` is required for AsyncAPI** until the test writer exists. Failing at start-up beats searching for an hour and failing at the end. `TestCaseWriter` is bound to the existing `NoTestCaseWriter`.
- `AsyncApiModule`, one for both modes, binding the driver unconditionally. It does not call `super.configure()`, like the RPC, GraphQL and Web modules: what `EnterpriseModule` binds there needs `FitnessFunction<RestIndividual>`.
- `AsyncApiStructureMutator`: one message more or one fewer. Unlike RPC's, it states its two bounds outright — a test keeps at least one message, and can grow to exactly `maxTestSize` — because mirroring RPC's arithmetic emptied a one-message test when `maxTestSize` was 2, and never let a test reach the maximum.
- **Seeding test cases is refused at start-up too**, for the same reason as writing them.
- `Main`'s `ASYNCAPI` branch binds the module instead of throwing.

## Testing

Three suites build the injector `Main` would, with the driver replaced by a fake, through one shared `AsyncApiTestInjector`.

- `AsyncApiReplyClassifierTest` (16): result vs error behind `$ref`s, undeclared payloads, non-JSON, `3.0` is an integer, most-specific-wins, type lists; and, on a document written for the purpose, `const` through a chain of references, `enum`, `oneOf`/`allOf`/`anyOf`, array items, booleans vs their spelling, an unknown type name, and a reference into the middle of a schema being given the benefit of the doubt.
- `AsyncApiBlackBoxFitnessTest` (13): every DTO field the driver is told; unique correlation ids across evaluations; result and error as distinct targets; undeclared reply and silence as the two faults; fire-and-forget covered by `PUBLISHED`; headers as a map without the stamped one, structured headers as JSON; a correlation id located in the payload; a reply address announced at run time being published without waiting; a driver that could not publish, with and without saying why; a reply without its correlation id recorded rather than judged.
- `AsyncApiStructureMutatorTest` (4): the two bounds above, the maximum being reachable, added messages coming from the document, nothing happening when only one message is allowed. The first fails against RPC's arithmetic.
- `AsyncApiModuleTest` (1): a whole MIO search, 100 action evaluations, black-box, through the module `Main` binds, against a stand-in that answers like the NCS service. Asserts that every operation was answered, that some operation reached **both** its declared replies, that nothing was a fault, and that no correlation id repeated.

All AsyncAPI suites plus the config ones: **193 tests, 0 failures**. jacoco over the AsyncAPI package: every new class above 90% line coverage.

---

## Review round

A full review over this PR found the following, all fixed here:

- **Messages went to the wrong destination.** The address ignored protocol bindings. A Kafka channel usually names no address and carries its topic in `bindings.kafka.topic` — `microcks.yaml`, already a test resource here, is exactly that shape — so messages were published to the channel's *name*. The parser's `effectiveAddress()` exists for this and was never called.
- **One odd reply ended the run.** Deciding whether a number is an integer asked Jackson for a `BigDecimal`, which throws for a value that overflows a double. Nothing between the classifier and the search loop catches it.
- **Valid replies were reported as faults**: an empty or non-JSON body; a `const`/`enum` written `1.0` against a reply carrying `1` (Jackson's equality is type-strict, JSON Schema's is by value); a `type` or combinator the classifier cannot read. A `$ref` pointing *into* a schema matched everything, which quietly disabled the undeclared-reply oracle for any operation written that way.
- **Faults ignored `isEnabledFaultCategory`**, so two experimental oracles fired on a default run and `--disabledOracleCodes` did nothing. **They now require `--useExperimentalOracles`**, like every other experimental oracle. They are also recorded on the action result, which is where the reports count faults from — before, a run covered fault targets while `totalNumberOfDetectedFaults()` said zero.
- `AsyncApiCallResult` was the only action result not overriding `matchedType`.
- **Problem-type auto-detection was a dead path**: with `--problemType` omitted, `Main` inferred ASYNCAPI and then threw on the `createTests` constraint, after the SUT had been started. It now turns test generation off with a warning.
- The class is `AsyncApiFitness`: it is bound as the only fitness and reads white-box coverage, so `BlackBox` in the name was untrue.

The guards that keep a valid reply from becoming a fault had no tests, which is why they could be wrong; they are covered now, along with the binding topic, the configured timeout, and the fault gate. The structure mutator's tracking is exercised with a real `MutatedGeneSpecification` rather than `null`, and the module test asserts the bindings `Main` resolves.

**216 tests, 0 failures.**


core/src/main/kotlin/org/evomaster/core/problem/asyncapi/service/AsyncApiFitness.kt
```diff
@@ -0,0 +1,324 @@
...
+        when (outcome) {
+
+            AsyncApiOutcome.REPLIED -> handleReplyTargets(fv, action, result, index)
+
+            AsyncApiOutcome.NO_REPLY -> {
+                val fault = idMapper.getFaultDescriptiveId(ExperimentalFaultCategory.ASYNCAPI_NO_REPLY, name)
+                fv.updateTarget(idMapper.handleLocalTarget(fault), 1.0, index)
+            }
+
+            AsyncApiOutcome.PUBLISHED, AsyncApiOutcome.PUBLISH_FAILED -> Unit
```
jgaleotti: what if none of the AsyncApiOutcome matches?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/service/AsyncApiFitness.kt
LautaroPetaccio: They must all match, as the compiler will not compile otherwise. Did you mean that or were you talking about a reply not matching?

LautaroPetaccio:
I forgot to include the replies which were not matched with the schema as possible targets. Now they're included.

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/service/AsyncApiFitness.kt
```diff
@@ -0,0 +1,484 @@
...
+    private lateinit var asyncApiSampler: AsyncApiSampler
+
+    /**
+     * Tells this run's correlation ids from those of an earlier run against the same broker, so
+     * that a reply left over from a previous run is not read as an answer to this one.
+     *
+     * Deliberately not drawn from [randomness]: a run repeated under the same seed would reuse
+     * the ids it used before, which is the one case this exists to tell apart.
+     */
+    private val runId: String = UUID.randomUUID().toString().take(8)
```
jgaleotti, line 97: This is problematic since it might lead to different behaviour even when the same -seed is introduced, right?

jgaleotti (changes requested):

core/src/main/kotlin/org/evomaster/core/problem/asyncapi/service/AsyncApiFitness.kt
LautaroPetaccio: You're right! Great catch. I've changed it to use `randomness` which takes into consideration the seed.

jgaleotti (approved):

PR #1763, opened 2026-09-14
capture Cypher queries run inside transactions

andyfelder16:
Looking at candidate SUTs for the Neo4j experiments, most queries are not run on the session directly: Spring Data Neo4j runs them inside a transaction, and driver code tends to use `session.beginTransaction()` or `session.executeRead(...)`. Those go through `Transaction`, `TransactionContext` and `QueryRunner`, which expose the same `run` overloads as `Session` but were not instrumented.

- One replacement class per type, since call sites are matched on the exact owner name: `Neo4JTransactionClassReplacement`, `Neo4JTransactionContextClassReplacement`, `Neo4JQueryRunnerClassReplacement`
- The tracking logic moved to `Neo4JOperationClassReplacement`, so each class only declares its five `run` overloads
- The existing test became an abstract base with one small subclass per type, so all four run the same 8 tests against a Testcontainers Neo4j
- Test driver bumped from 4.4.12 to 5.20.0, `TransactionContext` only exists in driver 5


jgaleotti (approved):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/Neo4JOperationClassReplacement.java
```diff
@@ -4,12 +4,29 @@
 import org.evomaster.client.java.instrumentation.coverage.methodreplacement.ThirdPartyMethodReplacementClass;
```
arcuri82, line 4: @andyfelder16  @jgaleotti  shouldn't we have at least 1 E2E test for this PR? or is support for Neo4J not ready yet for E2E?

arcuri82 (changes requested):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/Neo4JOperationClassReplacement.java
andyfelder16: I am currently working on data insertion for Neo4j. I can add E2E tests to this PR now, or add them together with the insertion PR, so they cover everything end to end. I am happy to go with whichever you prefer.

arcuri82: @andyfelder16  thx! you can have them on a different PR (so I review this one here now). but, in future, might be best to have E2E in same PR of a new introduced feature if that feature is finished and ready to have a E2E (PRs that are work-in-progress feature of course can't have E2E yet, as they would fail)

arcuri82 (approved):

client-java/instrumentation/src/main/java/org/evomaster/client/java/instrumentation/coverage/methodreplacement/thirdpartyclasses/Neo4JOperationClassReplacement.java
andyfelder16: I created this PR of E2E tests: https://github.com/WebFuzzing/EvoMaster/pull/1773 

I'm working on the data insertion logic, once I finish it I will also add E2E tests with it.

PR #1773, opened 2026-09-17
add end to end tests for Neo4j heuristics, and the option that enables their instrumentation

andyfelder16:
- `instrumentMR_NEO4J` added to `EMConfig`, experimental and off by default like the ones for Redis and DynamoDB, plus its entry in `docs/options.md`. It was missing: all the Neo4j method replacements belong to `ReplacementCategory.NEO4J`, which `methodReplacementCategories()` never listed, so no Cypher query was captured in an actual search
- New module `spring-rest-neo4j`, following the structure of `spring-rest-redis`: a Testcontainers-based `Neo4jController`, and one `App` / `Rest` / controller / `EMTest` per scenario
- Two scenarios, each with a `POST` that creates a node and a `GET` that returns 200 or 404 depending on whether a `MATCH` finds it: `session` runs its queries with `Session.run`, `transaction` runs them inside `readTransaction(tx -> tx.run(...))`
- Each test asserts that the search reaches both the 404 and the 200 of the `GET`
- Scenarios that need generated data will come with the data insertion PR


jgaleotti (approved):

core-tests/e2e-tests/spring/spring-rest-neo4j/src/main/java/com/neo4j/session/findnode/Neo4jSessionFindNodeRest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Collections;
+
+/**
+ * Runs its queries directly on the session, in auto-commit transactions.
+ */
+@RestController
+@RequestMapping(path = "/neo4jsessionfindnode")
+public class Neo4jSessionFindNodeRest extends AbstractNeo4jRest {
+
+    @PostMapping("/person/{name}")
```
arcuri82: based on Black-Box heuristics (not necessarely now, but in future) it could connect a `POST:/person/{name}` to a `GET:/findPerson/{name}`. To avoid this, and make more likely that finding the solution is for Neo4J heuristics and not something else, I would rename `POST:/person/{name}` into something like `POST:/x/foo/{y}/bar`

core-tests/e2e-tests/spring/spring-rest-neo4j/src/main/java/com/neo4j/SwaggerConfiguration.java
```diff
@@ -0,0 +1,37 @@
+package com.neo4j;
```
arcuri82: this get confusing with the real `org.neo4j`. maybe call it `com.foo.neo4j`?

core-tests/e2e-tests/spring/spring-rest-neo4j/src/main/java/com/neo4j/SwaggerConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.web.context.request.WebRequest;
+import springfox.documentation.builders.ApiInfoBuilder;
```
arcuri82: should not use springfox for new modules, but rather `SpringDoc`

arcuri82 (changes requested):

arcuri82 (approved):

PR #1729, opened 2026-08-31
Cassandra Actions

gonzalotguerrero:

PR #1777, opened 2026-09-18
NPE in TaintedArrayGene

gonzalotguerrero:
#1729 adds a new gene class for CQL's duration data type

After merging _master_ into that branch, test case with seed 1384 in class _GeneRandomizedTest_ began to fail due to a NPE in TaintedArrayGene's **unsafeCopyValueFrom** implementation, more specifically in
`return this.arrayGene?.unsafeCopyValueFrom(other.arrayGene!!)
            ?: true`
due to _other.arrayGene_ being null.

Copying from an unresolved gene into a resolved one threw a NullPointerException, and the opposite direction reported a successful copy without copying anything. This PR suggests returning **false** for both scenarios.

**TaintedArrayGeneTest.testChoiceCopiesIntoTheAlternativeInTheSameState()** represents the scenario for seed 1384 (for the current list of available gene classes) from which the NPE originated.


jgaleotti (approved):

core/src/main/kotlin/org/evomaster/core/search/gene/collection/TaintedArrayGene.kt
```diff
@@ -129,6 +129,14 @@ class TaintedArrayGene(
            return false
         }
 
+        /*
+            Resolving a taint adds the array as a child, so a resolved and an unresolved gene do not
+            have the same structure, and copying a value cannot turn one into the other
+         */
+        if(this.isResolved() != other.isResolved()){
```
arcuri82: rather than returning false, should handle this case. eg, `this.arrayGene` could be put to `null` or created, and `taintedValue` could be copied

arcuri82 (changes requested):

arcuri82 (approved):

PR #1778, opened 2026-09-18
Cancel superseded CI runs on the same branch

LautaroPetaccio:
Pushing several times to a branch leaves every run going, so a build for a commit nobody is waiting on any more keeps competing for runners with the one that matters. This groups the workflow per branch and cancels the run still in progress, leaving only the newest commit building.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```

**IMPORTANT:**
`master` is exempt, so a merge is never cut short by the one behind it. Its runs still share a group, which means **two merges in quick succession queue rather than overlap**. If that turns out to matter, appending `github.sha` for `master` gives it a lane of its own.

### Documentation pushes are unaffected

A push that only touches documentation produces no run at all, because `paths` is evaluated before a run exists and `concurrency` only acts on runs that already exist. A build carrying a code change is not cancelled by someone pushing prose on top of it.

`docs/options.md` is the one exception, and deliberately so: it is generated from `EMConfig` and `ConfigToMarkdownTest` checks it is current, so a push touching it does start a run. Nothing is lost when that run supersedes another, since a run checks out a commit and builds the whole tree rather than a diff, so the replacement covers the earlier change as well.

### Why the group is keyed on `github.ref`

The workflow triggers on `push` and the `pull_request` trigger above is commented out, so the ref is the branch. If that trigger is ever restored, the group needs `github.head_ref` alongside it, otherwise pushes to a pull request branch land in a different group and stop superseding one another.


jgaleotti (approved):

jgaleotti:
hi @arcuri82 , this PR is related to our discussion on Slack. I checked with @LautaroPetaccio and it seems that Pushes to documentation do not cancel previous Builds that might be currently running CI.

arcuri82 (approved):

PR #1780, opened 2026-09-21
Ai endpoint report

MohsenTaheriShalmani:
Adding ai snapshot report for the ML metrics per endpoint


core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -161,6 +166,9 @@ class Statistics : SearchListener {
      */
     private val snapshots: MutableMap<Double, List<Pair>> = mutableMapOf()
 
+    /** Values captured at each interval, rather than recomputed from the final classifier state. */
+    private val aiEndpointSnapshots: MutableMap<Double, List<List<String>>> = mutableMapOf()
```
arcuri82, line 177: add some comment on the `key` of this map and its `value` types

core/src/main/kotlin/org/evomaster/core/search/service/Statistics.kt
```diff
@@ -193,8 +201,65 @@ class Statistics : SearchListener {
...
+    internal fun writeAIEndpointStatistics() {
+        if (!canReportAIEndpointStatistics()) {
+            return
+        }
+
+        writeAIEndpointCsv(config.aiEndpointStatisticsFile, AI_ENDPOINT_HEADERS, getAIEndpointRows().asSequence())
+    }
+
+    private fun canReportAIEndpointStatistics(): Boolean =
+        config.isEnabledAIModelForResponseClassification() && this::aiResponseClassifier.isInitialized
```
arcuri82: should have an explicit boolean to specify whether we want or not to create such file, like we have `writeStatistics`

arcuri82 (changes requested):

arcuri82 (approved):

